PR #4, opened 2026-06-22
New codes (2BGA qldpc codes) + heuristic distance estimation notes

FarLab:
Add weight-8 coset two-block codes [[180,{16,18,20},<=14]] + distance-estimation note

Three upper-bound (d<=14) weight-8 CSS codes for the (currently unpopulated) weight-8 track, each with explicit weight-14 X/Z logical witnesses; the local verifier passes (earned_distance upper_bound, max check weight 8).

Construction: coset-based two-block (2BGA) codes on the metacyclic group C30 x| C6 (order 180), qubits on cosets G/H with a non-normal H=C2 -- a generalization of bivariate bicycle, not standard BB on a torus -- with a,b supports found by simulated annealing. Distance is a decoder (syndrome-decoder) and QDistRnd upper bound, cross-checked and slope-corroborated at 14, but NOT certified exact.

docs/distance_via_simulation.tex(+pdf): the decoder-simulation distance estimation method (syndrome-decoder upper bound, sub-threshold slope as support), including the multiprocessing / threaded-C++ parallelization notes.

 ---
  ## Code submission
  
  - Parameters: [[180,20,≤14]], [[180,18,≤14]], [[180,16,≤14]] — three weight-8 codes (eff $kd^2/n$ = 21.8,
  19.6, 17.4 assuming $d=14$)
  - Tracks: weight-8 (any-connectivity)
  
  ### Checklist

  - [x] JSON files under codes/ conforming to schema/code.schema.json — three files: codes/180-20-14.json,
  codes/180-18-14.json, codes/180-16-14.json
  - [x] Distance witness(es) included for each reported side — explicit weight-14 X and Z logical operators
  per code
  - [x] python verify/qldpc_verify.py codes/<file>.json passes locally — all three: exit 0, earned_distance
  upper_bound, max_check_weight 8
  - [x] Construction and references filled in under provenance — coset 2BGA on metacyclic $C_{30}\rtimes C_6$;
  ref arXiv:2606.17268; method in docs/distance_via_simulation.tex
  - [x] If this may be equivalent to an existing entry, noted in provenance.notes — noted that four
  [[180,16,14]] variants share parameters (one representative submitted), and that these are not standard
  bivariate bicycle (coset generalization) and not 2D-local

  Populates the currently-empty weight-8 track. The stated bar is the Liang–Eberhardt–Chen $k{=}12$ family
  ([[240,12,12]], [[292,12,14]], [[399,12,18]], [[282,12,14]]). These submissions add higher-$k$, smaller-$n$
  frontier points:
  
  - [[180,20,≤14]] Pareto-dominates [[240,12,12]] and [[292,12,14]] on $(n,k,d)$ — smaller $n$ (180), higher
  $k$ (20), equal-or-higher $d$ (≤14).
  - [[180,18,≤14]], [[180,16,≤14]] are further new $k{>}12$ points at $n{=}180$.
  - [[399,12,18]] stays on the frontier independently (higher $d$); these don't claim to beat it.

  Caveats (stated plainly): distances are upper bounds (witness-certified $d\le14$,
  decoder/QDistRnd-corroborated and slope-consistent, but not exact-certified — they show as d≤). And these
  are not 2D-local — unlike the planar LEC codes, they enter only the any-connectivity weight-8 track, not the
  2d-local-* tracks; the comparison is on $(n,k,d)$ at check weight ≤ 8 only.


vprusso (commented):
Reviewed the three 2BGA codes and the distance notes. Mechanically they are clean: all three pass the verifier (n and k recomputed over GF(2), CSS commutation holds, max check weight 8 so the weight-8 track is correct), the witnesses are valid weight-14 logicals, and CI is green. Three things to address before merge.

1. [[180,16,14]] is dominated by a published code in the cited paper. The weight-8 table in arXiv:2606.17268 includes [[168,16,15]]: same k=16, fewer qubits (168 < 180), and higher distance (15 > 14), so it strictly dominates [[180,16,14]] (kd2/n 21.4 vs 17.4). This one is not a record; suggest dropping it, the same call we made on [[288,7,12]].

2. The distances are upper bounds, not certified. All three are confidence=upper_bound (d <= 14 from the weight-14 witness). [[180,18,14]] and [[180,20,14]] would become the top kd2/n entries on the board (19.6 and 21.8, above the current best of 14.2) on an unverified distance. If the true distance is below 14, those headline numbers are overstated. Before merging them as records, the d=14 should be certified exact, or they should be flagged as provisional rather than posted as top-of-board records. n=180 is more tractable than the cases that defeated MILP earlier, so certification is worth attempting (running it now, will follow up with the result).

3. Missing baselines. The paper's published 2BGA weight-8 codes ([[84,16,8]], [[112,16,10]], [[128,16,12]], [[168,16,15]]) are not on the board. If we are opening up 2BGA, those are the bar to beat and should be seeded as baselines, which is also what makes the domination in point 1 visible rather than hidden. The k=18 and k=20 codes do look genuinely new (the paper's weight-8 codes are all k=16), so they are a real higher-k contribution if the distance holds.

Caveat: I checked against the paper's abstract listing, not the full tables, so worth confirming against the full paper that nothing tabulated dominates the k=18/20 codes.

Recommendation: drop [[180,16,14]]; certify the k=18/20 distances (or mark them provisional rather than records); and seed the paper's published codes as baselines so the comparison is honest. The codes that survive certification would be a strong addition (high-k weight-8, above the current kd2/n bar).

FarLab:
Thanks for the review! Dropped [[180,16,14]], but can't certify the exactly the distances (I did try it, but it took very long, more than a day, and still couldn't solve it). Maybe we should add a description on how the upper bounds are obtained once we have the harness for these heuristic distance estimation in place, because it does provide some confidence on the distance. @vprusso

vprusso:
Following up on the certification: I also couldn't close the exact distance for [[180,18,14]] or [[180,20,14]], same as you. The exact methods all stall at this size, MILP via HiGHS and Gurobi, and the connected-cluster algorithm from the codeDistance package (Webster et al., arXiv:2603.22532), which is expected for d=14 codes this large. So d<=14 as upper bounds is the right call, and the explicit decoder-found witnesses make those bounds trustworthy. Good to merge on that basis.

Agree on documenting how the upper bounds are obtained once the heuristic-distance harness is in place. That adds real confidence without overstating: they stay d<= (not certified d=), with a description of how the bound was found and why it is likely tight. The codeDistance package's QDistEvol and the connected-cluster/SAT backends are worth a look there as a reference point.

One remaining nice-to-have: seeding the paper's published 2BGA weight-8 codes ([[84,16,8]], [[112,16,10]], [[128,16,12]], [[168,16,15]]) as baselines, so these are measured against the bar. That needs constructing those codes, so it can be a follow-up rather than blocking this PR.

PR #5, opened 2026-06-22
Heuristic distance verify

FarLab:
Adding the heuristic distance verification harness.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -23,4 +23,4 @@ jobs:
       - name: verify all submissions
         run: uv run --frozen python verify/verify_all.py
       - name: distance refutation gate (changed submissions only)
-        run: uv run --frozen python verify/gate_changed.py origin/${{ github.base_ref || 'main' }}
+        run: uv run --frozen --with ldpc python verify/gate_changed.py origin/${{ github.base_ref || 'main' }}
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -15,8 +15,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0          # full history so the gate can diff vs the base
       - uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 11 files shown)


vprusso (commented):
Reviewed the CI-gating commit (63798a8, "Gate only changed submissions in CI").

This is the right scoping. Making the RIS refutation opt-in (refute=False by default) keeps the bulk paths cheap (test_verifier, verify_all, and the site build still skip the ~10s search), and the new gate runs it only on the code/example files a PR actually touches. The three-dot diff plus fetch-depth: 0 is the correct way to get the base for the comparison.

Verified locally: ran gate_changed.py on a correct code and on an over-claimed copy of the same code (claimed d bumped from 6 to 10). It passes the correct one ("no logical lighter than 6 in 5380 RIS trials") and refutes the over-claim with a checkable weight-6 witness, exiting 1. So the gate does what it says.

A few minor things, none blocking:

1. changed_codes() catches every exception and returns [] (gate nothing). For an integrity gate that is fail-open: if the base ref ever fails to resolve (origin/main missing, base_ref empty, a shallow clone), CI goes green without gating anything and nothing flags it. fetch-depth: 0 should make origin/main available so the normal path works, but this PR adds no codes so the gate path is never exercised in its own CI. Worth confirming on the first code-adding PR that the Actions log shows the file being gated, and maybe failing closed (or printing a loud warning) when a BASE was passed but the diff errored.

2. On push to main the step still runs but diffs HEAD against itself, so it is a no-op that still pays the full-history checkout. Could guard the step with `if: github.event_name == 'pull_request'`.

3. gate_changed.py imports heuristic_distance at module top, so a future dependency drifting out of the frozen lockfile would hard-fail the gate, whereas the inline path in qldpc_verify skips on import error. Failing closed here is arguably the safer choice for a gate; just noting the asymmetry. Currently fine since it only needs numpy and the local gf2.

Scoping the diff to `-- codes examples` is a good touch: editing verify/ or schema/ won't trigger code gating. Looks good to merge once the gate is confirmed firing on a real submission.

PR #6, opened 2026-06-23
Starter kit

FarLab:
## What and why

The repo could **verify and rank** codes but had no tooling for the front half of the loop: actually *constructing* a code, *estimating its distance*, and *packaging* a submission. A newcomer (or an LLM doing autoresearch) had to write all of that from scratch before they could engage.

This PR adds a `research/` starter kit that fills that gap. It sits directly on top of the verifier's own GF(2) core (`verify/gf2.py`), so what it computes is exactly what the board records. The core is **numpy-only** — no new required dependencies.

## The loop

```
pick a family -> build (HX,HZ) -> estimate distance -> package -> verify -> PR
   bb / group_algebra / coset     surrogate (witness free)   submit    verify/
                  |                                ^
      sweep with search.py          confirm with distance.py
      (screen -> rank -> frontier)  (exact MILP d= / decoder)
```

Start here:

```
uv run python research/recipes/01_build_and_submit_bb.py   # one code, end to end
uv run python research/recipes/02_search_a_family.py       # sweep a family, take the best
```

## What's in it

| File | Role |
|---|---|
| `css.py` | `compute_k`, `verify_css` + the verifier's GF(2) core re-exported (single source of truth) |
| `bb.py` | bivariate-bicycle codes on a torus — the entry family |
| `group_algebra.py` | 2BGA on any finite group (+ `perm_group`/`dihedral`/`metacyclic`/`sym`/`alt`; reaches odd k) |
| `coset.py` | coset 2BGA on G/H — the record-efficiency family |
| `surrogate.py` | `distance_rand` (upper-bound distance) + `lightest_logical` (the witness a submission needs) + `mixed_volume` (fast k bound) |
| `search.py` | `screen` -> `pareto_frontier`, `sample_bb`, `update_leaderboard` — the discovery funnel |
| `distance.py` | `exact_distance` (MILP `d=`) and `decoder_distance` (BP+OSD), **reusing** `verify/certify.py` and `decode/distance.py` |
| `submit.py` | `make_submission`: `(HX,HZ)` -> schema-valid JSON, witnesses pre-checked against the verifier's own criteria |
| `recipes/` | runnable end-to-end examples |
| `GETTING_STARTED.md` | LLM/newcomer-facing guide |

## Design choices

- **Reuses, doesn't re-port.** Distance confirmation calls the repo's existing certifiers rather than duplicating them; the GF(2) core is shared with the verifier so the two can't drift.
- **Honest distance semantics.** `distance_rand` is documented and labeled an *upper bound*; submissions default to `confidence="upper_bound"`; exact certification stays the separate server tier.
- **Dependency-tiered.** Constructors, surrogate, search, and packaging are numpy-only. The heavy deps (`scipy`, `ldpc`) are isolated behind a new optional `research` extra and only load when you confirm a distance.
- **Scoped.** The specialized open-boundary planar engine for `2d-local-bilayer` is intentionally left out.

## Verified

- Recipes 01 and 02 build codes that pass the **real verifier** in-process.
- `exact_distance` certifies `[[72,12,6]]` as **d=6 exact** via the existing MILP; `decoder_distance` corroborates **d=6**.
- Every module is independently runnable (`uv run python research/<module>.py`) with a self-test.

README and CONTRIBUTING now point newcomers here.


vprusso (commented):
Reviewed the research/ starter kit (the three commits on top of the #5 merge: 8e38764, c6b0f0a, 3fadb3a).

Strong addition, and it fills a real gap: the repo could verify and rank but had nothing for the build / estimate / package half of the loop. The design claims hold up under inspection and testing.

- Reuse is genuine, not re-ported. css.py re-exports verify/gf2.py; exact_distance loads verify/certify.py and decoder_distance loads decode/distance.py by explicit file path (a clean way around the research/distance.py vs decode/distance.py name clash) and calls their real entry points. The construction side and the board share one GF(2) core, so they can't drift.
- Honest distance semantics. surrogate.distance_rand is documented and labeled an upper bound, submissions default to confidence="upper_bound", and exact stays the separate server tier.
- Dependency tiering works as advertised. I ran recipes 01 and 02 with only numpy and they completed end to end; scipy/ldpc are isolated behind the new `research` extra and only load in distance.py.

Verified locally:
- Recipe 01 builds [[72,12,6]], surrogate d<=6, packages a schema-valid doc, and the real verifier passes it with tier=upper_bound.
- Recipe 02 screens 400 BB candidates, builds a Pareto frontier, and packages + verifies the winner ([[108,4,12]]).
- coset.py self-test confirms H={e} reduces to the 2BGA [[72,12,6]].

Minor, none blocking:

1. This branch is stacked on #5 and actually contains a merge of it (9bb4cd7), so the diff against main double-counts the heuristic-distance harness already reviewed in #5. Worth deciding the merge order: land #5 first and rebase this, or note that merging this brings #5 along.

2. Recipe 02 prints the Pareto frontier with a duplicate ([[42,4,6]] appears twice) — two different exponent sets with the same parameters, both kept. Dedup the printed frontier by (n, k, d) so it does not read like a bug.

3. Each research/ module has a self-test but none run in CI. That matches decode/ being offline tooling, so it is fine, but a tiny in-process smoke test (recipe 01 builds a code the verifier accepts) would catch drift if verify/gf2.py or the schema changes, which is the exact risk the single-source-of-truth design is meant to avoid.

Looks good to merge once the #5 stacking is sorted.

PR #7, opened 2026-06-25
Add installation instructions and Makefile shortcuts

willzeng:
## Summary

- Adds an **Installation** section to the README with uv setup steps and common commands for new contributors
- Adds a **Makefile** with three targets to reduce friction:
  - `make build` — regenerates the leaderboard site into `docs/`
  - `make verify CODE=codes/foo.json` — runs the local verifier on a submission
  - `make example` — verifies the bundled example code (`examples/72-6-6.json`)


willzeng:
Note that this also updates the `uv lock` file which is a little sloppy to do all together in one PR so I can split it if you prefer.

PR #22, opened 2026-06-26
Add [[120,8,12]] bivariate bicycle code (autoresearch)

FarLab:
Weight-6 CSS code on Z_12 x Z_5, found by automated random search over the bivariate-bicycle family. Non-dominated on the bivariate bicycle (periodic) / weight-6 board: extends the k=8 family to distance 12 (vs board [[90,8,10]]), encoding efficiency kd^2/n = 9.6.

Distance is a converged surrogate upper bound (stable through 16k random-restart trials), confidence=upper_bound -- not exactly certified. Novelty vs the BB literature is unverified; may coincide with a known code. Periodic (toroidal) connectivity, not 2D-local.

## Code submission

- Parameters: [[n, k, d]] = [[120, 8, 12]]
- Tracks: bivariate bicycle (periodic), weight-6

### Checklist

- [x] One JSON file under codes/, conforming to schema/code.schema.json — codes/120-8-12.json (branch currently holds 12; narrow to this file for a single-code PR)
- [x] Distance witness(es) included for each reported side — X and Z witnesses both present
- [x] python verify/qldpc_verify.py codes/120-8-12.json passes locally — exit 0, "ok": true
- [x] Construction and references filled in under provenance — bivariate-bicycle construction + arXiv:2308.07915

Track: bivariate bicycle (periodic) (also enters weight-6).

This adds a new non-dominated point in the k = 8 family. The nearest board entries at k = 8 are [[90,8,10]] and [[108,8,10]], both at distance 10. [[120,8,12]] extends the k = 8 frontier to distance 12 — it is not dominated by either (higher d at modestly larger n), and is not dominated by the higher-k [[144,12,12]] (larger n, different k). Encoding efficiency kd²/n = 9.60, above the k = 8 board leader [[90,8,10]] (8.89).

Caveats: distance is a converged surrogate upper bound (stable through 16k random-restart trials), not exa automated random search overthe BB family; novelty vs. the BB literature is unverified — it advances this board but may coincide with a known code.


FarLab:
@willzeng @vprusso I made this PR to test the auto-research capabilities of the harness in place. It seems to work. Lessons learned from this small experiment: having a larger list of known techniques/family of codes would make the procedure easier + some kind of research prompts that we provide so the barrier to entry gets even lower (user just fires up their llm and chooses a research direction and auto-research starts).

vprusso:
Deep-checked the code before signing off: the board verifier passes (w=6, CSS ok, k=8), and the d<=12 bound holds under 600k random-restart RIS trials per side, well past the 16k here. So kd^2/n = 9.6 is a solid upper bound rather than a loose surrogate, and it is non-dominated (beats the current k=8 best, [[90,8,10]] at 8.89). Good to land on the BB-periodic / weight-6 boards as a d<= entry. Distinctness from the literature is still unverified, as you note; worth a line in provenance if anyone recognizes it.

On the auto-research lessons, from running a fair amount of the same loop on this repo:

1. A family list helps, but the metadata that actually saves cycles is whether each family's frontier is saturated. Random search over weight-6 BB and the notched k=5/6 families mostly re-finds dominated points or k=4 filler, since those are carefully constructed (IBM BB and friends) and basically mined out. The real headroom was the 2D-local / tile family, where the same kind of search moved the board frontier from kd^2/n ~3.3 to 8.0. So tagging each family "saturated vs has headroom" is the high-value annotation.

2. The main trap is surrogate-distance collapse. Low-trial RIS hands back inflated distance upper bounds, an optimizer happily chases them (Goodhart), and they fall apart under deeper search. I had a [[390,8,<=26]] drop to 24 and a [[1080,60,<=116]] drop to 18 once verified properly. So any auto-research candidate needs a deep re-verification pass before it counts as a find. This one passed that pass, which is why I trust the 12.

3. On research prompts lowering the barrier: the new Open challenges panel on the board is a first cut at exactly that. It states the live bars to beat (the 2d-local gap to the 9.75 tile bar, the high-rate Kasai targets), so a contributor can point their LLM straight at a stated target. Per-track "open problems" that an LLM can ingest would be the natural next step.

PR #27, opened 2026-06-26
Harden distance refutation

FarLab:
Refutation fails closed. verify()'s refuter exception handler recorded distance_not_refuted=True (a silent PASS); a gate that fails open is no gate. It now records a FAILURE with "failing closed -- manual review required". The refuter is pure Python and deterministic, so an error there is a real tooling problem, not something to wave through.

Weekly board sweep -- verify/refute_board.py + .github/workflows/refute-weekly.yml. Re-runs the RIS refutation over every codes/ + examples/ entry with a FRESH RANDOM seed each run (Mondays + on-demand). The per-PR gate (gate_changed.py) stays deterministic so it is reproducible; this job varies the seed on purpose, so over time it accumulates search coverage and can catch an over-claim that any single fixed-seed run missed -- the safety net for claims that never paid the per-PR gate (baselines, direct commits, pre-gate codes) and for the back catalogue as the refuter improves. Fails closed on per-code errors. A failing run prints the seed so it can be reproduced (--seed N).

Deliberately NOT done: refuting the whole board on every PR. That is expensive (~4 min) and, because the gate is deterministic, just re-confirms the same pass. verify_all.py stays structural-only (~0.4 s); per-PR refutation stays in gate_changed.py.

New test_refute_gate.py covers both: an over-claim is accepted by refute=False and rejected by refute=True, and a simulated refuter error fails closed.

### DIFF
--- a/.github/workflows/refute-weekly.yml
+++ b/.github/workflows/refute-weekly.yml
@@ -1,0 +1,21 @@
+name: refute-weekly
+
+# Whole-board distance refutation with a fresh RANDOM seed each run. The per-PR
+# gate (verify.yml -> gate_changed.py) is deterministic; this weekly job varies
+# the seed on purpose, so over time it accumulates search coverage and can catch
+# an over-claim that any single fixed-seed run missed. Runs on a schedule and
+# on-demand.
+
+on:
+  schedule:
+    - cron: "0 6 * * 1"      # every Monday 06:00 UTC
+  workflow_dispatch: {}        # allow manual runs
+
+jobs:
+  refute-board:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+      - name: refute every board entry (random seed)
+        run: uv run --frozen python verify/refute_board.py
--- a/site/build.py
+++ b/site/build.py
@@ -736,7 +736,7 @@ def load_entries():
         slug = os.path.splitext(os.path.basename(p))[0]
         with open(p) as f:
             doc = json.load(f)
-        rep = verify(doc)
+        rep = verify(doc)   # site render: structural checks only, refutation is a CI/cron job
         if not rep["ok"]:
             continue
         cert = cert_info(slug)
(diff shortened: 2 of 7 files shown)


vprusso:
Reviewed. The fail-closed change is the important fix and it is correct: the refuter exception handler used to record `distance_not_refuted=True` (a silent pass), and a gate that fails open is not a gate. It now records a failure with "failing closed, manual review required". Good catch. The weekly randomized sweep is a sound idea too: a fixed-seed gate can be tuned against, and re-running with fresh seeds accumulates coverage over time. The new gate test passes.

One thing to reconcile before merge. The description says the per-PR gate stays deterministic while only the weekly job randomizes, but as written the per-PR gate is also non-deterministic: `gate_changed.py` draws a random seed when none is passed (`secrets.randbelow`), and `verify.yml` calls it without `--seed`. The code comment even notes "a merge can fail later". So a contributor's PR can go green and then flip red on a re-run or on merge, which is rough for a community board.

Two ways to resolve, pick one:

- Make the per-PR gate deterministic as the description says: have `verify.yml` pass a fixed `--seed` (or default `gate_changed.py` to a fixed seed), and let only `refute-weekly.yml` randomize. Per-PR CI stays reproducible and you still get accumulating coverage from the weekly job.
- Or keep the per-PR gate random on purpose, but then update the description and CONTRIBUTING so contributors know a green check is not permanent, and maybe run a couple of seeds per PR to cut the flake rate.

My vote is the first: deterministic per-PR, random weekly. It matches the stated design and avoids flaky required checks.

Also needs a rebase: it touches `build.py` and `qldpc_verify.py`, both of which moved on main since this branched, so it currently conflicts. Happy to take the deterministic-per-PR tweak plus the rebase if you want me to land it.

FarLab:
I think its fine to keep the random seed, so a submission can't try to optimize for the particular deterministic seed used. If it passes the CI when opening a PR but fails the merge, that is not a bad thing and that means we can remove that code afterwards. 

Github says we can merge automatically, there is no conflict it seems.

PR #34, opened 2026-06-26
README: link uv's install docs instead of duplicating install steps

vprusso:
Spelling out how to install uv in our README only goes stale as uv changes. State that uv is required and link its install docs; uv run handles the environment after that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,15 +31,7 @@ frontier rather than one winner. See `TRACKS.md`.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


vprusso:
Superseded by #35, which already made this exact change (link uv's install docs) along with the rest of the README trim. Merging this now would be a no-op: the branch's new Installation text is identical to what is already on main. Closing rather than merging an empty duplicate.

PR #86, opened 2026-07-01
Require earned distance witnesses

### LINKED ISSUE #83: Require earned distance witnesses before accepting or scoring submissions
A submission should not pass verification or appear on the board with only `distance.d`.

Today the schema requires only `distance.d`, while `distance.X` and `distance.Z` are optional. The verifier only records an earned global distance when side witnesses validate. That leaves a direct integrity gap: a malformed or hostile submission can claim an arbitrary `distance.d` without proving even an upper-bound witness, pass structural verification, and be scored/rendered as `d <= claimed`.

This should be closed at both the schema/verifier boundary and the site boundary.

Acceptance criteria:

- Require both `distance.X` and `distance.Z` for normal code submissions, or otherwise make the verifier explicitly fail any submission without an earned global distance.
- `verify()` should record a failing check when `earned_distance.d` cannot be established.
- `site/build.py` should refuse to score or render an entry without `rep["earned_distance"]["d"]`, even if `rep["ok"]` somehow remains true.
- Add a regression test: replacing a valid code distance block with `{ "d": 99 }` must fail verification and must not produce a board entry.

Related: #72 tracks broader submission integrity and distance overclaim defenses, but this is a more basic missing-witness acceptance bug.

bachase:
Closes #83.

## Summary

- Require both distance.X and distance.Z witness blocks in the submission schema.
- Make the verifier earn global d only when both side witnesses validate and distance.d matches min(dX, dZ).
- Make the site score entries from earned_distance.d and skip entries without an earned distance.
- Add adversarial regressions for bare distance.d claims and invalid one-sided distance earning.
- Make exact certification fail exactness when either side is missing.

## Verification

- uv run python verify/test_verifier.py
- uv run python verify/verify_all.py
- uv run python verify/test_refute_gate.py
- uv run python site/build.py
- uv run python research/test_smoke.py
- uv run python -m py_compile verify/qldpc_verify.py verify/test_verifier.py verify/certify.py site/build.py

### DIFF
--- a/schema/SCHEMA.md
+++ b/schema/SCHEMA.md
@@ -34,9 +34,8 @@ Two principles drive the format:
   of checks; each check is the sorted list of distinct qubit indices (0-based,
   `< n`) it acts on. So `H_X` has `len(checks.X)` rows.
 - `distance.d`: claimed code distance, must equal the minimum over the
-  provided sides.
-- `distance.X`, `distance.Z` (each optional, but at least the minimizing side
-  is needed to certify `d`):
+  earned X and Z side distances.
+- `distance.X`, `distance.Z` (both required):
   - `value`: claimed minimum weight of a nontrivial logical of that type.
   - `confidence`: `"upper_bound"` or `"exact"`.
   - `witness`: support of a logical operator of that Pauli type and weight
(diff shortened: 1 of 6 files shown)


FarLab:
@bachase got some merge conflicts after I merged PR #88.

bachase:
Will do.

@FarLab this was also the one I was most unsure about the impact; that is whether this is too high a bar for submitters even though it closes a loophole.

FarLab:
Thanks @bachase, I think it looks good and sharp to find this loophole!

PR #113, opened 2026-07-07
CI gate: accelerated deep-refutation pass for frontier claims

FarLab:
## Summary

Follow-up to #112, giving CI the "better verification in the same time" it enables. When the gf2_fast extension is built (CI now builds it, best-effort, in both checkouts), frontier-advancing submissions are gated by a rebalanced deep battery:

- **1 python RIS seed** (unchanged budget: 25k+120n trials, ≤240s) — the manifest-pinned, audited floor, and an in-production canary: it would expose a false-negative bug in the extension by finding what the fast pass missed.
- **The syndrome-decoder cross-check** (unchanged) — mechanism diversity.
- **A gf2_fast witness-returning RIS pass at 150× the python trial target, capped at 8M trials** — the wall-clock freed by dropping 2 of the 3 python seeds, reinvested.

Without the extension (build failure, missing compiler), the full pre-accelerator battery of **3 python seeds runs unchanged** — a broken build can never leave the gate shallower than it was. Dominated (non-record) submissions are unaffected either way.

## Soundness

The accelerator only *proposes* `(weight, side, support)`. A find counts as a refutation only after the pinned python stack (`gf2.py`) validates the witness: kernel membership, anticommutation with a logical, exact weight. Invalid proposals are discarded. False positives are therefore impossible; false negatives are covered by the retained python seed and the parity tests (which now run in CI, since the workflow builds the extension before the self-test step).

Reproducible: fixed `threads=4`, so a verdict reproduces from the printed seed alone. Trust: for code PRs the gate runs from the trusted checkout, so the extension is compiled from **base-branch source** — a submission cannot alter the C++ it is gated by. The 5-file validator manifest is untouched (`gate_changed.py` is base-branch-trusted, not manifest-pinned).

## Why

The [[294,8,20]] over-claim needed ~60k RIS trials to expose; the previous deep gate missed it 3/3 runs. At gf2_fast depth that logical falls in seconds. The same search depth (2M trials/side) refuted 5 of 7 staged n≥300 candidates this week that had passed the 8k-trial packaging gate.

## Measured

| Gate | Search volume (n=320 record) | Wall-clock |
|---|---|---|
| Old deep gate (3 python seeds) | ~190k trials | ~12 min worst case |
| This PR (1 python seed + fast) | 63.4k python + 8M fast trials | **10m42s measured** |

Same time, ~42× the search volume.

- `_fast_refute` refutes the known-inflated [[320,6,30]] candidate with a validated weight-28 witness in 24.5s (500k trials).
- End-to-end deep gate on the merged [[320,6,28]] record: `ok … (deep, 1 RIS seeds x 63400 trials (<=240s each) + fast x 8000000)`, exit 0.
- `test_refute_gate.py` and the quick suite pass; fallback tiers verified: `_budget(320, deep, fast=False)` → 3 seeds, `fast=True` → 1 seed.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -40,6 +40,20 @@ jobs:
           fetch-depth: 0
           path: trusted
       - uses: astral-sh/setup-uv@v5
+      # Build the optional RIS accelerator in each tree that runs searches. Best
+      # effort (continue-on-error): the python passes are the gate's baseline and
+      # run unchanged without it, so a failed build only costs the extra depth.
+      # For code PRs the gate runs from the trusted tree, so the extension is
+      # compiled from BASE-BRANCH source -- a PR cannot alter the C++ it is gated by.
+      - name: build optional refutation accelerator (trusted tree)
+        if: github.event_name == 'pull_request'
+        working-directory: trusted
+        continue-on-error: true
+        run: make fast
+      - name: build optional refutation accelerator (submitted tree)
+        working-directory: submitted
+        continue-on-error: true
+        run: make fast
       - name: detect changed code submissions
         if: github.event_name == 'pull_request'
         id: changes
--- a/verify/gate_changed.py
+++ b/verify/gate_changed.py
@@ -1,14 +1,16 @@
 """Distance-refutation gate for the codes changed in a PR (CI).
 
-Runs independent bounded, fixed-seed refutation searches -- RIS
+Runs independent bounded, fixed-seed refutation searches -- python RIS
 (heuristic_distance), the syndrome decoder (decode/distance, needs ldpc), and,
-for frontier-advancing claims, a ~60x-deeper accelerated RIS pass (gf2_fast,
+for frontier-advancing claims, a ~150x-deeper accelerated RIS pass (gf2_fast,
 built via `make fast`) whose finds only count after the pinned python stack
 validates the witness -- only on the code/example submissions changed in this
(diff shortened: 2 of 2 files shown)


FarLab:
End-to-end deep gate run completed locally on the merged [[320,6,28]] record (all four mechanisms):

```
refutation seed = 424242  (reproduce: python verify/gate_changed.py --seed 424242 <files>)

ok       codes/320-6-28.json (deep, 3 RIS seeds x 63400 trials (<=240s each) + fast x 3804000): no logical lighter than 28
```

Exit 0. The fast pass ran 3,804,000 trials (60× the 63,400-trial python target, under the 4M cap) and agreed with the python seeds and the syndrome decoder: nothing lighter than the claimed d=28 — consistent with the 2M-trials/side convergence study that corroborated this code.

FarLab:
Rebalanced per review discussion: with the accelerator present, deep claims now get **1 python seed + a 150×/8M-trial fast pass** (was 3 seeds + 60×/4M); without it, the original 3-seed battery is unchanged. Timed end-to-end on the [[320,6,28]] record:

```
refutation seed = 424242  (reproduce: python verify/gate_changed.py --seed 424242 <files>)

ok       codes/320-6-28.json (deep, 1 RIS seeds x 63400 trials (<=240s each) + fast x 8000000): no logical lighter than 28

1806.12s user, 281% cpu, 10:41.94 total
```

10m42s vs the old gate's ~12 min worst case, with ~42× the search volume. The retained python seed is the audited floor and the in-production canary against extension false negatives.

PR #120, opened 2026-07-07
UI/UX cleanup: UF brand typography, hover cascades, sharper corners

seunomonije:
UI fixes done with an LLM. Can run with

```
make build
python3 -m http.server 8642 --directory docs
open http://localhost:8642
```


FarLab (commented):
Reviewed the diff plus a rebuild-from-branch experiment (`make build` on the PR head) and some archaeology on main. Overall: **this PR matters more than a style pass — it fixes a live production bug — but it should merge in reduced form, with two small UI regressions addressed.**

## The headline: this PR fixes a live bug on the site

PR #106 committed literal git conflict markers into `site/build.py`'s CSS template string (`site/build.py:427-437` on main: `<<<<<<< HEAD … >>>>>>> df2c121`). Because they sit inside a valid Python string literal, the build runs green and every rebuild since July 6 has stamped them into all 50 generated pages — they are on the live site right now (view-source of the leaderboard, ~line 125). Browsers skip the garbled CSS lines, so the `.how` card styling silently degrades. This PR's `build.py` resolves those markers. That makes it worth merging promptly.

## Blocking: drop `docs/*` from the PR (46 of the 47 files)

The PR is `CONFLICTING` with main solely because it commits regenerated `docs/*.html`. The site workflow (`.github/workflows/site.yml`) triggers on `site/**`, rebuilds, and commits `docs/` itself on merge — a `build.py`-only PR self-deploys and has zero conflict surface. The committed docs are also already stale two ways: rebuilding on the PR head dirties `docs/index.html` and one code page (they were generated from an intermediate commit, before the final colgroup change), and the pages for the three codes merged since branching ([[168,20,14]], [[240,13,15]], [[216,15,11]]) are missing. Rebase + `git rm` the docs changes and the conflicts vanish.

## Two real regressions in the new CSS

1. **Hover cascade makes text transiently unreadable** (`header.hero p a` / `.topnav a`): the yellow fill sweeps left→right positionally over .35s, but `color` transitions *uniformly* (.25s, delayed .1s). So on hover-in the leading text is yellow-on-yellow (invisible) for the first ~100ms, and the trailing text darkens against the still-dark unfilled background mid-sweep — the comment says the delay keeps text "readable while the fill arrives," but a uniform color can't stay readable against a positional wipe. Either clip the text color with the same gradient (`background-clip:text`) or drop the delay/sweep for the instant swap.
2. **Chip ellipsis hides information with no recovery** (`.tchip` + `chips()`): the new `overflow:hidden;text-overflow:ellipsis` clips longer family labels ("generalized bicycle", "hypergraph product") at desktop widths in the `table-layout:fixed` board, and the chips' `title` attributes are generic category descriptions, not the label — so the truncated text is unreadable by any means. Cheap fix: set `title` to the label text.

## Suggested (non-blocking)

- **Guard against #106 recurring**: no test renders the templates, which is why committed conflict markers sailed through CI. A tiny convention-discovered `verify/test_*.py` asserting no `<<<<<<<`/`>>>>>>>` lines in `site/build.py` (and in one rendered page) closes the class of bug.
- `font-family:'Space Mono',…` is now duplicated ~20× with two different fallback stacks (`.mono` has 'SF Mono',Menlo); a `--mono` custom property would give one source of truth.
- The detail-page stagger (six selectors + `backwards` fill + reduced-motion reset) could be a single container fade with no visible difference.

Verified clean, for the record: colgroup widths sum to 100% and match the 11 headers; the sort JS keys on `data-c` (immune to width changes); f-string brace doubling is balanced in all new blocks; `prefers-reduced-motion` covers both new animations; no conflict markers survive in the PR head; the `nth-of-type` delay ladder maps correctly to the four detail-page sections.

🤖 Review generated with [Claude Code](https://claude.com/claude-code) (Claude Fable 5), requested by @FarLab

vprusso:
Thanks for putting this together. After reviewing, we're going to pass on merging this one, so I'm closing it. Reasoning:

**The core UF branding is already in `main`.** The typography (Manrope H1/H2, Space Mono for CTAs, header numbers, and table/container headers) and the link/button hover states landed in #103–#107, and the "QEC Challenge" naming plus the type-tag overflow fix landed in #114. So the Space Mono additions and the chip-overflow handling here largely duplicate what's already shipped.

**The corner-sharpening runs counter to the brand feedback.** The `qec_brand_feedback.pdf` review explicitly called the softer, curved corners "acceptable" and noted they align with the established UF Survey branding, so reducing every `border-radius` to 2–4px is undoing something the brand review signed off on rather than fixing a flagged item.

**The hover cascades and page fade-in animations weren't part of the requested scope.** They replace the simpler, already-merged hover states with motion effects that are a separate design decision we'd want to make deliberately, not fold in here.

There are a couple of genuinely new bits worth keeping (Space Mono on the modal headers and detail-page parameter values, and `:hover` states on the modal close and copy buttons). If we want those, we'll pull them into a small, verified change separately.

Also, for the record: this branch collided with an unrelated bug in `main`, where commit 43fa232 had left Git conflict markers inside the CSS string (leaking into every generated page). That's now fixed independently in #126.

PR #128, opened 2026-07-08
Fix FAQ question entity rendering

bachase:
Summary: Decode FAQ question text before escaping so pre-encoded HTML entities do not render literally, then regenerate docs/faq.html. Tests: uv run python site/build.py, git diff --check, browser render check on local faq.html.


bachase:
These were rendering oddly on chrome osx

FarLab (approved):

PR #154, opened 2026-07-14
Corroborate [[254,44,21]] upper-bound distance

vprusso:
Verification follow-up for the current efficiency leader [[254,44,21]] (#152): adds the `certs/heuristic/` corroboration cert, upgrading it from plain `d ≤` to the corroborated `d ≤*` tier.

Evidence in the cert, produced by `verify/heuristic_certify.py`:

- RIS at 400k trials: lightest logical exactly 21, nothing lighter
- Independent BP+OSD syndrome decoder at 250k trials: agrees, nothing lighter
- Combined verdict: corroborated

This is on top of the evidence already in the submission (1M-trial ladder, 4 fresh adversarial seeds x 1.5M trials, CI's 20-minute deep refutation gate). Exact certification (d =) via the MILP is out of reach at n = 254; it already times out at n = 120, so corroborated is the strongest achievable tier for this entry.

The board rebuild will pick up the tier change on the next site build after merge.


FarLab:
Interesting, I think all these recent new code submissions should be corroborated upper bounds (they all pass the deep RIS and BP+OSD checks). The normal upper bound vs corroborated is a remnant from early implementations. We should have just one upper bound tier which is the corroborated one.

PR #156, opened 2026-07-14
Collapse the corroborated tier into the upper bound

### LINKED ISSUE #155: Upper bound vs corroborated upper bound
I think all these recent new code submissions should be corroborated upper bounds (they all pass the deep RIS and BP+OSD checks). Or we should just call it an upper bound. The normal upper bound vs corroborated is a remnant from early implementations. We should have just one upper bound tier which is the corroborated one.

FarLab:
Fixes #155.

The board showed three distance-confidence tiers, but the middle one (`d<=*` corroborated) stopped signaling anything: only 5 of 50 codes ever got a heuristic cert (nothing automates `verify/heuristic_certify.py`), while since #94/#113 every merged code faces the deep per-PR gate (adaptive RIS up to 8M accelerated trials + BP+OSD cross-check) and the weekly fresh-seed board sweep — strictly more scrutiny than the cert recorded. A regular upper bound on this board already means what "corroborated" used to claim.

Changes (display/docs only — no code entries touched):

- `site/build.py`: the tier is now `exact` or `ub`; the `d<=*` badge, amber color, chart series, and legend entries are gone. The `d<=` badge tooltip, board legend, and FAQ now state the refutation-tested meaning explicitly.
- `TRACKS.md`: the Layer-3 tier list merges the corroborated description into the upper bound.
- `verify/heuristic_certify.py`: docstring updated — it stays as a manual deep-check tool; `certs/heuristic/` stays as historical evidence.

The built `docs/` pages are intentionally not included; the site workflow rebuilds and commits them on merge.

Verified: `site/build.py` builds clean (60 detail pages, no `corroborated`/`d<=*` remnants in output — remaining mentions in a few detail pages are provenance-notes text from the code JSONs), `run_tests.py` passes 6/6, and the trusted-stack integrity pin is untouched.

### DIFF
--- a/TRACKS.md
+++ b/TRACKS.md
@@ -55,9 +55,10 @@ Distance confidence is orthogonal to the tracks:
 
 - `d<=` (upper bound): a submission exhibits an explicit logical operator of the
   claimed weight, and the verifier confirms it is a genuine nontrivial logical,
-  so the distance is at most that weight.
-- `d<=*` (corroborated): an upper bound that an independent search has tried and
-  failed to beat. Evidence, not a proof.
+  so the distance is at most that weight. The claim is also refutation-tested:
+  independent searches (the deep RIS + BP+OSD gate at PR time, and weekly
+  fresh-seed sweeps of the whole board) try and fail to beat it. Evidence, not
+  a proof.
 - `d=` (exact): a server-side integer program has proven no lighter logical
   exists. This is NP-hard and does not scale, so large codes carry tight upper
   bounds while small and moderate codes are certified exact. The board shows an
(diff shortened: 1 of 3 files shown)


vprusso (approved):

vprusso:
Checked this against today's entries ([[254,44,21]] #152, its heuristic cert #154, and [[254,44,24]] #157): no conflicts, and the change reads correctly for them. Both went through the post-#133 deep gate, so collapsing the tier loses nothing for these entries; the #154 cert staying under certs/heuristic/ as historical evidence is the right call. Their provenance notes mention "corroborated" as free text, which you already account for. One nit: the body says 5 of 50 codes carry heuristic certs; with #154 merged it is now 6, not that it changes the argument.

Issue #158, opened 2026-07-14
The uncapped weight-9plus cell turns the headline kd^2/n into a check-weight treadmill


vprusso:
Today's campaign surfaced a board-design consequence worth deciding on deliberately rather than by accretion.

## The data

Within the weight-9plus x unrestricted cell, raising GB support weight converts check weight into distance almost mechanically. All of these are divisor-design cyclic GB codes at nearly the same n:

| code | supports | w | kd^2/n |
|---|---|---|---|
| [[258,32,22]] | 5+5 | 10 | 60.03 |
| [[254,44,21]] | 6+6 | 12 | 76.39 |
| [[254,44,24]] | 8+8 | 16 | 99.78 |

The last two share the same n, k, and divisor family; the only change is support weight. Nothing in the current taxonomy stops 10+10 supports (w=20) next, and there is no reason to expect the trend to flatten soon. The BPT-style motivation for kd^2/n (TRACKS.md, FAQ) is that it is bounded for bounded-weight codes, so an unbounded weight class quietly removes the boundedness that makes the number meaningful, and the "best kd^2/n on the board" headline tile inherits that.

The per-cell Pareto frontier handles this correctly (w is an axis; lighter-check codes keep their records). It is specifically the headline scalar and any cross-weight reading of the efficiency column that degrade.

## Options

1. Band the top class: split weight-9plus into weight-10, weight-12, weight-16, ... cells (nesting like the existing classes). Records stay per-band; the treadmill becomes explicit competition per band.
2. Cap the headline: keep the cell as is, but restrict the "best kd^2/n" tile (and any global efficiency claim) to codes at or below some weight, e.g. w <= 8 where the published enumerations live.
3. Weight-normalized efficiency: report kd^2/n alongside w and sort the headline by a weight-adjusted figure. More design work, probably overkill.

Option 1 or 2 (or both) seems cheapest. Related but orthogonal to #156 (distance-tier collapse); this is about the weight axis, not the confidence axis.

willzeng:
Good points here @vprusso. Broadly this seems to be to come down to how reasonable it is to have very high weight codes. They certainly are hard to engineer, but this challenge could help determine how valuable it is to have high weight.

Of the three options you suggest, I like #3. Weight-normalized efficiency.

An alternative could be to figure out how to put the weight into the target metric? Is there a clear way for how to do that? E.g. kd^2/(nw)?

@FarLab

FarLab:
[...]

PR #159, opened 2026-07-14
Add record-progress chart: running best kd^2/n per weight class

vprusso:
Adds an ecdsa.fail-style progress chart to the index, between the stat tiles and the primary tracks: the running best kd²/n per check-weight class, stepping through the board's record events.

Stacked on #156 (branched from its head so the two don't conflict): merge #156 first and this diff reduces to the chart alone.

## Design decisions

- One step line per weight class (w ≤ 6, w ≤ 8, any) rather than a single global line, so the chart never reads as cross-weight comparison. The classes nest, and the divergence of the "any w" line makes the check-weight-for-distance trade of #158 visible instead of hidden inside one number.
- The x axis is record events in date order, not elapsed time: the literature seeds span 1996 to 2024 while challenge entries land weekly, so a linear time axis would crush the challenge era into a sliver. Year markers keep the timeline legible, and the subtitle states the spacing convention.
- Log y axis (the running best spans 1.3 to 99.8), ticks at 1/2/5/10/20/50/100.
- Each record is a dot: hovering names the code (weight, efficiency, date), clicking opens its detail page, and the existing cross-highlight wiring picks it up (hovering a dot highlights the code's table row, and vice versa).
- The series palette (amber/blue/violet) passes the lightness-band, chroma, CVD-separation, and contrast checks; identity is carried by the legend and direct end labels, not color alone.
- On phones the panel scrolls horizontally like the tracks grid, instead of scaling the text away.

Verified with desktop and mobile screenshots; the story reads correctly: slow literature ramp, dense challenge-era staircase, and the top line ending at the current 99.78 leader.


willzeng (approved):
@vprusso Looks great! This is an excellent addition. I added a y-axis label but otherwise will merge. 

Image

PR #222, opened 2026-07-17
Add [[1524,766,14]] pair-partition CPM CSS baseline (arXiv:2607.14091)

MathysRennela:
Literature baseline from Okada & Kasai, arXiv:2607.14091v1 (15 Jul 2026). Pair-partition CPM-based CSS code, J=4 column weight, L=14 row weight, prime lift P=271, girth 6. Distance d=14 certified by explicit non-stabilizer witnesses on both sides (regenerated with the paper's own C++ verifier against the reconstructed exponents). Advances the weight-9plus x unrestricted board. Passes the full trustless verifier (passed: true).


FarLab:
Did this get cancelled automatically?

MathysRennela:
> Did this get cancelled automatically?

Yes it got cancelled after a timeout.

PR #223, opened 2026-07-17
Add [[3122,1788,16]] pair-partition CPM CSS baseline (arXiv:2607.14091)

MathysRennela:
Literature baseline from Okada & Kasai, arXiv:2607.14091v1 (15 Jul 2026). Pair-partition CPM-based CSS code, J=4 column weight, L=16 row weight, prime lift P=489, girth 6. Distance d=16 certified by explicit non-stabilizer witnesses on both sides (regenerated with the paper's own C++ verifier against the reconstructed exponents). Advances the weight-9plus x unrestricted board. Passes the full trustless verifier (passed: true).


FarLab:
This might take forever haha

MathysRennela:
> This might take forever haha

is there a reason not to do it? (like extra cost for github actions)

FarLab:
I think its fine doing it once, but we may have to think about when many people open PRs with such big codes.

Issue #168, opened 2026-07-15
Locality track: a theorem-backed score f(n,k,d,w,D) with certified embeddings


FarLab:
Follow-up to #165, turning the discussion there into a concrete challenge design: use the BPT tradeoff itself as the score.

## The inequality

For a stabilizer code with qubits on a D-dimensional grid (one qubit per site, n sites) whose stabilizer generators are **range-w local** (each supported in a w×…×w box):

**k · d^(2/(D−1)) ≤ 64 · g(D) · w^(2+2/(D−1)) · n**,  with **g(D) = D(D−1)·(4D)^(2/(D−1))**.

This is Bravyi–Poulin–Terhal (arXiv:0909.5200) with the w- and D-dependence made explicit; the constant comes from an elementary, entropy-free derivation via the Bravyi–Terhal cleaning lemma (arXiv:0810.1983) that we have written up with all constants tracked. The **exponents are provably optimal** (see the tiling construction in the comments below; also implicit in Dai–Li arXiv:2409.15203 and Dai–Li–Tang arXiv:2503.22651). The multiplicative constant — including the D-shape g(D) — is **not**: the true ceiling is only known to lie between ~ρδ^(2/(D−1)) (flat in D, from tiled good codes) and 64·g(D) (~D²), and the D² comes from our proof's cubical bookkeeping, not from any matching construction.

## The score

The score contains only what is proven tight — the exponents — plus a fixed display constant:

**f = 16 · k · d^(2/(D−1)) / ( w^(2+2/(D−1)) · n )**.

The normalization makes the **rotated surface code score exactly f = 1** (n = d², k = 1, range 2) at every size; the challenge reads: **maximize f; any score above 1 beats the surface code.** Unlike kd²/n, f is scale-free (a code family converges to a constant instead of inflating with n), and the exponent-saturating tiled-good-code family scores 16·ρδ^(2/(D−1)) — roughly **flat across dimensions**, which is the calibration a cross-D score should have. (It beats the surface code iff ρδ² > 1/16 in 2D — a concrete open target.)

The theorem is then quoted separately as the **proven ceiling**: f ≤ 1024·g(D), i.e. 2¹⁷ at D=2. The ceiling is loose and D-dependent; it — not the score — is what gets versioned ("ceiling v1"). Future tightening of the constant changes the quoted ceiling and never any published score, in any dimension.

[...]

PR #234, opened 2026-07-17
Weight-aware figure of merit kd²/(n·w²) on the board (#165)

### LINKED ISSUE #165: Proposal: a weight-aware global figure of merit alongside per-cell kd$^2$/n
## Motivation

The board ranks codes by kd²/n inside (locality × weight-class) cells, and the new record-progress chart (#159) is explicit that lines are "compared within a class, not across." That's the right default — but it means we have no single number that says whether a weight-10 code at kd²/n = 60 is doing more or less with its resources than a weight-6 code at 14. Low check weight is the resource the whole challenge is about; a global figure of merit should charge for it.

## What BPT actually says about w

The natural normalization comes from Bravyi–Poulin–Terhal, but the paper never states the w-dependence explicitly — Eq. (3) is k ≤ cn/d² with c "depending only on locality of the projectors and dimension of the Hilbert space." Tracing the proof, for 2D-local quantum codes:

**kd²/n ≤ c₀ · w⁴ · log₂ q**

The w⁴ enters in two places:

- **w² from the entropy step (Eq. 14):** k ≤ |C|·log₂ q where C is the set of corner cutout regions separating diagonally-adjacent A-blocks. Each corner must have linear size ~w so no w×w projector straddles two A-blocks, and there are ~n/R² corners, so |C| ~ w²·n/R² (the paper writes "|C| ~ n/R²", silently absorbing the w²).
- **w² from the correctability step:** the buffer around an R×R block is two layers of thickness w, |B_C| = c·wR, and |B_C| ≥ d forces R ≥ d/(cw); substituting R² ≥ d²/(cw)² into n/R² squares that w.

In D dimensions (Eq. 4) the same steps give c(w) ~ w^(2 + 2/(D−1)) — w⁴ at D=2, w³ at D=3, → w² as D → ∞. The classical 2D bound (Eq. 5) is only *linear* in w. **Caveat:** this exponent is extracted from the proof, not stated in the paper — worth an independent check before we bake it into the site.

## Proposal

Add a weight-aware saturation ratio, displayed alongside (not replacing) the per-cell kd²/n:

- **2D-local tracks:** kd²/(n·w⁴) — up to the absolute constant, this is "fraction of the BPT ceiling achieved," and it is directly comparable across weight classes within the track. (log₂ q = 1 for us; everything on the board is qubit-based.)
- **Unrestricted-locality codes:** there is no BPT ceiling (good qLDPC codes make kd²/n unbounded), so any exponent is a convention, not a theorem. Options: reuse w⁴ for consistency, pick an exponent empirically, or only define the global metric where the bound actually applies and leave unrestricted codes on plain kd²/n.

Implementation is small: `site/build.py` already computes `eff`; this adds one more computed column, an optional chart series, and a legend/FAQ note. Records, cells, and frontiers stay exactly as they are — display-only at first, with a separate discussion before any re-ranking.

## Open questions

1. Is w⁴ right? Someone should independently re-trace the proof (or find a reference that states c(w) explicitly).
2. Max check weight vs. average check weight in the denominator? BPT's argument uses the max, but the average may be the fairer resource measure.
3. Should the global metric ever drive a global frontier/records, or stay informational?

vprusso:
Implements the consensus reached in #165. After Farrokh ran the cheap dimension estimators over the board and found that every high-rate code is an expander with no fixed geometric dimension, the thread converged (Farrokh, Will, Mathys) on a single dimension-free convention: penalize check weight by w², the D→∞ limit of the Bravyi-Poulin-Terhal exponent w^(2+2/(D-1)) (w⁴ at D=2, w³ at D=3).

## What this adds

- A computed `effw = kd²/(n·w²)` per entry.
- A sortable kd²/nw² column on the board table, between kd²/n and w.
- The same figure on each code detail page.
- An `effw` search filter (e.g. `effw>=0.5`).
- An FAQ entry explaining the metric and why w² is the honest exponent for expander codes.

## What this deliberately does not do

Display-only, per the proposal. Records, per-cell frontiers, and the record-progress chart still rank by kd²/n. No re-ranking here; that was flagged in the issue as a separate discussion.

## Effect

It reshuffles rather than rescales, exactly as Farrokh reported: on the current board [[390,68,28]] (w=13, effw 0.809) overtakes [[390,82,30]] (w=16, effw 0.739), and the w=10 [[258,32,22]] pulls even with the w=16 [[390,82,27]].

The BPT w-exponent is extracted from the proof rather than stated in the paper (open question 1 in the issue); the FAQ and column tooltip present w² as a convention, which is why this stays display-only for now.

Closes #165.


FarLab:
Thanks for this PR @vprusso. There is still some theory behind this score function I am trying to understand: as mentioned in #168 , the distance d also has an exponent D which would make the dependence on d disappear as D goes to infinity. One approach to remedy this is to use the (slightly) more complicated score function f (defined in #168), but then we need to compute/estimate (D, w) from the submitted code or a layout should be supplied. The submitter is incentivized to supply a layout since the heuristic estimates for (D,w) usually underestimate the score f. So this could be one approach. I am still experimenting with this by doing analysis on the current leaderboard (which has many codes so can do useful data analysis).

vprusso:
You're right, and it's an inconsistency in the metric, not just a framing quibble. kd²/(n·w²) keeps d² (the d exponent at D=2) but takes w² (the w exponent as D grows). Those are the same 2/(D−1) factor in the tradeoff, so pairing them isn't a limit of anything: push D to infinity consistently and 2/(D−1) goes to 0, which kills the d dependence too and leaves k/(n·w²), which no one wants.

So I don't think this column should be sold as the dimension-free BPT ratio. I've reworded its FAQ and tooltips on this PR to call it what it is: a heuristic weight penalty on kd²/n, kept only because it needs no layout and still charges for check weight, with no claim to dimensional consistency. It now points at f (#168) as the consistent quantity.

That leaves two honest ways to go, and I'd defer to your leaderboard analysis on which:

- keep this column as an optional, clearly-labeled heuristic sort key next to kd²/n, or
- drop it once f covers enough of the board.

On the incentive you mention: f (stages 1 and 2 of #168) already computes (D, w) from a layout for the 24 embedded codes, and a heuristic embedding search under-scores relative to a supplied layout, so submitters are indeed better off providing their own. Happy to hold this PR display-only until your data lands.

vprusso:
Closing in favor of the split settled in #235: keep kd²/n as the incumbent-relative score (qubit savings versus surface-code patches at matched k and d), and use f (#168) for the ceiling-relative one. The kd²/(n·w²) penalty here muddies the first without adding what f already gives.

PR #235, opened 2026-07-17
Locality-track score f: foundation (#168, stage 1)

vprusso:
First stage of the theorem-backed locality score from #168. Stacked on #234 (the #165 weight-aware metric) so the diff shows only the #168 changes; I will retarget this to main once #234 merges.

You picked the full track build. That is genuinely a multi-PR effort with some research-grade pieces (the range-w convention, the constants-tracked derivation, embedding search), so I am staging it and checking in between stages rather than landing it all at once. This is stage 1 of 6: the score foundation.

## The score

f = 16 · k · d^(2/(D−1)) / ( w^(2+2/(D−1)) · n ), normalized so the rotated surface code scores exactly 1; f > 1 beats it. The exponents are the provably-optimal Bravyi-Poulin-Terhal exponents (tight via tiled good codes), so f is a genuine saturation ratio against a proven ceiling, not a convention like kd²/n. The loose proven ceiling f ≤ 1024·g(D) = 2¹⁷ at D=2 is quoted separately and versioned (ceiling v1).

The range w here is the geometric interaction range (box of sites a stabilizer fits in), measured from layout coordinates and pinned by the surface-code calibration (a weight-4 plaquette spans one lattice step per axis, giving w=2). This is a different quantity from the algebraic check weight used in #165; the two are not interchangeable.

## What is in this PR

- `verify/locality_score.py`: the score, g(D), the versioned ceiling, and the box-range measurement. Self-calibrates the surface code to f=1 at every distance.
- `verify/qldpc_verify.py`: computes box_range and the score in the report; None for unrestricted/expander codes, where f is undefined.
- `site/build.py`: shows f (with D, range w, and proven ceiling) on the detail page of every embedded code, plus an FAQ entry.
- `docs/locality-score.md`: the definition, the exponent-optimality argument, the versioned ceiling, and the range-w convention, with citations.
- `TRACKS.md` pointer, and `verify/test_locality_score.py`.

## Result on the current board

All 24 certified 2D-local codes now score. The best, [[392,8,15]] planar BB, reaches f=0.907; none exceed the surface code yet, consistent with the open challenge in the issue (exhibit a 2D-local family with f > 1).

## Deliberately deferred to later stages

f* over multiple certified embeddings; a CI heuristic embedding-search fallback (D=2, then D=3); D=3 support; obstruction audits (spectral/separator upper bounds, flagging expander codes); and the track-leaderboard integration plus the layout-improvement credit convention. Stage 1 is display-only on detail pages and does not re-rank any track.

Relates to #168.


FarLab:
What is really interesting about this score is that the codes from https://arxiv.org/pdf/2504.08887 do NOT beat the surface code, contrary to what they claim about how their codes are an order of magnitude more efficient. The kd^2/n is not the right metric to compare these kinds of codes. Worth researching this more!

vprusso:
The mechanism is clean: f = (kd²/n)·16/w⁴, so at range w a code beats the surface code exactly when kd²/n > w⁴/16. At range 3 that threshold is 5.06, and the best planar code on the board, [[392,8,15]], sits at kd²/n = 4.59, landing at f = 0.91, about 10% short.

The inversion is sharper than that. The kd²/n leaders among the 2D-local codes are the f losers: [[294,12,14]] has the highest kd²/n on the board (8.0) but range 7, so f = 0.05, and [[263,16,12]] has kd²/n = 8.76 at range 5, so f = 0.22. f and kd²/n nearly reverse the ranking, because the high-kd²/n codes buy their efficiency with long-range checks and f charges w⁴ for that.

I'd frame it a little differently from "kd²/n is the wrong metric," though. kd²/n measures rate at fixed physical-qubit count, which is a legitimate axis, and the planar codes do win it; their "order of magnitude" is a rate claim and it stands. f measures something else, how tightly a code saturates the BPT bound for the range it actually uses, and there the surface code is the reference. The two disagree because kd²/n ignores the range budget and f prices it in. So f doesn't refute their efficiency claim, it reframes it: they get the rate by spending range.

One caveat before leaning on the conclusion: these f values use the submitted layouts, and w is only as small as the best certified embedding. A weight-6 bilayer check can in principle fit a 2×2 footprint (4 sites × 2 layers = 8 slots ≥ 6), i.e. range 2 rather than 3. If [[392,8,15]] admits a range-2 embedding its f jumps to 4.59 and it does beat the surface code. So "the planar codes don't beat the surface code" is right now a statement about the layouts we have, not about the codes. Settling it means finding tighter embeddings (the f* and embedding-search stages here) or a lower bound showing range 3 is forced. That's the open question.

FarLab:
"The two disagree because kd²/n ignores the range budget and f prices it in. So f doesn't refute their efficiency claim, it reframes it: they get the rate by spending range." Yes, exactly this. They beat the surface code because they pay with range.

FarLab:
A follow-up on the "kd²/n is not the right metric" point above, after a literature pass on why everyone uses it anyway: kd²/n is the standard FoM across the BB/qLDPC literature — the gross code's headline "FOM = 12", qudit BB codes advertising kd²/n = 20 (arXiv:2606.17268), and it is the direct optimization target in [qcode-discovery](https://github.com/qiskit-community/qcode-discovery) and the LLM-guided BB search (arXiv:2606.02418) — always cited as "motivated by the BPT bound," yet applied to expander codes where BPT does not bind. The rationalization that actually survives without BPT is **baseline-relative**: k rotated surface-code patches at distance d cost n_SC = kd² data qubits, so kd²/n = n_SC(k,d)/n is simply the qubit-savings factor versus building the same (k, d) out of the incumbent architecture. That reading is purely operational — no geometry, no theorem — and it explains the metric's behavior: it has no ceiling (a ratio against a fixed baseline has no reason to be bounded), and the d² exponent belongs to the *baseline*, not the code being scored (papers that benchmark against the repetition code use kd/n instead, arXiv:2603.22532). So the clean division of labor for the board: **kd²/n is the incumbent-relative score** ("how many fewer qubits than surface patches at matched k and d" — legitimate on any hardware, but the comparison concedes range for free, which is exactly the "they pay with range" point above), while **f is the ceiling-relative score** (saturation of a proven bound at the range actually used). Neither refutes the other; they price different resources against different references.

FarLab:
So according to Fable, the way people use kd^2/n is actually not really motivated by the BPT bound, because in the expander case (D to infty) it wouldn't make any sense: the BPT bound only works for fixed dimension D. I guess it started as a FOM using the BPT bound with D=2 and it stayed. There is another motivation for why we should use kd^2/n (baseline-relative score), so for simplicity we should just go for that. We are seeing kd^2/n getting larger and larger, but that is not necessarily a bad thing! @vprusso @willzeng @MathysRennela @seunomonije

vprusso:
Agreed, and the baseline-relative reading is the one that survives without BPT. So kd²/n = incumbent-relative (qubit savings vs surface patches), f = ceiling-relative, both kept for different questions.

Given that, I'm closing #234. A w² penalty bolted onto a savings ratio is more confusing than useful, and f already charges for weight the principled way when a layout exists.

MathysRennela:
> So according to Fable, the way people use kd^2/n is actually not really motivated by the BPT bound, because in the expander case (D to infty) it wouldn't make any sense: the BPT bound only works for fixed dimension D. I guess it started as a FOM using the BPT bound with D=2 and it stayed. There is another motivation for why we should use kd^2/n (baseline-relative score), so for simplicity we should just go for that. We are seeing kd^2/n getting larger and larger, but that is not necessarily a bad thing! @vprusso @willzeng @MathysRennela @seunomonije

I agree with @FarLab, this is not really a bad thing. My reasoning behind penalizing high weight is that w>=9 is not very realistic for the foreseeable future.

vprusso:
Closing the loop here since the metric discussion finished elsewhere: kd²/n stays as the incumbent-relative headline (#165), weight realism is handled by the weight-class cells plus the w<=32 cap that landed in #251, and K_G was dropped as a column.

That leaves this PR's scope question. Per @FarLab's comment on #256, f is not production-ready as a score for expander codes (no natural finite D). What this PR ships is the machinery for the case that IS well-defined: certified layouts, measured range, f on the 24 2D-local codes, plus the f*-over-embeddings plumbing. Options: merge it as the research substrate with f shown only on embedded codes' detail pages (current behavior), strip the display and merge just the verifier computation, or park the PR until the dimension question settles. Branch is rebased on main and green; @FarLab's call.

FarLab:
closing this as we already have implemented a form of this.

PR #236, opened 2026-07-17
Add [[281,121,4]] hypergraph product of eham16 × eham16

MathysRennela:
Hypergraph product construction (arXiv:0903.0566). CI runs the verifier: CSS / k / weight / distance witnesses / locality. model: hy3.


FarLab:
Hi @MathysRennela, are these codes directly from the paper or did you use techniques from that paper to get the codes? Just want to make sure I understand these. I also opened an issue #254 to make this more clear: when someone submits a code directly from a paper, we want the authors of the paper to be in the "authors" field, but we also want to credit the submitter of the code, so another field called "submitted_by" would be needed in such cases.

MathysRennela:
> Hi @MathysRennela, are these codes directly from the paper or did you use techniques from that paper to get the codes? Just want to make sure I understand these. I also opened an issue #254 to make this more clear: when someone submits a code directly from a paper, we want the authors of the paper to be in the "authors" field, but we also want to credit the submitter of the code, so another field called "submitted_by" would be needed in such cases.

The codes are my construction, but they are a direct application of the construction of the paper. I can't claim novelty: I just took their technique and applied it to existing codes on the board to discover new codes.

FarLab:
I see, that makes it your codes then!

PR #224, opened 2026-07-17
Add [[684,8,85]] 2BGA affine (q=19) — corrected from refuted d=100

MathysRennela:
## Summary

Corrected submission for the `[[684,8]]` 2BGA over the affine group of F_19.

**This replaces the refuted `codes/684-8-100.json` with `codes/684-8-85.json`.**

The original `d=100` claim (this PR's first version) was **REFUTED by CI**:
- syndrome-decoder pass found a **weight-97** X-logical
- accelerated RIS-fast pass (8M trials) found a **weight-85** X-logical

Both are lighter than the claimed 100, so the honest upper bound is `d <= 85`.

### Honest parameters
- **n=684, k=8, d≤85** (upper bound; `d_X≤85`, `d_Z≤104`)
- **Locality:** weight-8, unrestricted
- **Construction:** 2BGA over the affine group Aff(F_19), supports a=[294,208,142,284], b=[6,323,76,217] (row weight 8)
- **References:** arXiv:2601.08824, arXiv:2604.20838
- **Author:** @MathysRennela (model: Hy3)

### Witnesses (both validated against the pinned python stack)
- X-logical, weight 85 (RIS-fast witness): in ker(H_Z), nontrivial (not a stabilizer product)
- Z-logical, weight 104 (original witness): in ker(H_X), nontrivial

`d = min(d_X, d_Z) = 85`. Verified by `verify/validate_candidate.py`: passed, not refuted.

> Note: this code does not advance its board cell (a `[[684,8,104]]`-class unrestricted entry already exists), so it is a correction of the record rather than a new board advance. The claim is now sound and cannot be refuted by the gate's deepest search (8M RIS-fast trials), which already found the weight-85 logical.

Issue #249, opened 2026-07-17
Cap blocklength at n ≤ 500 (verification-budget rule)


FarLab:
Proposal: add \`"maximum": 500\` for \`n\` in the code schema (currently 5000), with a validator message and a CONTRIBUTING note.

**Rationale — evidence quality, not just CI time.** The adaptive gate's trials scale linearly with n but wall-clock is capped (240 s/seed), so large codes pass with relatively weaker evidence per qubit. Our track record shows the gate already thins out near the top of the current range: the 2026-07-06 convergence checks refuted 5 of 7 gate-passed candidates at n ≥ 300 on 2M-trial deep reruns, and exact certification (MILP) scales worse than the sampler. Most recently, PR #222 ([[1524,766,14]], n = 1524) had verification time out in CI — a claim that size is beyond what the pipeline can currently stand behind at all.

**Side effect:** with n capped, every board quantity is bounded — kd²/n included (Singleton gives ≈ 9,400 at n = 500; the constructible frontier is far lower) — so the score can't be inflated by scale alone, and the board is honestly a finite-length benchmark.

**Versioning:** raise-only. The cap states what the verification machinery can vouch for today (all 62 board codes are n ≤ 390); if gf2_fast/MILP tooling improves, we raise it — backward-compatible by construction.

@willzeng @vprusso @MathysRennela

MathysRennela:
Since #244 passed the CI gate, I would suggest putting the limit at n=1000 until we figure out how to:
1. Run more efficient tests (if that is even possible)
2. Define a better score metric

I am afraid that limiting n to 500 right now might lead us to a biased data set.

MathysRennela:
Following some investigations, I also think we should cap the check weight at 32. 
Beyond that, checking the validity of codes is not reasonable; and weight-32 is already unrealistic for near term platform but it provides a playground for people to find more reasonable codes by studying their structures.

@FarLab @vprusso

PR #251, opened 2026-07-17
Blocklength and check-weight caps (issue #249)

vprusso:
Implements #249: two verification-budget caps.

- `n` at 700: what the machinery has demonstrated (the gate handled n=684 in CI on #224, including catching its d=100 overclaim). Keeps every baseline and the pending n=684 submission in scope; larger codes go to the playground (#257). Started at 500, moved per the thread (FarLab 600, measured capacity 700).
- Check weight at 32: beyond this, validating a claim is not practical, and weight 32 is already beyond near-term hardware. Everything on the board is w <= 16; the cap lands at the weight-32 frontier #260 opened.

Both raise-only, both stated in the validator message, CONTRIBUTING, and SCHEMA.md. Trusted-stack manifest re-pinned.


FarLab:
You are right, we have a few codes with n>500. Maybe cap n<=600 so we dont remove codes?

seunomonije:
Just adding this for the record, but an experimental "playground" version with huge code sizes may be something to explore purely for the fun of it

vprusso:
Bumped the cap to 600 and kept the three baselines (530/568/590). Nothing is removed now.

@seunomonije agreed, a separate playground board with no cap would be a good home for the big ones (the [[1524,766,14]] from #222, etc.) without the verification promise attached. Out of scope for this PR, but worth its own issue.

MathysRennela:
I am actually about to fix #224 and it feels to me that it is reasonable to have codes up to n=1000.

MathysRennela:
I closed PR #222 & #223 because their n is way above 1000.

vprusso:
Settled on 700: Mathys measured the gate holding to n=700 on #224, and it caught the d=100 overclaim there in CI, so that is the demonstrated capacity, not a guess. Keeps all baselines and #224 in scope. Above 700 goes to the playground (#257). Raise-only as tooling improves.

vprusso:
Also caps check weight at 32 now, per https://github.com/unitaryfoundation/qldpc-challenge/issues/249#issuecomment-5015680395. Everything on the board is w<=16; the cap lands exactly at the weight-32 frontier #260 opened.

MathysRennela (approved):

FarLab (approved):
LGTM

Issue #254, opened 2026-07-18
Provenance: separate discovery credit from submission credit


FarLab:
The \`provenance.authors\` field currently does two jobs: discovery credit (baselines correctly carry paper authors) and submission credit (submissions carry contributor handles). Those diverge for literature-derived submissions, where a submitter's handle would misattribute discovery.

**Sweep result (2026-07-18, Claude Fable 5 + web verification):** all 32 novelty-unflagged submissions plus the two \`novelty: unknown\` codes were checked against the literature — the planar-BB batch against every parameter triple in arXiv:2504.08887 (+ arXiv:2412.04181), the GB batch against Panteleev–Kalachev / Wang–Pryadko tables and QEC-pages data, the 2BGA batch against the Lin–Pryadko exhaustive enumeration (arXiv:2306.16400) and the coset paper (arXiv:2606.17268). **Zero verbatim matches — every entry is a new instance of a published family.** So the board is currently clean; this proposal is structural, before the problem occurs. Two upgrades fell out: [[254,44,21]]/[[254,44,24]] can move \`unknown\` → \`new_parameters\` (nearest Z_127 lit: PK [[254,28]], UB [[254,14,14]]), and [[180,18,14]]/[[180,20,14]] exceed the *exhaustive* plain-2BGA optimum at order 90 (d = 7) — the coset construction provably left the enumerated family.

**Proposal:**
1. \`authors\` = who constructed/discovered the code. For a verbatim literature code, the paper's authors — never the submitter.
2. New \`submitted_by\` (handle, required, displayed on the site) — submission credit always survives, so the incentive does too.
3. New \`reference\` (arXiv/DOI), required when \`novelty: known_parameters\`; encouraged always (family paper). Currently references only live in free-text \`construction\`.
4. Require a \`novelty\` declaration on new submissions (CI checks field presence; correctness stays a PR-review norm, since the verifier can't check novelty).

Convention for the gray zone: verbatim re-encode → paper authors + \`submitted_by\`; new instance of a published family → submitter is the author, family paper as \`reference\`; modified/improved → submitter as author, original as \`reference\`, delta narrated in \`construction\`. Retroactively: fill \`novelty\` on the 32 unflagged entries (evidence above); no author rewrites needed.

@willzeng @vprusso @MathysRennela @seunomonije

MathysRennela:
currently the scripts to submit codes do not properly record models.

MathysRennela:
[...]

Issue #253, opened 2026-07-18
The board as a dataset: five findings from a first data-analysis pass


FarLab:
With 62 verified codes in one schema, the board is now a dataset. I ran a first analysis pass (analysis: Claude Fable 5); posting the results as a showcase to encourage more of this — none of it is a proposal, all of it took minutes.

1. **Distance is ~92% predictable without a decoder.** A 5-variable linear model (log n, rate, weight, avg degree, spectral gap λ₂) predicts log d with leave-one-out R² = 0.92. Rate is the dominant cost (strong negative coefficient); λ₂ enters positively — a direct expansion→distance link in the data. The residuals are useful: over-performing *claims* are natural deep-verification priorities, and the model is a free triage prior for candidate sweeps. Notably [[390,82,27]] over-performs its prediction (d=27 vs 19.8) — the record holder is structurally anomalous, not just big.

2. **Half the board is Pareto-dominated.** Only 31/62 codes survive on the (n,k,d,w) frontier. Nearly every dominated 2D-local code is beaten by a non-local one — the locality tax in its bluntest form (the tracks add that axis back). A few within-class dominations exist too (e.g. [[180,18,14]] and [[180,20,14]] both by [[168,20,14]]) — maybe worth a "superseded" flag on the site.

3. **The Steane code is still the most Singleton-efficient object on the board** (S = (k+2d−2)/n = 0.71; nothing else above 0.5 except tiny codes). The modern records sit at S ≈ 0.35 with a balanced k-vs-2d budget split (~0.52) — k-heavier than the kd²/n-optimal 2/3, because constructible distance is the scarce resource.

4. **The frontier's distance scaling is d ~ n^0.84** (Pareto codes, k ≥ 6) — well above √n at n ≤ 400, with a five-code d ≥ 1.2·√n club led by [[320,6,28]] (d = 1.57·√n). Envelope over families, not one family's scaling — but the board has constructively left the surface-code regime.

5. **Board archaeology:** 36/62 codes carry a model attribution, but 9 of the top-10 by kd²/n are AI-assisted and 10/10 are GB/2BGA-family. Human/literature codes hold the locality and small-n territory; the sweeps own the efficiency frontier.

If you have a question the dataset could answer (girth vs distance? symmetry vs efficiency? decoder thresholds vs λ₂?), post it here — most of these take minutes to run.

MathysRennela:
I am toying around with new models so I fed this data analysis to a panel. 
[...]

Issue #164, opened 2026-07-15
Weight penalization on leaderboard


MathysRennela:
> Good points here @vprusso. Broadly this seems to be to come down to how reasonable it is to have very high weight codes. They certainly are hard to engineer, but this challenge could help determine how valuable it is to have high weight.
> 
> Of the three options you suggest, I like #3. Weight-normalized efficiency.
> 
> An alternative could be to figure out how to put the weight into the target metric? Is there a clear way for how to do that? E.g. kd^2/(nw)?
> 
> @FarLab  

 _Originally posted by @willzeng in [#158](https://github.com/unitaryfoundation/qldpc-challenge/issues/158#issuecomment-4980817550)_

My proposal to fix this issue: add a weight-penalized ratio, and double the leaderboard with a simulation-backed LER board.

**In short:** `kd²/n` is the right metric for what this challenge can *verify* (trustless, cheap, structural), but it is the wrong metric for what a code is *for*. We could: keep `kd²/n`, add a weight-penalized companion scalar (`kd²/(n·w)`) that is computable today, and add a separate simulation-backed board for sub-threshold logical error rate.

**Why:** `kd²/n`  rewards block size, not code quality (it scales like $n^2$). It also ignores check weight, which drives up decoder cost, and it says nothing about thresholds, the actual metric that determines whether we want to use a code or not. For example, the surface code has good threshold with `kd²/n ≈ 1`.

**What I propose:**

1. **Keep `kd²/n`** since the existing Pareto frontier is over `(n↓, k↑, d↑, w↓)`.

2. **Add `kd²/(n·w)`** to the board output, where `w` is the average check weight (or the maximum check weight for a conservative variant). Easy to compute as is.

3. **Add a second, optional, maintainer-run board** ranking by **`p_L(p*)` with a fixed decoder** (e.g. BP+OSD, or a specified local decoder) at `p* ∈ {10⁻³, 10⁻⁴}`. It is computationally expensive but I think it is worth it. The most important part is to report `(L, N)` with a posterior lower bound so under-simulated codes can't falsely win.

**Scope of the changes needed:**

- TRACKS.md — document `kd²/(n·w)` and the `p_L` board, and the caveat that `kd²/n` is a structural proxy only.
- build.py — emit `kd²/(n·w)` alongside `kd²/n` in `stats.json` / docs.
- schema — optionally add the `p_L` result as an optional, maintainer-attested field (not self-declared, to avoid gaming).

**Open questions:**
[...]

Issue #127, opened 2026-07-08
Correlated sweeps in distance estimate (fix: one seed per code)


MathysRennela:
TLDR: Distance estimates are biased across a sweep, because they share the same seed; each code needs its own seed based on its fingerprint.

## Problem
`research/kit/search.py:73` passes the same seed to `distance_rand` for every candidate in a sweep:

```python
d = distance_rand(HX, HZ, trials=trials, seed=seed)
```

This means that distance upper bounds are correlated: structurally similar code get identical draws, which biases the ranking and the Pareto frontier. This may mean that a code can look worse/better purely as an artifact of the correlated seed.

## Solution

Give an independent draw to each code, from its fingerprint.
Line `research/kit/search.py:73`  becomes for example: 

```python
d = distance_rand(HX, HZ, trials=trials, seed=seed+int(fp,16))
```

The `seed` argument is kept as the base for the fingerprint hash, so runs stay reproducible.

PR #256, opened 2026-07-18
AIQxQIA 2026 full-paper draft

vprusso:
Draft submission for AIQxQIA 2026 (https://aiqxqia2026.cnr.it/), full-paper format per the instructions for authors: CEUR-WS ceurart template, single column, single-blind with authors named, body over 4 pages plus references (currently ~5.3 body pages, 6 with references). The template class and its dependencies are vendored in `talks/aiqxqia-2026/` so the folder drops into Overleaf or EasyChair as-is. Compiled PDF included.

Adapted from the whitepaper at a higher level, per Mathys: repository file names and CI implementation details are gone, and the framing targets the workshop topic "AI techniques for quantum error correction." Structure: design principles, the cell/Pareto board, the two figures of merit honestly labeled (incumbent-relative kd²/n per the #165/#168 discussion, plus the ceiling-relative locality score f), the three-tier distance verification, the trust architecture for machine submissions, and a new operational-lessons section with the refutation-depth data (5/7 staged candidates refuted pre-submission; 11/11 candidates reading d=31-34 at 8M trials all collapsing to d≤30 at 80M×3; the [[684,8]] d=100 claim refuted to ≤85 in CI).

Before submission (deadline July 31):
- Confirm the author list. I included Vincent, Mathys, Farrokh, Seun, and Will; the email thread also has Ben and Brad, whose surnames/affiliations I could not confirm, so they are a TODO in the source.
- Add ORCIDs and emails; update the `\conference` line when the workshop announces dates and location.
- The board statistics (124 codes, 29 exact, best eff 189.2) are as of today and should be refreshed at submission time.


MathysRennela:
I read the paper entirely. I think it is good as is and I really liked the inclusion of the story at the end about [[684,8]]. It shows how the board catches potential over-claims.
Main caveat for me: I think we should settle issues #51, #127, #158, #164, #165, #168, #249, #253, #254, and #257 as they have an impact on the final form of the challenge.

vprusso:
Triaged the ten. Most are further along than the open count suggests:

- #158/#164/#165: settled in-thread (kd²/n stays incumbent-relative, weight penalty dropped with #234, f is the ceiling-relative score). Can be closed with a summary line each.
- #249: PR #251 at n<=700, green, needs FarLab's review.
- #168: stages 1-2 are PR #235, green, reviewable now.
- #257: consensus plus Mathys's submitter restriction; needs building.
- #127: small fix, unclaimed.
- #51: 29 codes certified exact; the rest is bounded by the cap discussion.
- #253, #254: genuinely open discussions.

Critical path for the paper: two reviews (#235, #251), one build (#257), and settling #254. None change the paper's text beyond the board numbers.

talks/aiqxqia-2026/paper.tex
MathysRennela, line 1: Should we add Sebastian and Seun as authors?

vprusso: Yes, I just ported the ones over from the abstract, but let's add them here too.

FarLab:
I like the paper too, but one thing I would remove is the score function f in section 4. The reason is that it is a bit more complicated and needs more research on how to implement it in a score function besides kd^2/n. The reason is that for expander-like codes, you have to pick a dimension D in which you want to embed it and compute the corresponding range w (this is different from check weight!). A natural choice is D=2, so we can compare with 2d local codes, but in those cases it is definitely going to lose as these codes are suited for all-to-all connectivity devices (D=infty), but in that case the formula doesn't really work. I think it is worth researching it more, but at the moment it is not "production ready" as a score. Maybe worth mentioning, but not implementing it yet. What do you think?

vprusso:
Agreed and done. Section 4 now mentions the ceiling-relative score as a public research direction, no formula, no board numbers, and names exactly the obstruction you describe: it needs an embedding dimension and a measured range, and expander codes built for all-to-all connectivity have no natural finite D. The f>1 open question is rephrased without the symbol. kd²/n stays the only implemented score in the paper.

vprusso:
Merging the paper as-is, but others should feel free to make edits to the paper ad they see fit. Note that Jul 31 is the deadline for the conference.

PR #261, opened 2026-07-19
adding a second rule to store witnesses

MathysRennela:
I took some time to evaluate the efficiency of non-frontier open source models (Hy3, Mimo-V2.5, Gemma 4, Nemotron 3, Laguna M.1, Laguna XS.1, North Mini Code). Most of them waste time in re-emitting witnesses because they do not properly store them. Adding a second rule that makes storing witnesses non-negotiable saves a lot of time in autosearch, and is valuable for participants who do not want to use frontier models.


FarLab:
Great idea!

FarLab (approved):

PR #279, opened 2026-07-24
Add geometric efficiency score + headline update (issue #276)

### LINKED ISSUE #276: Add "geometric efficiency" score + update headline on the website
We will call kd^2/n the "operational efficiency" (against the surface code baseline).

We will call the (more complicated) f(n,k,d,D,r) the "geometric efficiency" as that is based on a geometry that accompanies the code (layout in dimension D -> gives range r). A submission should contain a layout that optimizes (D,r). Most of the cases D=2 and the layout optimization is finding the lowest range r. This is another axis to improve on.

If no layout is given because it is an expander code, we will not compute f, but only kd^2/n.

Definitions and disclaimers should be clearly visible when hovering over the score headlines (on top of the page).

Remove "certified exact" headline number from that row and add "geometric efficiency" in place.

Add a toggle on the top right of the chart for "geometric codes" (with layout given) and expander codes (no layout).

@willzeng @MathysRennela @seunomonije Did I miss anything? Comment below.

FarLab:
Implements #276 with the conventions from the issue thread ([comment](https://github.com/unitaryfoundation/qldpc-challenge/issues/276#issuecomment-5068860668)). Touches `site/build.py` and `TRACKS.md` only; `docs/` regeneration is left to the site workflow on merge.

## The score

**f = 4kd²/(nρ²r⁴)** — the BPT ratio priced by the verified layout: r = measured Euclidean interaction radius (units of the unit qubit spacing), ρ = layers.

- Constant **4** = (√2)⁴, so the rotated surface code is exactly **1** (the 16 from the working note belongs to the box-side convention and would put the surface code at 4).
- **ρ² layer charge**: the unique exponent making re-packaging score-neutral (folding halves the area within reach, gains 4 in r⁴, pays 4 in ρ²) — layers must earn their density.
- Computed **only from a verifier-accepted layout**; no layout → no f, labeled "no layout" (a certification status — not "expander", which nobody verified).
- **d ≥ 5 headline guard**: f is scale-free, so a [[4,2,2]] plaquette scores 2; constant-distance tilings are barred from the headline (f-note §6).
- **Distance-tier inheritance**: f from an upper-bound d is shown as ≤ in the headline and flagged in tooltips.
- **D = 2 scope**: schema stores planar coordinates; D > 2 (different BPT exponents) reserved.

## Site changes

- Statsbar: "certified exact" card → **best geometric efficiency** (currently reads **1** — surface code, certified exact — i.e. the reference is on the board and the challenge is to beat it at growing d). kd²/n card relabeled **best operational efficiency**. Both carry full hover definitions/disclaimers.
- Chart toggle top-right of the plots: **with layout / no layout**, composing with all existing filters.
- Table: sortable **f** column (searchable as `f>=0.1`, `with-layout` / `no-layout` terms), per-cell tooltips with r and ρ.
- Detail pages: f and r cells in the parameter grid.
- `stats.json`: `best_geometric_efficiency` for the README badge path.
- TRACKS.md: "Two efficiency scores" section documenting both scores, conventions, and guards (caps stay coarse eligibility gates; f prices range continuously — a trivial √n-range layout is priced into irrelevance, f ~ n⁻²).

## Validation

Site built locally on merged main: headline/toggle/column/detail pages verified, all four emitted script blocks parse (node syntax check), self-tests 6/6. No `codes/` files touched (scope check clean).

Closes #276.

https://claude.ai/code/session_01ANsG3sNTjV7GZq3K8ipzAe

### DIFF
--- a/TRACKS.md
+++ b/TRACKS.md
@@ -120,9 +120,12 @@ surface code to 1, but they answer different questions.
     "expander code".
   - g inherits the distance tier: computed from an upper-bound d it is an
     upper bound, and the site marks it so.
-  - The headline card requires d >= 5: g is scale-free, so constant-distance
-    tilings (a [[4,2,2]] block on one plaquette scores f = 2) would "beat"
-    the surface code without encoding anything asymptotic.
+  - The headline card requires d >= 3: g is scale-free, so d = 2 tilings
+    (a [[4,2,2]] block on one plaquette scores g = 2) would "beat" the
+    surface code without encoding anything asymptotic. No known d = 3-4
+    code with an honest layout comes near 1, but no theorem caps that band
+    either; the threshold will be raised (or replaced by the d_min(w, rho)
+    rule of the working notes) if a small-d packing exploit materializes.
   - D = 2 only for now: the schema stores planar coordinates. The D > 2
     generalization changes the exponents (kd^(2/(D-1)) per the BPT bound)
     and is reserved.
--- a/TRACKS.md
+++ b/TRACKS.md
@@ -92,6 +92,43 @@ checks span about 5.83 and open-boundary corner stabilizers reach about 6.71,
 both constant in n. The verifier also reports layout diagnostics (interaction
 radius, qubits per site, minimum spacing, density, bounding box).
 
+## Two efficiency scores
+
+The site shows two figures of merit (issue #276); both normalize the rotated
+surface code to 1, but they answer different questions.
+
+- **Operational efficiency** `kd^2/n`: the Bravyi-Poulin-Terhal saturation
+  ratio, incumbent-relative. Bounded and meaningful for 2D-local /
+  bounded-weight codes at comparable n; grows with n for high-rate codes, so
+  it is compared within tracks.
+- **Geometric efficiency** `f = 4 k d^2 / (n rho^2 r^4)`: the same ratio priced
+  by the layout the code ships with. `r` is the measured interaction radius
+  (max check diameter, Euclidean, in units of the unit qubit spacing), `rho`
+  the layer count. The constant 4 = (sqrt 2)^4 puts the surface code
+  (r = sqrt 2, rho = 1) at exactly 1. The `rho^2` factor is the capacity
+  charge: it is the unique exponent that makes re-packaging a layout into
+  layers score-neutral (folding halves the area within reach, gaining
+  (sqrt 2)^4 = 4 in r^4, and pays exactly 4 in rho^2), so layers are neither
+  free nor forbidden -- they must earn their density.
+
+  Conventions and guards:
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Approve. Reviewed from a clean checkout; I wrote the earlier f prototype (#235), so I checked the convention closely.

Math is exactly right. The board's rotated surface codes [[25,1,5]], [[49,1,7]], [[81,1,9]] all score f = 1.000 (r = sqrt2, rho = 1), so the constant 4 = (sqrt2)^4 normalizes the reference on the board itself. The other topological entries fall where they should: color [[7,1,3]] at 0.32, toric at 0.125 (periodic, larger radius), and the rho^2 layer charge visibly bites on the bilayer planar codes. Self-consistent throughout.

Mechanics: builds clean, tests pass, both headline cards render (geometric 1, operational 303.61), and the f column, with-layout/no-layout toggle, and searchable f>= are in place.

The conventions are the careful part and they hold up: Euclidean interaction_radius (which the verifier already measures, so no new trusted input), the repackaging-neutral rho^2 exponent, the d>=5 headline guard against constant-distance tilings scoring >1 trivially, "no layout" as a certification status rather than the unverified "expander" label, D=2 scope, and <= inheritance from upper-bound distances. This is a cleaner version of the prototype and it resolves the not-production-ready objections from the paper thread by scoping f to exactly the case where it is well-defined.

It does add a second headline number, a departure from "one number," but that is the #276 decision this implements, not a review question. No blockers from me.

willzeng:
Great to see it! Have a couple questions to discuss about this before merge. Added a few to the issue

willzeng (changes requested):
Thanks for the comments. This makes sense and is close! A few small improvement suggestions and one question

Are all these generalized bicycles listed without an f score because they are missing layouts or is it something fundamental?
image

Would propose that replace `f` with `g` for geometric efficiency in the button and the column label:
image

Should have definitions of the operational and geometric efficiencies between these headline numbers and the chart section:
image
Would be good to write out what all of k, d, n, etc are here so that the parameters are clear.

FarLab:
For the generalized BB question: they are missing layouts indeed. If a layout is present, it will automatically compute the geometric score. In separate PR's, I can start searching for layouts for these codes. I am not sure yet if they are expander like or not. Even if they are expander like, we can still embed them in D=2 and see how they score :)

willzeng (approved):
Got it. Good to merge from me then. I think we should start doing some searches and fill in the layouts and G scores for ones in the table before we make it public. What do you think @FarLab ?

That isn't blocking for this PR though

FarLab:
Sounds good, I'll go ahead and merge it. I will start some autoresearch loops too to improve geometric efficieny and fill the board with codes from literature too.

PR #312, opened 2026-07-28
Site: card layout for the board on narrow screens

### LINKED ISSUE #305: Board table: hide/collapse columns on mobile and narrow displays
On mobile and other constrained displays the board table has 11 columns, which overflows badly. Allow hiding or collapsing columns so the table is usable on a narrow screen.

Mathys suggests some columns are redundant (e.g. n, k, d are also encoded in the `[[n,k,d]]` label) and could be hidden. Options: a responsive layout that drops lower-priority columns below a width breakpoint, or a column-visibility toggle. Decide which columns are essential on mobile vs droppable.

Raised by @MathysRennela in https://github.com/unitaryfoundation/qldpc-challenge/issues/299#issuecomment-5094317217

vprusso:
Closes #305. Updated per @willzeng's review.

Rather than drop the `g` column on phones (it is one of the two headline metrics, and hiding it would suggest the challenge ignores locality), the board now switches to a card layout on narrow screens (<=820px), the same pattern ECDSA.fail uses: each row becomes a card with every field shown as a labeled line (type, n, k, d, kd^2/n, g, w, authors, model, date), no horizontal scroll and nothing dropped. The record star moves onto the card title. Each `td` carries a `data-label`; the header row is hidden in card mode. The desktop table is unchanged.

Rendered at 390px (cards) and 1240px (table) and eyeballed.

### DIFF
--- a/site/build.py
+++ b/site/build.py
@@ -701,7 +701,22 @@ border-bottom:1px solid var(--ln);font-size:15px}}
 margin-left:8px}}
 .board td.star{{display:none}}
 .board td.auth{{max-width:none}}
-.board td.typecell{{flex-wrap:wrap}}}}
+.board td.typecell{{flex-wrap:wrap}}
+/* n, k, d are already printed inside the [[n,k,d]] name: drop their rows */
+.board td.col-n,.board td.col-k,.board td.col-d{{display:none}}
+/* date: small, in the card's top-right corner */
+.board tr{{position:relative}}
+.board td.date{{position:absolute;top:12px;right:14px;width:auto;padding:0;
+font-size:11.5px;color:var(--mut)}}
+.board td.date::before{{content:none}}
+.board td.codecell{{padding-right:92px}}
+/* kd2/n, g, w: a horizontal stat trio — big value, small label beneath
+   (column-reverse puts the ::before label under the number) */
+.board td.m3{{display:inline-flex;flex-direction:column-reverse;
+align-items:center;justify-content:flex-start;gap:3px;width:32.8%;
+padding:9px 0 5px;font-size:19px;font-weight:700;
+font-variant-numeric:tabular-nums;text-align:center}}
+.board td.m3::before{{font-size:10.5px}}}}
 /* phones: reclaim horizontal space and shrink oversized headers */
 @media(max-width:560px){{.wrap{{padding:0 14px}}
 header.hero{{padding:34px 0 30px}}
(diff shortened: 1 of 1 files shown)


willzeng (changes requested):
@vprusso I don't think we should drop the g column. It is important to be there so that folks don't dismiss the challenge as not including locality and it is one of the two main metrics to optimize against.

On ECDSA.fail, when you get to a narrow enough window, the rows turn into cards like this: 

Image

I'd propose that we do the same so that we have more space to show all the data. What do you think?

MathysRennela:
> @vprusso I don't think we should drop the g column. It is important to be there so that folks don't dismiss the challenge as not including locality and it is one of the two main metrics to optimize against.
> 
> On ECDSA.fail, when you get to a narrow enough window, the rows turn into cards like this: 
> 
> Image
> 
> I'd propose that we do the same so that we have more space to show all the data. What do you think?

Here's a screenshot from my phone. I think g fits if we get rid of the star or the empty space after the code spec.

image

vprusso:
Good call, I dropped the column-hiding and did the ECDSA.fail card approach instead. On narrow screens (<=820px) each row is now a card showing every field as a labeled line, so g stays visible along with n/k/d, type, authors, model and date, with no horizontal scroll. Record star sits on the card title; desktop table is unchanged. Pushed to this branch.

willzeng:
This is good but I think still a bit busy. for example on mobile it now is: 
image

Would propose that on the card:
- the n, k, and d rows are indeed redunant and can just be dropped 
- the date just goes in the top right liked the ECDSA fail card to save space
- kd^2/n and g and w aren't done vertically as rows but are three parameters horizontal across the card, like with the number larger and the label (e.g. g) in smaller text below

I'll start making a design update along these lines and push it soon

willzeng:
Pushed d136c16 with a round of card polish from local review: dropped the n/k/d rows (redundant with the [[n,k,d]] name), moved the date to the card's top-right corner, and laid out kd²/n | g | w as a horizontal stat trio (big value, small label beneath) instead of stacked rows. All changes are inside the ≤820px media query — desktop table untouched. Happy to adjust or revert any of it if you had different plans for the card @vprusso 

Looks like this: 
image

willzeng (approved):

willzeng:
Actually I simplified it a bit more: 
image

MathysRennela (approved):

PR #316, opened 2026-07-28
Normalize model names on current board

MathysRennela:
Normalizes the model names (previously: Hy3 was disclosed with 4 different variant names).

In the future, we should consider a canonical form for model names (Provider, model name, version number), and set up a warning for submissions with unregistered models. But it's a bit an overkill to set up this machinery right now.

This PR requires to bypass the CI gate as it modifies multiple codes at the same time.


MathysRennela (commented):
hand-checked all files. All changes are +1/-1, at the exception of codes/390-82-38.json which was missing a trailing newline. (+2/-2)

FarLab (approved):

PR #326, opened 2026-07-29
Site: reference links + sorting, score-box redesign, copy tightening

vprusso:
A few site presentation fixes.

**References page** ([`references.html`](https://unitaryfoundation.github.io/qldpc-challenge/references.html))
- Entries sort alphabetically by bibtex key.
- Each key is now a link to its canonical source: journal DOI when there is one, else the arXiv abstract, else the arXiv DOI, else the entry's resource URL. `ref_target` also pulls a URL out of `howpublished`, which is what the Grassl codetables.de entry uses (it previously linked nowhere).
- Added the OmniQEC reference (arXiv:2607.25865).

**Index**
- Redesigned the operational/geometric-efficiency explainer: two cards with the formula highlighted and a separate legend strip, instead of the cramped side-by-side paragraphs.
- Tightened the primary-tracks intro (removed the "one click away" filler and an em dash).

Site generator only (`site/build.py`, `refs.bib`); `docs/` is rebuilt by CI on merge. Rendered references, score box, and primary-tracks copy locally.

### DIFF
--- a/refs.bib
+++ b/refs.bib
@@ -378,3 +378,16 @@
   howpublished  = {\url{https://errorcorrectionzoo.org}},
   note          = {arXiv:2606.11484; the Error Correction Zoo}
 }
+
+@misc{yan2026omniqec,
+  title         = {OmniQEC: Discovering Practical Quantum Error-Correcting
+                   Codes by an {AI} Scientist},
+  author        = {Yan, Ge and Li, Shanchuan and Ma, Pengyue and Zhang, Qixin
+                   and Ma, Pingchuan and Wang, Jianping and Hsieh, Min-Hsiu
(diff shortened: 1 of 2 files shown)


FarLab:
Looks good. The geometric efficiency and its formula fall on different lines. can we make the box a bit wider so its on the same line?
Scherm­afbeelding 2026-07-29 om 16 46 31

vprusso:
Fixed. Raised the score-card grid min to 380px so the cards drop to full-width single-column a bit sooner, which leaves room for the geometric title and its formula on one line. Checked at 710px (your case), 1100px (still two columns), and 390px mobile: formula stays on the title line in all three.

FarLab (approved):
LGTM

PR #327, opened 2026-07-29
Remove non-participant files before soft launch (#325)

### LINKED ISSUE #325: Remove irrelevant files before soft launch
Identify and remove irrelevant files before soft launch.

FarLab:
Part of #325 — a sweep of everything tracked, checking each file against what the site, CI, and docs actually reference.

## Removed

- **`qec_brand_feedback.pdf`** (root, 572 KB): design-feedback material, referenced by nothing; also the largest tracked file.
- **`docs/distance_via_simulation.pdf` + `.tex`** (~300 KB): sat in the served Pages directory but no page links to them (they rode in with an old submission, b9efbd5).

## Untracked (not deleted)

- **`talks/`** (15 files, ~1 MB): conference papers, abstracts, and slides — maintainer material, not something a challenge participant needs; nothing in the site or verification pipeline references it. Removed from the index only: working copies stay on disk for the maintainers, git history keeps the record, and `talks/` is now gitignored so drafts don't land by accident. (The AIQxQIA/QSW sources will move to their own repo.)

## Kept

- **`docs/qec_challenge.pdf` + `.tex`**: the whitepaper is linked and embedded by the site (`site/build.py`), so it stays, source included.
- Everything else tracked is load-bearing: `codes/` + `certs/` (board data), `verify/` (hash-pinned trust anchor), `research/` (starter kit), `notes/` + `fieldnotes/` (rendered on the site), `docs/` HTML (the deployed Pages site), and the root tooling.

Also adds `.DS_Store` to `.gitignore`.

Net: ~1.9 MB and every non-participant-facing file out of the tree.


vprusso (changes requested):
One thing to resolve before this merges: `docs/distance_via_simulation.tex` is not actually orphaned. Two board codes reference it in their notes, `codes/180-18-14.json` and `codes/180-20-14.json`, both contain the line "See docs/distance_via_simulation.tex for the estimation method," and it renders on their code pages. Deleting the file leaves those two notes pointing at nothing, which is the kind of dead reference a pre-launch sweep is meant to prevent.

Either fix works: keep `distance_via_simulation.{tex,pdf}`, or drop that sentence from the two notes in this same PR (a small edit to the `notes` field of each).

The rest holds up:
- `qec_brand_feedback.pdf` is referenced nowhere in the tree. Safe to remove.
- The whitepaper is correctly kept: `site/build.py` links and embeds `qec_challenge.pdf`.
- Untracking `talks/` is safe. Nothing in the site, CI, or docs references `talks/`, #256 (the workshop paper) is already merged, and no other open PR touches `talks/`.

MathysRennela (commented):
Maybe let's find another place to store the papers.

FarLab:
Good catch — my sweep grepped the site and docs but not `codes/`, and the two [[180,*,14]] coset notes do cite the file. Fixed as suggested: dropped the "See docs/distance_via_simulation.tex for the estimation method." sentence from both `notes` fields (the rest of each note is untouched), and re-ran `verify/qldpc_verify.py` on both codes — both pass. The rendered code pages still show the old sentence until the next site rebuild picks up the change post-merge.

@MathysRennela agreed — the plan (also on #325) is to rehome the paper sources in their own repo after the AIQxQIA deadline (July 31); this PR only untracks them, working copies stay put.

Issue #302, opened 2026-07-27
Keep code table and leaderboard visible when scrolling past the plots


vprusso:
When you scroll past the code scatter plots, the code table and the leaderboard slide underneath the plots instead of staying visible. Mathys wants them to stay put (static) rather than be covered.

Screenshot in the source comment.

Note: the plots were made sticky on desktop in #273 (`.explorer .plots{position:sticky;top:0}`), and the board has its own scroll context (`.boardscroll{max-height:64vh;overflow:auto}`). This report suggests the sticky-plots behavior is now hiding the table/leaderboard in some layouts. Need to reconcile: either the plots should not overlap the table, or the sticky region needs a bound so the table is never occluded.

Raised by @MathysRennela in https://github.com/unitaryfoundation/qldpc-challenge/issues/299#issuecomment-5094317217

MathysRennela:
Additional issue (mobile version only): on mobile, the table with all the codes has been replaced with a list of all the score cards; the leaderboard is visible only after scrolling through all of them.

@vprusso @FarLab

FarLab:
@MathysRennela can you make a new issue for this?

PR #339, opened 2026-07-30
Site: collapse the mobile card list behind a Show-all button

### LINKED ISSUE #336: All score cards are displayed in the mobile version of the website
> Additional issue (mobile version only): on mobile, the table with all the codes has been replaced with a list of all the score cards; the leaderboard is visible only after scrolling through all of them.
> 
> @vprusso @FarLab  

 _Originally posted by @MathysRennela in [#302](https://github.com/unitaryfoundation/qldpc-challenge/issues/302#issuecomment-5124173573)_

FarLab:
Fixes #336.

Below the 820px breakpoint the board rows render as stacked cards in normal page flow (#305 / #312), which put ~150 full-width cards between the charts and the contributor leaderboard — on a phone the leaderboard was only reachable after scrolling the entire list.

**The fix**: the mobile card list now shows its first 10 cards plus a **Show all 151 codes** button; tapping it reveals the rest. Desktop is untouched (the bounded scroll box still applies and the button never renders there).

**Why JS, not CSS**: the filter toggles rows via inline `style.display` and sorting re-appends rows to reorder the DOM, so a static `nth-child` cap would swallow search matches and pick the wrong cards after a sort. The cap runs inside the existing `apply()`:

- any active narrowing (search text, range sliders, literature/layout toggles) disables the cap, so every match is always shown;
- the collapse walks current DOM order, so it stays correct after reorders;
- a `matchMedia` change listener re-runs `apply()` when the viewport crosses the breakpoint, so rows hidden by the cap never leak into the desktop table;
- chart dots key off the match set, not the cap, so capped cards stay plotted.

Site generator only (`site/build.py`, +31/−1); `docs/` is rebuilt by CI on merge. Verified in the built output: single `#showall` button with the right count, both CSS rules emitted, inline JS passes `node --check`.

@MathysRennela — this is the mobile half of what you reported; the desktop occlusion half is #335. Would appreciate a check on a real phone viewport.


vprusso (approved):
Resolved the conflict and pushed the merge (this branch is green now). The conflict was a clean add/add: this PR inserts `.showall{display:none}` right after `.boardscroll`, and main had added the `@media(min-width:821px) and (min-height:800px)` explorer-flex rule (#302) at the same spot. Kept both, nothing else touched.

The cap logic is right. Verified in a real viewport:
- mobile (innerWidth 780, matchMedia max-width:820 true): 10 cards visible, `#showall` shown.
- desktop (innerWidth 1400, matchMedia false): all 151 rows visible, `#showall` hidden.

Putting the cap inside `apply()` rather than CSS is the correct call: it survives the inline-display filtering and the sort re-append, any active narrowing disables it so matches past the cap still show, and the matchMedia listener re-runs it across the breakpoint so capped rows don't leak into the desktop table. Chart dots key off the match set, so capped cards stay plotted.

PR #341, opened 2026-07-30
feat: submit [[128,21,8]] eff=10.5 — odd-k 2BGA on C4×Q16

MathysRennela:
## [[128,21,8]] — 2BGA on C4×Q16 (odd k)

**Efficiency**: kd²/n = 21×64/128 = **10.500**

Nonabelian 2BGA code with odd k=21, unreachable by abelian BB constructions. Weight-8 checks (row weight 8).

### How found

Systematic 2BGA sweep over nonabelian groups of order 60–200, filtering for odd k, screened at 400 RIS trials. Distance held at 4,000 trials.

Strategy: look→build→confirm. Targeted odd-k niche as the only construction that can reach Pareto slots abelian methods are structurally locked out of. Four phases explored (weight-4 BB, weight-8 meta/dihedral, odd-k nonabelian, HP/LP/BP products); only odd-k produced board-advancing codes.

### Reproduction

```python
from gap_bridge import all_groups_of_order
from group_algebra import build_2bga
groups = all_groups_of_order(64)
g = groups[59]  # C4xQ16
cayley = np.array(g['cayley_table'], dtype=np.int64)
HX, HZ = build_2bga(cayley, [6,24,37,60], [5,33,44,51])
```

### Board advancement

Advances the **weight-8 × unrestricted** board cell. Current best in that cell was eff=9.0 ([[54,6,9]]); this code reaches eff=10.5.


vprusso (approved):
Verified locally against main's gate: [[128,21,8]] passes, d=8 upper bound on both sides, unrestricted / weight-8. New non-dominated point on the weight-8 frontier (eff 10.5); no board code dominates it and it does not duplicate an existing (n,k). Note included. Low distance, so the PR-time refutation is reliable.

PR #342, opened 2026-07-30
feat: submit [[128,13,10]] eff=10.156 — odd-k 2BGA on C8:Q8

MathysRennela:
## [[128,13,10]] — 2BGA on C₈:Q₈ (odd k)

**Efficiency**: kd²/n = 13×100/128 = **10.156**

Nonabelian 2BGA code with odd k=13, unreachable by abelian BB constructions. Weight-8 checks (row weight 8). Trades k for d vs [[128,21,8]] — lower k (13 vs 21) but higher d (10 vs 8).

### How found

Same search as [[128,21,8]]: systematic 2BGA sweep over nonabelian groups of order 60–200, filtering for odd k, screened at 400 RIS trials. Distance held at 4,000 trials.

Strategy: look→build→confirm. Targeted odd-k niche as the only construction that can reach Pareto slots abelian methods are structurally locked out of.

### Reproduction

```python
from gap_bridge import all_groups_of_order
from group_algebra import build_2bga
groups = all_groups_of_order(64)
g = groups[90]  # C8:Q8, GAP index 182
cayley = np.array(g['cayley_table'], dtype=np.int64)
HX, HZ = build_2bga(cayley, [2,6,17,23], [6,16,37,59])
```

### Board advancement

Advances the **weight-8 × unrestricted** board cell.


vprusso (approved):
Verified locally against main's gate: [[128,13,10]] passes, d=10 both sides. It strictly dominates the board's [[140,10,10]] (lower n, higher k, same d and weight), so a clear frontier advance at eff 10.16. Note included.

PR #380, opened 2026-08-03
Board: race submitted topological codes in the g view, add a geometric-efficiency leaderboard

### LINKED ISSUE #356: Leaderboard for geometric efficiency
Currently, the leaderboard only ranks contributors based on the best operational efficiency $kd^2/n$. 
I propose to add an option to switch to a leaderboard based on the best geometric efficiency.

### LINKED ISSUE #376: Geometric efficiency progress graph only shows exact distance codes
I recently added a code ([[37,7,3]]) that tops the geometric efficiency for non-surface codes but it won't show up in the record progress.

The most likely reason is that I didn't send any exact distance certificate, but the code should show up nonetheless.

FarLab:
Fixes #376. Fixes #356.

Both issues are about the same surface of the board — the geometric efficiency `g` — and both are one-line-ish changes plus a toggle in the same function, so they land together. The one-code-per-PR rule is CI-enforced on `codes/`; this touches only `site/build.py`.

## #376 — the g record chart was hiding submitted colour codes

The g view filtered on `!r.topo`, i.e. **every** code whose family is `topological`. The intent was narrower: hold back the *seeded* surface/toric tilings, which **are** the `g = 1` ceiling and would otherwise freeze the record race at their 1997 publication date. They are already drawn as the dashed reference line.

Keying that on family alone also excluded three genuine submissions — the top three by `g` on the whole board:

| code | g | date | was |
|---|---|---|---|
| `[[37,7,3]]` | ≤0.489 | 2026-08-02 | hidden |
| `[[37,1,7]]` | 0.380 | 2026-07-26 | hidden |
| `[[19,1,5]]` | 0.378 | 2026-07-26 | hidden |

The primary-tracks grid already had this right (`family == "topological" and origin == "baseline"`), and its comment claimed the record chart used "the same convention as the record chart" — it did not. Both now call a single `geo_reference()` helper, so the two cannot drift apart again.

After the fix, `[[37,7,3]]` is the current g record holder in all three weight series:

```
series any w record points:
   2023        80-9-9   g=0.0092
   2023-06-28  24-6-4   g=0.0816
   2025        18-4-4   g=0.0988
   2025        12-4-2   g=0.1088
   2026-07-25  15-7-3   g=0.3429
   2026-07-25  16-6-4   g=0.3594
   2026-07-26  19-1-5   g=0.3779
   2026-07-26  37-1-7   g=0.3803
   2026-08-02  37-7-3   g=0.489
```

@MathysRennela — your guess in the issue was that the missing exact-distance certificate caused it. It did not: nothing in the chart filters on distance tier, and an upper-bound `g` just renders with a `≤` marker (as `≤0.489` does above). The certificate is unrelated.

## #356 — leaderboard toggle for geometric efficiency

The Leaderboard header gets a `kd²/n` ⇄ `g` pill, matching the ones the record chart and tracks grid already use.

- **Both rankings are computed in Python** and carried on each row as `data-erank` / `data-grank`. The client only reorders, so there is no second comparator in JS that can disagree with the first.
- **Eligibility matches the headline card**: a verifier-accepted layout and `d >= GEO_MIN_D`. The 4 of 7 contributors without such a code show `·` and sort last, which is the honest reading — no layout shipped, so no geometric claim. The subtitle says so explicitly in g mode.
- Upper-bound distances propagate: a `g` inherited from `d ≤` shows as `≤0.489`.
- The contributor modal gains a **best g** tile and per-code `g` values, formatted with the same 3-significant-digit convention as the rest of the site.

Ranking in g mode (current data):

```
1  @mathysrennela   best g = ≤0.489      👑
2  @FarLab          best g = 0.38
3  @vprusso         best g = 0.0768
4  @yifanhong       best g = ·
5  @seunomonije     best g = ·
6  @willzeng        best g = ·
7  @shassinger      best g = ·
```

## Verification

- Built the site, then drove the **real generated page** in jsdom (with the real `style.css`): the toggle reorders rows, renumbers ranks, moves the crown, and swaps both the hero card and the subtitle — in both directions, with exactly one metric column, one hero and one subtitle visible per mode.
- The new `.lbbtn` class is deliberately distinct from `.rcbtn`: the chart's handler binds `document.querySelectorAll('.rcbtn')` globally, so reusing that class would have hijacked the chart state. Confirmed no element matches both.
- Confirmed `geo_reference()` agrees with the old inline predicate on all 183 entries, so the tracks grid is byte-identical.
- `run_tests.py --skip-slow`: 6/6 files pass.

`docs/` is intentionally not included — the `site` workflow regenerates and commits it on merge to `main`.

## One thing I did not change

The **headline stat card** still reports `best geometric efficiency 1` achieved by `[[25,1,5]]` — the reference tiling itself, since `geo_pool` in `build()` includes baselines. Reading it as "nobody has beaten the surface code yet" is defensible, but it does present the normalization reference as the record holder, which sits oddly next to a chart and a leaderboard that both exclude it. Left alone as a visible headline change beyond these two issues; happy to follow up if you want it to show the best contender with the ceiling as context.

---
Authors: @FarLab
Provenance: Claude Fable 5 (Claude Code)

### DIFF
--- a/site/build.py
+++ b/site/build.py
@@ -391,6 +391,7 @@ color:var(--ink)}}
 .lb{{margin:18px 0 8px;border:1px solid var(--ln);border-radius:14px;
 background:#fff;overflow:hidden}}
 .lbhead{{display:flex;justify-content:space-between;align-items:center;gap:16px;
+flex-wrap:wrap;
 padding:16px 20px;background:var(--soft);border-bottom:1px solid var(--ln)}}
 .lbh{{font-size:16px;margin:0;font-family:'Space Mono',ui-monospace,monospace;
 text-transform:uppercase;letter-spacing:.03em}}
@@ -443,6 +444,15 @@ a.lbm:hover b{{text-decoration:underline;color:var(--ac)}}
 .lbm b{{font-size:17px;font-variant-numeric:tabular-nums}}
 .lbml{{font-size:11px;color:var(--mut);margin-top:1px;white-space:nowrap}}
 .lbrow{{cursor:pointer}}
+/* Leaderboard metric toggle (issue #356): each row and the hero carry both
+   scores; the mode on the section picks which one is on show, and the crown
+   follows whoever is rank 1 in the active ranking. */
+.lb[data-mode=eff] .lbmgeo,.lb[data-mode=eff] .lbhgeo,
+.lb[data-mode=eff] .lbsgeo,.lb[data-mode=geo] .lbmeff,
+.lb[data-mode=geo] .lbheff,.lb[data-mode=geo] .lbseff{{display:none}}
+.lbmnone b{{color:var(--mut);font-weight:400}}
+.lbcrown{{display:none}}
+.lbrow.lbtop .lbcrown{{display:inline}}
 .qchip{{margin-left:12px;font:inherit;font-size:12.5px;border:1px solid #f59e0b;
 background:#fef3c7;color:#92400e;border-radius:999px;padding:3px 11px;
 cursor:pointer;vertical-align:3px}}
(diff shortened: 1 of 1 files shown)


vprusso (approved):
Verified both fixes against a local build of the branch.

#376: the root cause diagnosis is right. The record chart filtered `!r.topo` (family alone) while the tracks grid used family + baseline origin, so the two had silently diverged despite a comment claiming otherwise. Routing both through a single `geo_reference()` helper (family == "topological" and origin == "baseline") is the correct fix, and I confirmed the chart data now emits `"ref": geo_reference(e)` and every consumer (chart filter, tracks grid, gitem) reads the same helper, so they can't drift again. The three submitted topological codes ([[37,7,3]], [[37,1,7]], [[19,1,5]]) now race, which is the intended behavior.

#356: the toggle is sound. Both rankings are computed in Python and carried as data-erank / data-grank; the client only reorders and renumbers, so there's no second comparator to disagree. Confirmed `.lbbtn` is a separate class from `.rcbtn` (the chart binds `.rcbtn` globally, so sharing it would have hijacked chart state). The dot-and-sort-last treatment for contributors with no eligible layout is the honest reading. Built the page and confirmed the toggle, dual ranks, crown move, and hero/subtitle swap all render.

On the headline stat card you flagged: agreed it reads oddly to show `best geometric efficiency 1` by [[25,1,5]] (the normalization reference itself) right next to a chart and leaderboard that both exclude it. Worth a follow-up to show the best contender with the ceiling as context, but it's pre-existing and out of scope here, not blocking.

Issue #377, opened 2026-08-03
Add results from arXiv:2607.28795 to the challenge leaderboard


vprusso:
https://arxiv.org/abs/2607.28795

vprusso:
Started on this:
- Reference added: #378 (arXiv:2607.28795).
- Cited in the reference-bars section of TRACKS.md: #379, listing [[300,60,14]] (kd^2/n 39.2), [[540,108,18]] (kd^2/n 64.8), and [[975,195,<=24]].

On seeding the actual codes (not just citing them): the construction is fully specified in the paper (Definition 4: mitten code = LP(A,B) with A=[g1+g2+g3, e+g4+g5], B=[h1+h2+h3, e+h4+h5] over F_2[G]; the 5-block H_X/H_Z are given explicitly). Table I gives the group per code, e.g. [[300,60,14]] over C_10 x S_3. The only missing piece is the explicit generators g_i, h_i, which are in the appendix (Table XIII) and the open-source toolkit. So a faithful reconstruction with a checkable distance witness is achievable, unlike the Kasai / tile-code bars.

Proposed next step: reconstruct [[300,60,14]] and [[540,108,18]] from the published generators, verify locally, and seed them as literature baselines attributed to the authors. Holding on that until the generator extraction is confirmed against the paper, to avoid seeding a mis-reconstruction under the authors names.

Separately, re: building circuit-level (not just code-level) verification: their Figure 3 pipeline is the model. It runs Level I (code-level: parameters, distance via their GPU estimator sQetch) then Level II (circuit-level: compile to a noisy Stim circuit, decode, measure logical error rate per round). Our harness currently stops at Level I. Adding a Level-II circuit-level-LER check is a substantial but worthwhile direction, and it lines up with the same message from OmniQEC (arXiv:2607.25865) and Hong (arXiv:2607.27644) that code-level kd^2/n does not fully predict operational performance.

FarLab:
Progress on seeding the mitten codes — the generator-extraction question is now settled, with one twist.

[...]

PR #385, opened 2026-08-04
Seed literature baseline: [[150,30,10]] mitten code (arXiv:2607.28795)

FarLab:
Part of #377. One of six mitten codes (Bhardwaj, Ma, Meister, King, Bluvstein, Preskill, Cain, Xu, Huang — arXiv:2607.28795, Table I) that fit the n ≤ 700 cap; [[780,156,≤22]] and [[975,195,≤24]] stay as TRACKS.md reference bars only.

- `origin: baseline`, attributed to the paper's authors (no @handle, per the CONTRIBUTING baseline exemption); the note documents the reproduction.
- Check matrices from the authors' published toolkit (github.com/a7b/yarn @ 82fb695, `processor_codes/mitten/`). Five of the paper's eight codes reproduce bit-for-bit from Table XIII generators + GAP `Elements(SmallGroup(·))`; the three direct-product-group codes use an undocumented element ordering, so the published artifact is used verbatim (details in the note and in #377).
- Witnesses found by `research/kit/submit.make_submission` at exactly the paper's claimed distance; `verify/validate_candidate.py` passed (schema, CSS, witnesses, refutation, dedup).

Attribution: Claude Fable 5 (Claude Code)


FarLab:
For the record: withdrawn because the board already carries this code. `codes/150-30-10.json` ([[150,30,10]] ZSZ-LP, @yifanhong, PR #359, from arXiv:2607.27644) is WL-equivalent to the mitten [[150,30,10]] — same parameters, same check weight 9, and the underlying groups are isomorphic (Z15 ⋊_11 Z2 ≅ C5 × S3). Two independent papers, one code; the earlier submission stands. This PR would have overwritten that entry — my collision check had run against a stale checkout; dedup for the remaining five mitten seeds was re-verified against current main and is clean.

Attribution: Claude Fable 5 (Claude Code)

PR #413, opened 2026-08-05
Add [[220,44,10]] non-abelian lifted-product (mitten) code

yhu1996:
## Code submission

- Parameters: [[n, k, d]] = [[220,44,10]]
- Tracks: weight-9plus × unrestricted (computed by the verifier: max check weight 9, no layout → unrestricted)

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`  (`codes/220-44-10.json`)
- [x] Distance witness(es) included for each reported side  (X and Z witnesses, weight 10)
- [x] `python verify/qldpc_verify.py codes/220-44-10.json` passes locally  (ok=True; not refuted; weight-9plus / unrestricted)
- [x] Construction and references filled in under `provenance`  (mitten LP(A,B), arXiv:2607.28795)
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`  (not an exact/WL duplicate of any board entry per validate_candidate; literature-novelty flagged unverified)

It extends the frontier at block length n=220: the current board has no code with n≤220, k≥44, and d≥10 simultaneously. The nearest higher-rate board neighbors are [[117,44,3]], [[148,44,4]], [[170,52,3]], [[201,78,3]] (all d≤4), so this code strictly improves distance at this (n, k) region — it beats those entries on the d axis while matching or exceeding k. Verified board_advancing=True, dominated_by=[] against the current board (commit 5ad05ad).

Distance is a witness-backed upper bound (d≤10); not exact-certified. 

### Provenance / trust 

This is a non-abelian lifted-product ("mitten") code, LP(A,B) per Bhardwaj et al.
(arXiv:2607.28795, Def. 4/8), over C11⋊C4 (metacyclic, order 44); n=5|G|, k=|G| (20% rate), uniform check weight 9.

Construction family (mitten / non-abelian lifted product) is published; this specific instance/order is not in the paper's tables or the authors' repo (github.com/a7b/yarn) — a new instance at an un-tabulated group order.

LLM generated: this code search was done by Jukebox


codes/220-44-10.json
```diff
@@ -0,0 +1,1996 @@
...
+    104,
+    108,
+    123,
+    126
+   ]
+  }
+ },
+ "provenance": {
+  "authors": [
+   "jukebox-autoresearch"
```
FarLab: If the code is new (so not explicitly in a paper), you should put your own github handle as the author and the model that you used in a "model" field. By model I mean the explicit LLM + version, e.g. GPT 5.6 sol, claude fable 5 etc...

PR #414, opened 2026-08-05
Add [[230,46,8]] non-abelian lifted-product (mitten) code

yhu1996:
## Code submission

- Parameters: [[n, k, d]] = [[230,46,8]]
- Tracks: weight-9plus × unrestricted (computed by the verifier: max check weight 9, no layout → unrestricted)

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`  (`codes/230-46-8.json`)
- [x] Distance witness(es) included for each reported side  (X and Z witnesses, weight 8)
- [x] `python verify/qldpc_verify.py codes/230-46-8.json` passes locally  (ok=True; not refuted; weight-9plus / unrestricted)
- [x] Construction and references filled in under `provenance`  (mitten LP(A,B), arXiv:2607.28795)
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`  (not an exact/WL duplicate of any board entry per validate_candidate; literature-novelty flagged unverified)

It extends the frontier at block length n=230: the current board has no code with n≤230, k≥46, and d≥8 simultaneously. The nearest higher-rate board neighbors are [[170,52,3]], [[201,78,3]] (both d=3), so this code strictly improves distance at this (n, k) region — it beats those entries on the d axis while matching or exceeding k. Verified board_advancing=True,
dominated_by=[] against the current board (commit 5ad05ad).

Distance is a witness-backed upper bound (d≤8); not exact-certified. 

### Provenance / trust

This is a non-abelian lifted-product ("mitten") code, LP(A,B) per Bhardwaj et al.
(arXiv:2607.28795, Def. 4/8), over D23 (dihedral, order 46); n=5|G|, k=|G| (20% rate), uniform check weight 9.

Construction family (mitten / non-abelian lifted product) is published; this specific instance/order is not in
the paper's tables or the authors' repo (github.com/a7b/yarn) — a new instance at an un-tabulated group order.

LLM generated: this code search was done by Jukebox


codes/230-46-8.json
```diff
@@ -0,0 +1,2082 @@
...
+    91,
+    141,
+    142,
+    182
+   ]
+  }
+ },
+ "provenance": {
+  "authors": [
+   "jukebox-autoresearch"
```
FarLab: If the code is new (so not explicitly in a paper), you should put your own github handle as the author and the model that you used in a "model" field. By model I mean the explicit LLM + version, e.g. GPT 5.6 sol, claude fable 5 etc...

FarLab:
Hi @yhu1996, thanks for the code submissions! I left a comment in your code which should be an easy fix (the same comment holds for your other code submission).

codes/230-46-8.json
```diff
@@ -2069,14 +2069,15 @@
...
+   "@yhu1996"
   ],
   "construction": "Mitten lifted product LP(A,B) (Bhardwaj et al arXiv:2607.28795, Def.4/8) on non-abelian group D23 (order 46): a0=[6, 12, 43], a1=[0, 28, 45], b0=[2, 5, 14], b1=[0, 10, 41] over F2[G]. n=5|G|, k=|G|, weight-9.",
   "references": [
    "arXiv:2607.28795"
   ],
   "date": "2026-08-05",
-  "notes": ""
+  "notes": "",
+  "model": "Jukebox (Claude Opus 4.8)"
```
FarLab: Thanks for the update! This should be just "Claude Opus 4.8" :)

yhu1996: Since I really want to give the credit to my agent Jukebox, can I add a "agent" field and put "Jukebox"?

FarLab: Good point, sure you can add Jukebox in a new "agent" field.

yhu1996: Thank you so much! "agent" field is not in the schema so I put it in the "note".

FarLab: As you might have realized already, at the moment we do not accept a new field. If you believe the agent is very important and deserves credit, you can open an issue so we can add it to the schema.

yhu1996: Yes, I noticed. For now I will just leave it in the note and the same changes made in another submission!

PR #416, opened 2026-08-06
site: compact model ensembles so they stop overrunning the date column (#407)

### LINKED ISSUE #407: Model field gets crammed when specifying multiple models
When specifying multiple models, the model field gets crammed as seen in the screenshot below (thanks @cbjuan for making us aware of this!):

Image

FarLab:
Fixes #407.

### Why it happened

`table.board` is `table-layout:fixed` with the model column at 14%, and `.board td.model` is `white-space:nowrap`. A model string wider than that column therefore could neither wrap nor widen the column — it simply painted over the date cell, which is what the screenshot in the issue shows. A single model name fits, so this only surfaced once ensembles arrived: `Claude Opus 4.6 + GPT-5.3-Codex + Gemini 3.1 Pro Preview` is 56 characters against a column sized for roughly 20.

### The fix

The authors column already solves exactly this problem — `authors_compact()`, whose docstring states the rule as "so every row is one line": it collapses a long author list to `Surname et al.` and keeps the full list in the cell's hover title and on the detail page. `models_compact()` follows that precedent: a single model renders in full, an ensemble renders as its first name plus a `+N` badge.

Nothing is lost. The full ensemble stays in the cell's `title`, on the code's detail page, and — importantly — in the row's `data-model` attribute, so searching the board for `gemini` still matches a row whose visible cell reads `Claude Opus 4.6 +2`.

Ensembles reach the site two ways: as a JSON list, joined with `, ` by `_model_str()` since #409, and as a single free-text string with ` + ` between names, which is what the seven codes currently on the board use. `model_parts()` splits on both.

I also added `overflow:hidden;text-overflow:ellipsis` to the cell. Compaction already handles the ensemble case; the clip is the structural guarantee, bounding a single long name at any font size so nothing can paint over a neighbouring column again.

### Verification

Built the site locally and checked the rendered output:

| visible in cell | hover title |
|---|---|
| `Claude Opus 4.6 +2` | `Claude Opus 4.6 + GPT-5.2 + Gemini 3 Pro Preview` |
| `Claude Opus 4.6 +2` | `Claude Opus 4.6 + GPT-5.3-Codex + Gemini 3.1 Pro Preview` |

Every single-model row is unchanged (`Claude Fable 5`, `Tencent Hy3`, `Xiaomi MiMo-V2.5`, …), the detail page for `288-32-8` still shows the full ensemble, and `data-model` still carries the full lowercased string for search. Helper checked against empty input, a plain name, a JSON list, both separators, stray whitespace, and an HTML-injection attempt (escaping holds). `run_tests.py`: 6/6 files pass.

No `docs/` changes are included — the `site` workflow rebuilds and commits the generated pages on push to main.

### DIFF
--- a/site/build.py
+++ b/site/build.py
@@ -681,11 +681,18 @@ color:var(--ink);min-width:2.4em;text-align:right}}
 .star{{color:var(--ac);width:18px}}
 .auth{{color:var(--mut);font-size:13px}}
 .board td.date{{color:var(--mut);font-size:13px;white-space:nowrap}}
-.board td.model{{color:var(--mut);font-size:13px;white-space:nowrap}}
+/* nowrap keeps the row one line; the table is table-layout:fixed, so without
+   the clip a name wider than the column paints over the date cell (#407).
+   models_compact() already shortens ensembles -- this bounds the single-name
+   case too, at any font size. */
+.board td.model{{color:var(--mut);font-size:13px;white-space:nowrap;
+overflow:hidden;text-overflow:ellipsis}}
 .nomodel{{color:#94a3b8;display:inline-flex;vertical-align:middle}}
 .modelmark,.modelicon{{display:inline-flex;align-items:center;gap:5px;
 vertical-align:middle}}
-.board td.model .modelname{{font-size:13px;color:var(--ink)}}
+.board td.model .modelname{{font-size:13px;color:var(--ink);
+overflow:hidden;text-overflow:ellipsis;min-width:0}}
+.board td.model .modelmark{{max-width:100%}}
 /* Let the wide board scroll instead of crushing columns, and progressively
    drop the secondary metadata columns (model, then date) on smaller screens.
    Under the breakpoints the table sizes to content so freed space redistributes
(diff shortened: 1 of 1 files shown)


vprusso (approved):
Built the branch and checked the generated output rather than the description.

Verified:

- The 7 ensemble rows compact to `Claude Opus 4.6 +2`, with the full string in the cell's hover title.
- `data-model` (build.py:3181) still carries the full lowercased ensemble, so searching `gemini` matches a row whose visible cell reads `Claude Opus 4.6 +2`. Checked on all 7.
- Detail pages under `docs/codes/` keep the full ensemble.
- The CSS renders well-formed. `.modelname` is a direct flex item of `.modelmark` (`display:inline-flex`), so the `min-width:0` is doing real work rather than being defensive: without it a flex item will not shrink below its content width and the ellipsis never appears.

The longest single model name currently on the board is 19 characters (`OpenAI GPT-5.6 Luna`), so the clip is a bound against future input rather than something active today, which matches what the comment claims for it.

Two notes, neither blocking.

`model_parts` splits on `,` as well as `+`. Nothing on the board has a comma inside a single model name, but the schema since #409 allows free text there, and `GPT-5, Turbo` would render as `GPT-5 +1`. Splitting the free-text form on `+` only, and letting the JSON list carry genuine ensembles, would remove that case.

In card mode (max-width 820px) `.board td.model` is redefined as `display:inline-flex`, and `text-overflow:ellipsis` has no effect on a flex container; the ellipsis that actually renders there comes from the `.modelname` rule. The comment on the cell rule reads as though the clip is guaranteed at any font size and any width, which holds only in desktop table mode. The behaviour is correct either way, it is the comment that claims more than the rule delivers.

No test for `models_compact`, but `authors_compact` has none either and there are no site helper tests in the repo, so that is consistent with existing practice.

PR #423, opened 2026-08-06
certify: 14 codes proven exact by MILP

vprusso:
Adds `certs/<slug>.json` for 14 codes whose distance was previously an upper bound. `site/build.py` reads `certs/` at build time, so the board moves from 52 to 66 certified exact.

| | | | |
|---|---|---|---|
| `37-7-3` | `56-6-7` | `72-8-7` | `72-12-6` |
| `54-6-9` | `78-4-10` | `80-8-10` | `80-9-9` |
| `84-6-10` | `90-8-10` | `108-8-10` | `128-12-8` |
| `192-12-8` | `288-8-12` | | |

Every side is infeasible at weight `d-1`, so `d` is exact on both. Solver is scipy/HiGHS through `verify/certify.py`, one MILP per logical-basis row. Run at `--tlim 300`, then 1800 for the codes that needed it. `288-8-12` took 10186s and `108-8-10` 5951s; the rest finished under an hour.

### Two names the certification disproves

`80-8-10` and `80-9-9` are now proven to have `d = 8`, but their `name` fields read `[[80,8,10]]` and `[[80,9,9]]`, and the filenames say the same. The board already shows the correct distance, because the site builds the parameter string from the `n`/`k`/`d` fields rather than from `name`, so the discrepancy is confined to the hover title and the filename.

I left it alone rather than mix a rename into a certification batch. Happy to follow up either way, but it is worth deciding: the certification is what makes those names provably wrong.

### The documented envelope is optimistic

Six targets are still unproven and running: `96-4-12`, `98-6-12`, `120-8-12`, `140-10-10`, `144-12-12`, `240-12-12`.

All six sit inside the envelope `CONTRIBUTING.md` documents as practical (`d <= 13`, `k <= 12`). The cost model there is `2 * k * tlim`, and it is the `k` end that bites: at `k = 12` a single code is 24 MILPs, so `tlim 1800` is a 12-hour worst case. The envelope holds for small `k` and is optimistic at `k = 8-12`. I would rather report that than quietly widen the bound.

### Selection note

Targets were chosen by each code's `distance.*.confidence` field, which turned out to be the wrong index: 18 of the 38 already had cert files, so those runs were redundant. `certs/<slug>.json` is the authoritative record of exactness, not the confidence field on the code. Only the 14 genuinely new ones are in this PR.

No `docs/` changes: the site workflow rebuilds the generated pages on push to main.

### DIFF
--- a/certs/108-8-10.json
+++ b/certs/108-8-10.json
@@ -1,0 +1,18 @@
+{
+ "name": "[[108,8,10]] bivariate bicycle code",
+ "d": 10,
+ "solver": "scipy/HiGHS MILP",
+ "sides": {
+  "X": {
+   "value": 10,
+   "exact": true,
+   "note": "no logical < 10 exists"
+  },
+  "Z": {
+   "value": 10,
+   "exact": true,
+   "note": "no logical < 10 exists"
+  }
+ },
+ "d_exact": true
+}
--- a/certs/128-12-8.json
+++ b/certs/128-12-8.json
@@ -1,0 +1,18 @@
+{
+ "name": "[[128,12,8]]",
+ "d": 8,
+ "solver": "scipy/HiGHS MILP",
+ "sides": {
+  "X": {
+   "value": 8,
+   "exact": true,
+   "note": "no logical < 8 exists"
+  },
+  "Z": {
+   "value": 8,
+   "exact": true,
+   "note": "no logical < 8 exists"
+  }
+ },
+ "d_exact": true
+}
--- a/certs/144-12-12.json
+++ b/certs/144-12-12.json
@@ -1,0 +1,18 @@
+{
+ "name": "[[144,12,12]] bivariate bicycle code",
+ "d": 12,
+ "solver": "CryptoMiniSat 5.14 SAT",
+ "sides": {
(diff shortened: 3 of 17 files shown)


FarLab (approved):

vprusso:
Pushed 3 more: 96-4-12, 98-6-12 and 144-12-12, proven exact by a SAT formulation (CryptoMiniSat, XOR clauses + sequential-counter cardinality) after resisting the MILP at tlim 1800. Batch is now 17 certs, board 52 to 69 exact. The PR body's straggler list shrinks to 120-8-12, 140-10-10 and 240-12-12, which time out under both solvers at 1200s/call. Title kept; the diff is the source of truth on the count.

vprusso:
Final addition: 240-12-12, both sides infeasible at weight 11 (30402s at tlim 3600). The batch is now 20 certifications and the board moves from 52 to 72 exact. Every code in the d <= 13, k <= 12 envelope this PR started from is proven; the split ended up 14 by MILP and 6 by the SAT backend, with the SAT wall times (0.7-8.4h) on codes the MILP could not finish at any tested budget.

PR #433, opened 2026-08-08
[[625,50,3]] holey rotated surface code — geometric efficiency 0.72 (r = √2)

aarontrowbridge:
## Code submission

- Parameters: [[n, k, d]] = [[625, 50, 3]]
- Tracks: weight-4 × local-2d-single (verifier-computed); family tag `topological`
- Distance confidence: upper_bound (witnesses both sides; d = 3 with explicit weight-3 logicals)

### Checklist
- [x] `python verify/qldpc_verify.py codes/625-50-3.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` — n/a (new parameters; see note for the nearest entries)

**weight-4 × local-2d-single**, on the k axis: the cell's entries sat at k ≤ 2; this lands k = 50 at n = 625 with d = 3 and w = 4 (`validate_candidate.py` labels it `board_advancing: true`, not dominated, not a duplicate).

It is also a claim on the board's **geometric-efficiency** metric: with its honest single-layer layout at the square-check packing floor (r = √2 exactly, unit min spacing), g = 4kd²/(nρ²r⁴) = **0.720** — the previous non-surface-code best was 0.489 ([[37,7,3]]). The research note documents the search (punctured 6.6.6 color codes vs holey rotated surface codes), the d = 3 → hole-spacing → k/n < 1/9 chain that keeps this family below 1.0, and the dead ends.

Construction: rotated surface code on a 25×25 vertex grid (the board's [[25,1,5]] pattern generalized), 49 plaquette holes punched on the parity-matched period-3 cell sublattice with margin 3 from the boundary; k = 1 + holes; d = 3 from the shortest hole-to-boundary string. Full details + reproduction in `notes/625-50-3.md`.


vprusso (approved):
Good to see you here, Aaron :)

Re-verified everything independently rather than from the PR text, and it all holds:

- Verifier: all 15 checks pass; computed class weight-4 x local-2d-single, radius 1.4142, single layer.
- Layout honesty, recomputed from the raw JSON: 625 distinct sites, minimum spacing exactly 1.0, and the max check diameter is sqrt(2) on both sides, so the r in the g formula is the genuine square-check packing floor, not a claimed value.
- k = 50 recomputed from the checks; both witnesses are weight 3, commute with the opposite side, and are nontrivial (rank test against the stabilizer).
- `validate_candidate.py` against the current board: passes, not refuted, no exact or WL duplicate, advances weight-4 x local-2d-single. (Run it from a checkout where the file is not already in codes/; from the PR branch it flags itself as its own duplicate.)
- The two frontier claims check out: the cell's prior entries top out at k = 2 ([[16,2,4]], [[36,2,6]], [[64,2,8]]), and the best non-surface g on the board is 0.489 for [[37,7,3]] (which is w = 6, so this is also the first w = 4 code anywhere near that number).

One upgrade available immediately: at d = 3 exactness is cheap, and I ran it. SAT certification (the CryptoMiniSat backend from #423) proves no logical of weight <= 2 exists, both sides, in under a second. So this can enter the board as d exact rather than an upper bound; I will push the cert file after merge, or you can note it here and a maintainer run will pick it up.

The note is a genuinely useful artifact beyond the code itself: the 6.6.6 puncture ceiling argument (family capped near 0.58 by the hexagon packing floor) is the kind of negative result that saves the next person a week, and the d = 3 -> hole-spacing -> k/n < 1/9 chain documents exactly why this family sits below 1.0 rather than leaving it as an empirical observation.

Approving.

aarontrowbridge:
thanks @vprusso! was a fun experiment!

PR #447, opened 2026-08-09
Add honest bilayer layout to [[56,6,7]] → local-2d-bilayer, co-authored @mathysrennela

MathysRennela:
## What this does

Adds an honest, verifier-accepted bilayer layout to the existing `[[56,6,7]]` submission (2BGA on dicyclic Dic_7). The code previously shipped **no** `locality` block, so it competed only as `unrestricted`.

**Layout**: qubits `i` and `i+28` share a site on a stacked 4×7 unit grid, `layers = 2`. The trusted verifier (`verify/qldpc_verify.py`) measures:

- interaction radius `r = 6.7082` (≤ 7.0 bilayer cap)
- max qubits per site `2` (≤ layers)
- min site spacing `1.0`
- computed `locality_class = local-2d-bilayer`

`checks`, `distance`, and `witnesses` are byte-identical to the original entry (`d = 7`, upper bound, both sides). A layout-certification section is appended to `notes/56-6-7.md` (the original note is preserved).

## What frontier does this advance?

**Track cell: local-2d-bilayer × weight-6.** Previously `unrestricted` (no cell at all); with this layout `[[56,6,7]]` joins the 2D-local cell as a Pareto contending member on `(n, k, d, w)`, and its **geometric efficiency `g = 4kd²/(nρ²r⁴)` becomes defined**: `g ≈ 0.0026` (upper-bound distance tier inherited), `kd²/n = 5.25`.

This is a correction to an existing entry (not a new code), so it is not subject to the one-new-code-per-PR cap.

## Authorship

This is a **submission**, so per the board's co-authorship rule (`check_authorship.py`: a contributor adding material value is listed as a co-author; the PR author must be a `@handle` author), **`@mathysrennela` is added as a co-author** for the layout contribution. Original code credit remains with `@vprusso`:

- `authors`: `["@vprusso", "@mathysrennela"]`
- Both authorship `--author` checks pass locally for either PR author.
- `provenance.notes` and `notes/56-6-7.md` document the layout, its certification, and reproduction.

## Scope

No changes to `verify/`, `schema/`, workflows, or the site builder. This touches only `codes/56-6-7.json` (adds `locality`, co-author, note) and `notes/56-6-7.md`.


MathysRennela:
@vprusso the leaderboard doesn't have a hard rule for it yet but we have so far considered that if somebody completes a submittion, they get co-authorship (like in the code that Sean corrected). LMK if you want me to handle it differently (like for literature baselines as in #446).

vprusso (approved):
Re-derived every geometric claim from the raw coordinates rather than reading them off the report, since a layout PR is entirely a claim about geometry.

Verified:

- `checks` and `distance` are semantically identical to main. Worth stating explicitly because the diff is 727 insertions against 496 deletions, so the file is fully rewritten by reformatting and "byte-identical" is not visible from the diff; comparing the parsed JSON is what settles it.
- 28 distinct sites, exactly 2 qubits per site against `layers = 2`, minimum site spacing exactly 1.0. So the radius is not bought by collapsing qubits onto a site.
- Measured interaction radius 6.7082 against the 7.0 bilayer cap, recomputed as the max Euclidean check diameter over both X and Z.
- Verifier: 15/15 checks, `locality_class = local-2d-bilayer`, `weight_class = weight-6`.
- Independent `g` computation reproduces 0.00259, matching the 0.0026 in the description.

The scope point holds: `check_submission_scope.py` counts ADDED code files, and this modifies an existing entry, so the one-new-code-per-PR cap does not apply.

One question, which is really the subject of #451 rather than this PR. This adds `@mathysrennela` as co-author on a submission, while #448 leaves the seven literature baselines with their original attribution untouched. That distinction looks right to me: a baseline's `authors` field records who published the code, and appending a layout contributor there would misattribute the science. But the two PRs landing with different conventions is exactly the ambiguity #451 is about, so it is worth having the rule written down before this sets precedent by merge order.

Approving on the technical content.

PR #448, opened 2026-08-09
Certify bilayer layouts for 7 literature baselines → local-2d-bilayer, g computable

MathysRennela:
## What this does

Certifies **honest bilayer layouts** for 7 literature baselines that shipped no `locality` block and therefore competed only as `unrestricted`. All use the stacked-grid bilayer pattern (qubits `i` and `i+n/2` share a site on an `l_x × l_y` grid, `layers = 2`), judged by the trusted verifier.

| code | grid | r (≤ 7.0) | per-site | g |
|---|---|---|---|---|
| `[[30,4,6]]` | 3×5 | 4.472 | 2 | 0.0120 |
| `[[32,8,4]]` | 4×4 | 4.243 | 2 | 0.0123 |
| `[[40,6,5]]` | 4×5 | 5.000 | 2 | 0.0060 |
| `[[42,8,3]]` | 3×7 | 6.325 | 2 | 0.0011 |
| `[[56,2,8]]` | 4×7 | 6.708 | 2 | 0.0011 |
| `[[60,8,6]]` | 5×6 | 5.831 | 2 | 0.0042 |
| `[[70,6,8]]` | 5×7 | 6.708 | 2 | 0.0027 |

Each now computes `locality_class = local-2d-bilayer`, competes in the 2D-local cell, and has a defined geometric efficiency `g = 4kd²/(nρ²r⁴)`.

## What frontier does this advance?

Each code moves from `unrestricted` (no 2D-local cell) into the `local-2d-bilayer` cell, where it now participates in the `(n, k, d, w)` Pareto comparison. `g` becomes defined for each (upper-bound distance tier inherited). These are corrections to existing entries, not new codes — so the one-new-code-per-PR cap does not apply.

## Authorship / scope

- **Authorship untouched**: all 7 remain literature baselines with their published authors (Liang–Liu–Song–Chen, Lin–Pryadko, Wang–Pryadko). No `@handle` authors were added, preserving the baseline convention.
- Layout credit and reproduction are recorded in each `provenance.notes` (found by @mathysrennela, Aug 2026).
- `checks`, `distance`, and `witnesses` unchanged for every file; only `locality` and the notes field were added.
- No changes to `verify/`, `schema/`, workflows, or the site builder.

Reproduction, e.g. for `[[30,4,6]]`: `locality = {"coordinates": [[float((i % 15) % 3), float((i % 15) // 3)] for i in range(30)], "layers": 2}`, then `uv run python verify/qldpc_verify.py codes/30-4-6.json` → `locality_class = local-2d-bilayer`, `interaction_radius = 4.472`.


vprusso (approved):
Recomputed all seven layouts from the raw coordinates rather than trusting the reported radii.

| code | sites | max/site | min spacing | measured r | <= 7.0 |
|---|---|---|---|---|---|
| `30-4-6` | 15 | 2 | 1.0 | 4.4721 | yes |
| `32-8-4` | 16 | 2 | 1.0 | 4.2426 | yes |
| `40-6-5` | 20 | 2 | 1.0 | 5.0 | yes |
| `42-8-3` | 21 | 2 | 1.0 | 6.3246 | yes |
| `56-2-8` | 28 | 2 | 1.0 | 6.7082 | yes |
| `60-8-6` | 30 | 2 | 1.0 | 5.831 | yes |
| `70-6-8` | 35 | 2 | 1.0 | 6.7082 | yes |

Every one honest on all three counts the class depends on: occupancy never exceeds `layers`, minimum site spacing is exactly 1.0 so no radius is bought by collapsing qubits, and the measured max check diameter is inside the bilayer cap. `checks` and `distance` are unchanged on all seven, and the verifier passes each with `locality_class = local-2d-bilayer`.

On scope: `check_submission_scope.py` gates on ADDED code files, and all seven of these already exist on the board, so seven files in one PR is within the rule as written. (When I ran the scope check inside a worktree it reported two unrelated files, which turned out to be my own untracked submissions leaking in from the parent checkout, not anything from this branch.)

Attribution: this leaves the literature baselines with their original author lists (Liang/Liu/Song/Chen, Wang/Pryadko) rather than appending a layout contributor, which I think is the right call for the reason you give in #451 -- a baseline's `authors` records who published the code, so adding a layout contributor there would misattribute the science. Noting only that #447 takes the other branch of that rule for a submission, so the two together are a good concrete test case for whatever #451 settles on.

Approving.

PR #449, opened 2026-08-09
[[700,57,3]] rectangular holey rotated surface code — g=0.73 (r=√2)

aarontrowbridge:
Rectangular 28×25 extension of #433 (625,50,3). Builds on 28×25 vertex grid (n=700), period-3 hole lattice margin 3 → 28 X-holes + 28 Z-holes + parent k=1 = 57. Even-Lx CSS fix: right Z edge parity flipped (1,0,0,0) to preserve H_X H_Z^T=0. Validated: `validate_candidate` passed true (pre-board), `qldpc_verify` ok true, w=4, r=√2 (1.4142), layers 1, local-2d-single, d=3 witnesses weight 3, g=9k/n=0.7329 >0.72. See research/candidates/build_rect_holey.py and notes/700-57-3.md.

Validator verdict (pre-board):
passed true, board_advancing true, refuted false (8000 RIS trials)
Verifier: r=1.4142 layers 1 class local-2d-single weight-4


codes/700-57-3.json
```diff
@@ -0,0 +1,6606 @@
...
+        494,
+        520,
+        546
+      ]
+    }
+  },
+  "provenance": {
+    "authors": [
+      "aarontrowbridge"
+    ],
```
FarLab, line 3793: author should be the github handle (with the @ sign) and the LLM model seems to be missing here. only mention the model used, not the agent/harness/IDE, eg GPT 5.6 sol, Claude Opus 5 etc...

aarontrowbridge:
Fixed: `authors` → `["@aarontrowbridge"]` and added `model: "Muse Spark 1.2"` (pushed 1e94330). Thanks for the catch — also pre-fixed the same on #455 and #456.

PR #464, opened 2026-08-11
Add [[700,85,3]] holey rotated surface code, g=1.093 (exact d=3)

MathysRennela:
## Code submission

- **Parameters:** [[700, 85, 3]]
- **Tracks:** weight-4 × local-2d-single (r = √2, single layer)
- **Distance confidence:** exact (certificate included below and in `certs/700-85-3.json`)

This advances the **weight-4 × local-2d-single** cell at the r=√2 packing floor. Geometric efficiency `g = 9k/n = 9·85/700 = 1.0929`, which beats:

- the previous best non-surface board entry `[[700,75,3]]` (g ≈ 0.964), and
- the surface-code normalization itself (g = 1.0).

The gain is **exact maximum-hole packing**. The family's earlier "g < 1" ceiling was derived for *uniform period-3 sublattices* (density 1/9), but the d=3 constraint is local (same-parity Chebyshev ≥ 3, any-parity Manhattan ≥ 3), so an **irregular maximum independent set** is not capped at 1/9. An exact MILP max-independent-set search over grid shape and hole placement finds **84 holes on a 25×28 grid** (interior density 0.174 vs 1/9 ≈ 0.111), giving k = 1 + 84 = 85.

### Verification

- `verify/validate_candidate.py`: **passed: true**, `board_advancing: true`, no exact/WL duplicate, refute gate finds no lighter logical in 8000 RIS trials.
- `verify/qldpc_verify.py`: ok, weight-4, local-2d-single, interaction radius √2, k = 85 recomputed, CSS commutes.
- **Exact distance certificate** (d = 3 on both sides), also committed as `certs/700-85-3.json`:

```json
{
  "name": "[[700,85,3]] holey rotated surface code (exact max-hole packing)",
  "d": 3,
  "solver": "scipy/HiGHS cutoff IP",
  "sides": {
    "X": { "value": 3, "exact": true, "note": "no logical < 3 exists" },
    "Z": { "value": 3, "exact": true, "note": "no logical < 3 exists" }
  },
  "d_exact": true,
  "tlim_per_solve": 120
}
```

This is corroborated by an independent exhaustive enumeration: there is provably **no weight-1 or weight-2 logical operator** on either side (a weight-2 X-logical {a,b} exists iff H_Z columns a,b are identical and K_X columns a,b differ; checked completely). The same exhaustive check passes on the merged board reference code as a control.

### Checklist
- [x] Distance witness(es) included for each reported side (weight-3, both sides)
- [x] `python verify/qldpc_verify.py codes/700-85-3.json` passes locally
- [x] Research note included: `notes/700-85-3.md`
- [x] Exact distance certificate included: `certs/700-85-3.json`


FarLab:
nice, this seems to beat the surface code!

PR #472, opened 2026-08-12
Add [[578,18,20]] Open-boundary weight-8 Tile code

npdeep:
## Code submission

- Parameters: [[n, k, d]] = [[578,18,20]]
- Tracks: local-2d-bilayer / weight-8 (computed by the verifier from H and the layout)

Family tag: tile (a self-declared filter, never used for ranking).

### Checklist
- [x] `python verify/qldpc_verify.py codes/578-18-20.json` passes locally
- [x] No known equivalence. If this may be equivalent to an existing entry, noted in `provenance.notes`

Extends the weight-8 / local-2d-bilayer Pareto frontier alongside [[384,12,17]]: it increases k from 12 to 18 and d from 17 to 20 at the same max check weight 8, trading 194 additional physical qubits.
Score kd^2/n = 12.457, max check weight 8, locality class local-2d-bilayer.

Construction: Open-boundary weight-8 Tile code (B=4, 14 x 14 bulk), with one support swap from the published B=4 tile (arXiv:2504.09171).

Research note: `notes/578-18-20.md`

---
Drafted by `qldpc submit`; edit before requesting review.


MathysRennela:
@npdeep the code checks out but there are some minor issues with the json you provided:
author field should name the author of the code (in your case, @npdeep); the reference on which the code was build is mentioned but not in the field reference; and no model was provided (if any was used).
This comment also holds for your other PRs (#469-#473). Take PR #464 as an example of how such fields are filled.

npdeep:
> @npdeep the code checks out but there are some minor issues with the json you provided: author field should name the author of the code (in your case, @npdeep); the reference on which the code was build is mentioned but not in the field reference; and no model was provided (if any was used). This comment also holds for your other PRs (#469-#473). Take PR #464 as an example of how such fields are filled.

Hi @MathysRennela . I have a custom setup I've built which doesn't neatly classify into one of the publicly known models. It was developed as part of my work at [Magic State Labs](https://www.magicstatelabs.com/). Would it be okay if I referred to the model as "Magic State Labs (Custom)"?

PR #503, opened 2026-08-14
CI: reject prose that cites files a reviewer cannot open

FarLab:
Adds `verify/check_prose.py`, run on every PR against the changed `notes/` and `fieldnotes/` files and the PR body.

## Why

An evidence trail that cites a file nobody else has is not one. Auditing the 276 code-submission PRs merged so far, and the notes currently on the board:

| | unresolvable reference |
|---|---|
| notes | 30 / 96 |
| PR bodies | 10 / 276 |
| fieldnotes | 0 / 11 |

Four habits produce it: local tooling named as if it were committed here (`research/kit/cornucopia.py`, `research/geometry_audit.py`, `research/kit/surrogate_fast.py`); literal scratch paths (`/tmp/dense_7.npz` in four of the dense-code notes) and private checkouts (`~/qldpc-challenge`, machine names); a fieldnote that was never committed, cited from both a note and a PR; and `research/candidates/` — which `AGENTS.md` defines as gitignored working output — offered as the audit trail in nine notes.

Notes are the worse case, and that is a side effect of something that went right. Note coverage went 0% → 85% and the ladder detail correctly moved out of PR bodies into notes; the unverifiable references moved with it, into the artifact a later searcher actually relies on. `notes/700-75-3.md` is already cited by a later submission (#462) for its boundary-parity failure mode, so a dangling pointer there costs someone a re-derivation.

## The rule, and its escape hatch

Every path named must resolve in the PR's own tree — **unless** the file names and pins an external source. That is the convention the good notes already use:

> taken verbatim from the authors' own published artifact: github.com/a7b/yarn @ 82fb695, `processor_codes/mitten/[[300,60,14]]/Hx.npy`
> — `notes/300-60-14.md`

A file that names its source is trusted to be citing that source rather than this repo, so the mitten notes and the qcode-discovery MILP notes pass unchanged. `research/candidates/` has no escape hatch: it is deliberately never committed, so the citation cannot be resolved by anyone.

Also refused: leftover drafting scaffolding (the `qldpc submit` footer, HTML comments, unticked checklist boxes — 37 PRs merged carrying one), session URLs, and a note whose `[[n,k,d]]` is missing or disagrees with its filename (`notes/270-54-10.md` leads with `[[270,54,12]]`; `notes/482-146-42.md` and `notes/578-18-20.md` state none).

**Only changed files are checked.** The 30 notes currently on the board are left as they are — this guards what arrives from here on.

## Calibration

Run over all 96 notes and 11 fieldnotes: 30 flagged, 77 clean, no false positives after two fixes — `verify/gf2_fast` resolves through its `.cpp`/`.so`, and `module.function` / `file.py::symbol` forms resolve through the module file. Each distinct problem is reported once. `notes/TEMPLATE.md` is exempt: a template is scaffolding by definition.

## Supporting changes

- **`AGENTS.md` gains a "Writing the submission" section.** This repo's record is that prose guidance does not hold and CI does: one-code-per-PR and the n≤700 cap have had zero violations since they became checks, while provenance fix-up commits went 3 → 1 → 23 across the same window despite five `CONTRIBUTING.md` edits. The section is written as the checker's rationale rather than as a second place to state the rule. `notes/TEMPLATE.md` gets the same rule where it will be copied.
- **`verify.yml` runs the authorship check before the refutation gate.** The cheap metadata checks ran last, which is how #278 spent a full ~15-minute gate on 18 valid layouts and then failed on a single authorship mismatch — those layouts were never resubmitted, and `codes/263-16-12.json` still sits at radius 5.0 today. The receipts step added since is untouched and still runs after the gate.

## Notes for review

- The prose job runs from the base branch, like every other check here, so a PR cannot weaken the check that judges it.
- The PR body reaches the checker as a **file via the environment**, never as a shell argument — it is untrusted input.
- The job runs on all PRs including docs-only ones, which `verify.yml` deliberately skips; it is stdlib-only and takes seconds.
- `check_prose.py` is **not** part of the hash-pinned trusted stack — it does not decide whether a code is valid, so no `check_validator_integrity.py --update` is needed.
- The step no-ops until this merges, since the base branch has no `check_prose.py` yet.

Full suite green locally: 8/8 test files, including the new `verify/test_check_prose.py`.

### DIFF
--- a/.github/workflows/prose.yml
+++ b/.github/workflows/prose.yml
@@ -1,0 +1,45 @@
+name: prose
+
+# Runs on every PR, including docs-only ones (verify.yml deliberately skips those,
+# but a note is exactly what needs checking here). Stdlib-only and seconds long.
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+
+jobs:
+  prose:
+    runs-on: ubuntu-latest
+    steps:
+      - name: check out submitted tree
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          path: submitted
+      # The checker runs from the base branch, like every other check here: a PR
+      # must not be able to weaken the check that judges it.
+      - name: check out trusted tree (base branch)
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          fetch-depth: 0
+          path: trusted
+      # The PR body is untrusted input: it reaches the checker as a FILE via the
+      # environment, never interpolated into a shell command.
+      - name: capture PR body
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: printf '%s' "$PR_BODY" > "$RUNNER_TEMP/pr_body.md"
+      - name: prose check (cited paths must resolve; no scaffolding left in)
+        working-directory: trusted
+        run: |
+          if [ -f verify/check_prose.py ]; then
+            python3 verify/check_prose.py \
+              --root ../submitted \
+              --base origin/${{ github.base_ref }} \
+              --body-file "$RUNNER_TEMP/pr_body.md"
+          else
+            echo "base branch has no verify/check_prose.py yet; skipping"
+          fi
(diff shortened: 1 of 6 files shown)


MathysRennela:
This is a really good feature. Models have a tendency to reference whatever is in the repo, regardless of what is publicly available.

MathysRennela (approved):

vprusso (approved):

PR #506, opened 2026-08-14
Add [[472,122,16]] pair-partition CPM baseline (Okada–Kasai girth-8 seed)

willzeng:
## Code submission

- Parameters: [[n, k, d]] = [[472, 122, 16]]
- Tracks: unrestricted / weight-8
- Distance confidence: upper_bound

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/472-122-16.json` passes locally
- [x] Construction and references filled in under `provenance`
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`

### What frontier does this advance?

Weight-8, unrestricted locality. This literature baseline strictly Pareto-dominates the existing `codes/472-122-14.json` baseline: same n = 472 and k = 122 at max check weight 8, with distance 16 instead of 14 (kd²/n rises from 50.7 to 66.2, second in the weight-8 cell behind [[584,150,18]]).

This seeds the pair-partition CPM catalogue of Okada and Kasai (arXiv:2607.14091, new `refs.bib` entry included). The code is their girth-8 seed instance `qc_472_122_16_g8`: a (3,8)-regular quasi-cyclic CSS pair at lift size p = 59 whose Tanner graphs both have girth exactly 8, found by displacement search in the CPM-label space starting from their girth-6 [[472,122,14]] instance. The authors certify d = 16 exactly (complete connected-support exhaustion through weight 14 on both sides, plus explicit weight-16 logicals); this entry records the repo-standard witness-backed upper bound, with witnesses taken from their published bundle and re-verified here (kernel membership, row-space non-membership, GF(2) ranks 175/175 giving k = 122).

CPM exponent arrays are restated in full in `provenance.construction`, so the entry is checkable against the paper without external files; the source bundle URL and SHA-256 hashes are pinned there as well.

`family` uses a new tag `pair-partition-cpm` (the construction is neither a bicycle nor a lifted product); happy to open the vocabulary issue the verifier suggests.

Issue #507, opened 2026-08-14
Add results from arxiv.2607.14091 to the leaderboard


willzeng:
More details:

https://kasai.ict.eng.isct.ac.jp/pair_partition_cpm_css_codes_20260714.html

willzeng:
Includes #519, #520, #521, #522, #523, #524, #525, #526

PR #508, opened 2026-08-14
Add [[184,50,10]] pair-partition CPM baseline (Lee, Okada–Kasai catalogue)

willzeng:
## Code submission

- Parameters: [[n, k, d]] = [[184, 50, 10]]
- Tracks: unrestricted / weight-8

### Checklist
- [x] `python verify/qldpc_verify.py codes/184-50-10.json` passes locally

Weight-8, unrestricted locality, small-n region. This literature baseline posts kd²/n = 27.17 at n = 184 — the best weight-8 score on the board below n = 472 (the current small-n leader is `codes/240-16-20.json` at 26.67) — and its k = 50 is the highest dimension of any weight-8 entry below n = 472 (current weight-8 entries in that range top out around k = 26).

The code is PPS184, one of Jong Yeon Lee's 19 contributed CPM-PP instances in the Okada–Kasai pair-partition CPM catalogue (arXiv:2607.14091): a (3,8)-regular quasi-cyclic CSS pair at lift size p = 23. The catalogue reports d = 10 as exact, but unlike the primary instances no public machine-verified certificate bundle covers it, so the entry records only the repo-standard witness-backed upper bound — d ≤ 10 on both sides from this repo's own 150k-trial RIS search — and says so explicitly in `provenance.notes`.

Reproduction was validated end to end: the published `.cpm` exponent file, expanded with the authors' own `expand_and_verify_cpm_matrices.py` (pinned commit), yields a CSS-orthogonal (3,8)-regular pair with GF(2) ranks 67/67, confirming k = 50. The CPM exponent arrays and lift convention are restated in full in `provenance.construction`, so the entry is checkable without external files.

Same new `family` tag (`pair-partition-cpm`) and `refs.bib` entry as PR #506; whichever merges first carries the bib entry, the other rebases cleanly.


willzeng:
Also from #507

PR #509, opened 2026-08-14
Add [[84,34,6]] pair-partition CPM baseline (Okada–Kasai small-p frontier)

willzeng:
## Code submission

- Parameters: [[n, k, d]] = [[84, 34, 6]]
- Tracks: unrestricted / any weight (weight-12 checks)

### Checklist
- [x] `python verify/qldpc_verify.py codes/84-34-6.json` passes locally

Unrestricted / any-weight, small-n region. At n = 84 the board's only entry is `codes/84-6-10.json` (kd²/n = 7.14); this code posts kd²/n = 14.57 at the same size with k = 34 — a rate-0.4 code on 84 qubits. Neither dominates the other (distance 6 vs 10), so both sit on the frontier.

This is the small-p frontier instance `qc_84_34_6` from the Okada–Kasai pair-partition CPM catalogue (arXiv:2607.14091): a (4,12)-regular quasi-cyclic CSS pair at lift size p = 7. It appears only in the authors' certificate-bundle set, not in the catalogue page's main tables. The bundle is machine-verified exact (`exact_distance_machine_verified` in the master manifest): complete lower-bound exclusion through weight 5 on both sides plus explicit weight-6 logicals, giving d = 6 exactly. The witnesses here are the bundle's published vectors, re-verified in this repo (kernel membership, row-space non-membership); the entry records the repo-standard witness-backed upper bound pending this repo's own certification.

Reproduction was validated end to end: the bundle's fixed-input `.cpm` (SHA-256 pinned in `provenance.construction`) expanded with the authors' own script gives a CSS-orthogonal (4,12)-regular pair with GF(2) ranks 25/25, confirming k = 34. The local deep gate ran 5.26M accelerated RIS trials plus 35k pure-Python trials without finding any logical lighter than 6.

Same new `family` tag (`pair-partition-cpm`) and `refs.bib` entry as PRs #506/#508; whichever merges first carries the bib entry, the others rebase cleanly.


willzeng:
Updated: distance confidence flipped to `exact` on both sides, backed by the authors' machine-verified certificate bundle (complete lower-bound exclusion through weight 5, SHA-256-pinned, `exact_distance_machine_verified` in their master manifest). The board will still display d ≤ 6 until repo-side certification, per policy.

PR #595, opened 2026-08-17
Add optional fast backend for research screening

MathysRennela:
## Summary

- add an optional `gf2_fast` backend to research distance screening
- keep NumPy as the default and support `backend="auto"` fallback
- expose CPU thread control through `distance_rand` and `search.screen`
- validate accelerator-proposed witnesses with the Python GF(2) stack
- add regression coverage and concise usage documentation

## Validation

- `uv run --frozen python research/test_fast_backend.py`
- `uv run --frozen python research/test_smoke.py`
- `uv run --frozen python verify/test_check_prose.py`
- `uv run --frozen python run_tests.py --skip-slow` — 10/10 test files passed

This PR contains only the optional research backend changes.

### DIFF
--- a/research/AUTORESEARCH.md
+++ b/research/AUTORESEARCH.md
@@ -163,7 +163,9 @@ update_leaderboard("board.json", records)   # merge + persist, so a sweep can re
 description of how the code was built — so you can point it at **your own generator** for any
 family. `sample_bb` is a ready-made one; a 2BGA or coset sampler over `group_algebra` / `coset`
 has the same shape. `backend="fast"` requires `make fast`; `threads=` controls its CPU workers.
-All screening values remain upper bounds, and finalists must still pass the validation gate.
+The `trials` value is backend-specific: NumPy iterations and fast RIS samples are
+not comparable screening budgets. All screening values remain upper bounds, and
+finalists must still pass the validation gate.
 
 ## 4. Package a submission
 
--- a/research/AUTORESEARCH.md
+++ b/research/AUTORESEARCH.md
@@ -6,11 +6,9 @@ tool. It is both the **operating manual for the research loop** and the **refere
 `research/` starter kit**: constructing a code, estimating its distance, and packaging a
 verifiable submission.
 
-Everything in `research/` is **pure numpy** (no deps beyond what the verifier needs) and sits
-directly on the verifier's own GF(2) core, so what you build here is exactly what the board
-records. Your job: given a research direction, construct and search qLDPC codes and surface
-**verified, genuinely-new candidates** for human review. You may write and run any code you
-like — but you do **not** get to decide whether a code is good. A trusted gate does.
+The default path in `research/` is pure NumPy. An optional bit-packed C++ RIS backend can be
+built with `make fast` for larger screens and confirmation runs; Python still validates its
+witnesses.
 
 Layout:
 
@@ -153,6 +151,10 @@ frontier.
 from search import screen, pareto_frontier, sample_bb, update_leaderboard
 records = screen(sample_bb(400, seed=7), min_k=4, min_d=4, trials=250)
 records[:5]                       # best by k*d^2/n (the board's headline metric)
+
+# ``auto`` uses gf2_fast when available and otherwise falls back to NumPy.
+records = screen(sample_bb(10_000, seed=7), min_k=4, min_d=4, trials=2_000,
+                 backend="auto", threads=8)
 pareto_frontier(records)          # the non-dominated codes over (n, k, d)
 update_leaderboard("board.json", records)   # merge + persist, so a sweep can resume
 ```
@@ -160,8 +162,8 @@ update_leaderboard("board.json", records)   # merge + persist, so a sweep can re
 `screen` consumes any iterable of `(spec, HX, HZ)` triples, where `spec` is a JSON-serializable
 description of how the code was built — so you can point it at **your own generator** for any
 family. `sample_bb` is a ready-made one; a 2BGA or coset sampler over `group_algebra` / `coset`
-has the same shape. Because the screening distance is the surrogate *upper bound*, `screen`
-ranks **candidates** — validate the winners before claiming anything (next).
+has the same shape. `backend="fast"` requires `make fast`; `threads=` controls its CPU workers.
+All screening values remain upper bounds, and finalists must still pass the validation gate.
 
 ## 4. Package a submission
 
--- a/research/kit/search.py
+++ b/research/kit/search.py
@@ -8,11 +8,12 @@ standard three stages:
     2. screen     skip k < min_k; estimate d with the surrogate; score by k*d^2/n
     3. rank       sort by efficiency and/or extract the Pareto frontier
 
-All pure numpy. The screening distance is ``surrogate.distance_rand`` -- an
-UPPER bound -- so ``screen`` gives you ranked *candidates*, not certified codes.
-Confirm the finalists' distance (``research/kit/distance.py``: exact MILP or decoder
-corroboration) before claiming anything, then package the winner with
-``submit.make_submission``.
+The default path is pure numpy. ``surrogate.distance_rand`` also supports an
+optional bit-packed C++ RIS backend (``backend="fast"`` or ``"auto"`` after
+``make fast``). The screening distance is always an UPPER bound, so ``screen``
+gives you ranked *candidates*, not certified codes. Confirm the finalists'
+distance (``research/kit/distance.py``: exact MILP or decoder corroboration)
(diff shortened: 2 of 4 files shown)


vprusso (commented):
Found one correctness bug plus two structural comments.

The bug: the fast backend breaks `screen()`'s no-logical filter. `gf2_fast` signals "no logical found" by returning weight `n+1` with an empty side (`gf2_fast.cpp` line 374), `_validate_fast_witness` accepts that sentinel, and `_distance_rand_fast` returns it as `int(weight) = n+1`. The numpy path returns `float("inf")` in the same situation, and `screen()` filters only on `d == float("inf")`. So under `backend="fast"` or `"auto"`, any candidate whose budget finds no logical gets `d = n+1`, passes the filter, and scores `k*(n+1)^2/n`, which outranks every genuine code in the run. A low-trials sweep fills its top ranks with exactly the codes the search failed on. One-line fix in `_distance_rand_fast`: return `float("inf")` when `weight > HX.shape[1]`.

On the `sys.path.insert` blocks: this PR follows the repo's existing idiom (`cli/qldpc.py` and the verify tests do the same), so no objection here specifically, but the pattern deserves a repo-level fix: declare `verify/` and `research/kit` as packages in `pyproject.toml` so plain imports work under `uv run`, then delete every per-file insert in one pass. Happy to take that as a separate PR.

On the `try/except ImportError`: `gf2_fast` is an in-repo C++ artifact built by `make fast`, not a pip dependency, and the documented contract is that the pure-Python path produces identical gate verdicts without it, so runtime optionality is deliberate. But this is now the third module probing for `gf2_fast` with its own try/except. One shared probe (`kit/_backends.py` exposing the availability flag plus the validated wrapper) would give the fallback a single testable home.

Doc nit: `trials` means different units in the two backends (numpy counts iterations of its own loop, fast counts RIS samples), so `trials=2000` is not comparable across `backend` values and a sweep that flips backends silently changes its screening depth. Worth a sentence in the docstring.

The validation boundary itself is right: re-checking accelerator witnesses with the Python GF(2) stack, with the correct side convention, is exactly the trust structure this repo runs on.

FarLab (approved):

FarLab:
LGTM after addressing the review from @vprusso!

MathysRennela (commented):
Addressed the correctness issue identified in the review.

- `gf2_fast`'s `n + 1` no-witness sentinel is now normalized to `float("inf")`, matching the NumPy backend so `screen()` filters failed searches instead of ranking them as high-distance candidates.
- Added a regression test that simulates the no-logical-found accelerator response.
- Documented that `trials` represents different units for the NumPy and fast backends and is not a comparable screening budget across them.

Validation on the PR branch:
- `make fast`
- `uv run --frozen python research/test_fast_backend.py`
- `uv run --frozen python research/test_smoke.py`
- `uv run --frozen python verify/test_check_prose.py`
- `uv run --frozen python run_tests.py --skip-slow` — 10/10 test files passed

I left the shared backend probe and package/import cleanup out of this PR as a separate follow-up, per the scope suggestion.

vprusso (commented):
The sentinel fix and the trials documentation both look right, and pinning the inf case with a regression test is better than what I asked for. Withdrawing my shared-probe suggestion: only two modules probe for gf2_fast (research/kit/surrogate.py and verify/heuristic_distance.py) and they sit on opposite sides of the research/verify boundary, so a shared helper would buy a cross-tree import for two call sites. My "third module" count was wrong.

Four smaller things, only one of which I would block on.

The `assert _fast is not None` in `_distance_rand_fast` should go. It disappears under `python -O`, and it is unreachable anyway since the only caller already tests that condition. Either drop it or make it a real raise.

The `try/except ImportError` around the gf2_fast import should stay. It is a C++ extension built by `make fast`, not a declared dependency, and the Makefile states that pure Python is the fallback, so on a checkout that has not run `make fast` the guard is the difference between a working CLI and an import-time crash. The trailing `# optional; build with make fast` comment is redundant with the except and could go.

Type hints: I am in favour, but repo-wide rather than here. Current usage runs 0 to 4 annotations per file across qldpc_verify.py, search.py and cli/qldpc.py, so annotating one new file makes the codebase less uniform. Worth a separate PR covering the public surfaces of verify/ and research/kit together.

The sys.path.insert pair in the new test matches six existing files (verify/test_refute_gate.py, test_check_prose.py, test_heuristic_distance.py, gate_changed.py among them). The real fix is declaring verify/ and research/kit as packages in pyproject.toml so plain imports work under uv run, then deleting all of them in one pass. Singling out the newest instance while the others stand does not seem worth holding this up.

On the backend dispatch: I would leave the if-chain as it is. It reads validate, validate, dispatch-with-fallback, error, and a match statement would fragment it, since "fast" and "auto" share a branch whose behaviour depends on whether the extension is present, so you would need a guard clause and end up with more lines for no clarity gain.

vprusso (commented):
Re-reviewed at 29ddacc. The sentinel fix is correct and I ran the test on a machine with the extension built: it passes, and both backends return 6 for `[[72,12,6]]` across seeds 0 through 5 at `trials=100`, so the equality assertions are not flaky despite `trials` meaning different units per backend.

The test carries four `try` blocks and every one can go. Concrete replacements below.

The import probe at the top duplicates work `surrogate` already did, and asks a slightly different question than the one that matters:

    try:
        import gf2_fast
    except ImportError:
        gf2_fast = None

`surrogate` performs exactly this probe at import and stores the result, so the test can read `surrogate._fast is not None` and delete the block entirely. That also tests the fact the code actually branches on (did the module find the extension) rather than a parallel fact (can this file import it), which can differ if `sys.path` is ever set up differently.

The two assert-raises blocks are six lines each for one assertion:

    try:
        distance_rand(HX, HZ, trials=1, backend="invalid")
    except ValueError:
        pass
    else:
        raise AssertionError("invalid backend was accepted")

One helper collapses both call sites and confines the only remaining `try` in the file to a single place:

    def raises(exc, fn, *a, **kw):
        """True iff calling fn(*a, **kw) raises exc."""
        try:
            fn(*a, **kw)
        except exc:
            return True
        return False

then `check("invalid backend rejected", raises(ValueError, distance_rand, HX, HZ, trials=1, backend="invalid"))` and likewise with `ImportError` for the missing-extension branch. Verified both return True here.

The fourth is the monkeypatch, and it is avoidable rather than merely tidier. The thing under test is one mapping, "the extension reports no-logical as `n+1` and we must return `inf`", but testing it currently requires a fake backend class, mutating module state, and a `try/finally` to put it back:

    class NoLogicalBackend: ...
    original_backend = surrogate._fast
    surrogate._fast = NoLogicalBackend()
    try:
        assert distance_rand(HX, HZ, trials=1, backend="fast") == float("inf")
    finally:
        surrogate._fast = original_backend

Extracting the mapping into a pure function in `surrogate.py` removes all of it:

    def _weight_or_inf(weight, n):
        """gf2_fast reports 'no logical found' as n+1; NumPy reports inf."""
        return float("inf") if int(weight) > n else int(weight)

with `_distance_rand_fast` ending `return _weight_or_inf(weight, HX.shape[1])`. The test becomes one line with no patching and no global mutation:

    check("sentinel maps to inf", surrogate._weight_or_inf(n + 1, n) == float("inf"))

Confirmed the extraction behaves as expected (`n+1 -> inf`, `6 -> 6`). Net effect on the test: four `try` blocks down to one, inside a named helper, and eleven lines of the sentinel test down to one.

Separately, the test uses bare `assert` while every other test in the repo uses the accumulating `check(label, cond)` helper (`verify/test_frontier_summary.py`, `verify/test_refute_gate.py`, `verify/test_heuristic_distance.py`, `research/test_smoke.py`). Bare asserts stop at the first failure and print nothing on success, so a run that fails tells you about one problem instead of all of them, and a run that passes shows no evidence of what was covered. Worth matching the convention while the file is being touched.

vprusso (approved):
Approving. The sentinel fix is correct and pinned by a test, the `trials` ambiguity is documented in both the API docstring and AUTORESEARCH, and I ran the suite on a machine with the extension built: it passes, and both backends return 6 for `[[72,12,6]]` across seeds 0 through 5, so the equality assertions are stable.

The `sys.path.insert` pair and the four `try` blocks are worth fixing but not here. The path boilerplate is forced by `[tool.uv] package = false` plus `run_tests.py` spawning each test with the test's own directory on `sys.path`, so it is a repo-wide question rather than anything this PR introduced (I was wrong in an earlier comment to suggest declaring packages in pyproject; that contradicts an explicit decision already in the file). Filing that separately along with the test-scaffolding question. The try-block cleanups from my last review are small enough to fold into whichever PR touches this file next.

PR #602, opened 2026-08-18
Refute [[666,150,79]]: independent search finds a weight-76 logical

vprusso:
## Fixup: distance value

The `[[666,150,79]]` entry claims `d <= 79`. An independent refutation pass on separate hardware, with seeds disjoint from the original search, found lighter logical operators on both sides. The entry is renamed to `codes/666-150-76.json` with the tighter witnesses attached.

| side | was | now | witness |
|---|---|---|---|
| X | 82 | **76** | weight-76, seed 901, 40M RIS samples |
| Z | 79 | **77** | weight-77, seed 77, 20M RIS samples |

`d = min(dX, dZ)` moves 79 to 76, and `k d^2 / n` moves 1405.63 to 1300.90.

## Method

Four rungs at 5M, 20M, 40M and 40M samples, roughly 105M total. The bound fell 79, then 77, then 76, then held across the last two rungs, so 76 is where it stabilised rather than where the budget ran out. Both witnesses were re-validated locally against the shipped check matrices: each commutes with the opposite-type checks, lies outside the rowspace of its own type, and has weight exactly equal to its claimed value.

Local verification of the corrected file passes all 13 checks, signature `e82a3ccdeb40aa59`.

## Authorship

@npdeep constructed the code and stays first author; @vprusso is added for the refutation, following the precedent @MathysRennela pointed out in this thread and visible on existing entries such as `90-20-7`. Without that, `check_authorship.py` blocks any third-party correction, which is the process gap written up in #603.

## Note

No exactness is claimed. 76 remains an upper bound and is open to the same treatment. `[[666,150,81]]` is a separate entry; a measurement there is in progress and will be reported separately.


vprusso:
Closing: this cannot pass CI as constructed. `check_authorship.py` requires the PR author to be a listed `@handle` author of any added or changed code, exempting only literature baselines, so a third party cannot correct another contributor's entry whether by rename or in-place edit. Refiling the measurement as an issue instead, and raising the process gap there. The prose failure was separate and mine: the body cited `codes/666-150-79.json`, which this PR deleted.

MathysRennela:
@vprusso we solved this issue before by adding the author of the refutation as a new author. 
I feel like this should be a rule in the future as providing a refutation is a valuable contribution.

npdeep:
This is great. It might make sense to make validation more stringent for the top positions in the leaderboard. At these code sizes, it is easy to be lucky.

PR #604, opened 2026-08-19
verify: standalone GPU RIS deep-search tool (ris_gpu)

FarLab:
## What

A standalone GPU random-ISD distance-search tool, with no dependency on another Python package beyond numpy (which the repo already requires):

- **`verify/ris_gpu.cu`** — CUDA binary, built by `make ris` into `build/ris_gpu`. The two device kernels are ported verbatim from [sqetch](https://github.com/Muzhou-Ma/sqetch) by Muzhou Ma (MIT, attribution in the file header); the torch/ninja JIT host stack is replaced with plain CUDA runtime calls. Two modes: `recover` (default — ladders toward the lightest logical operator it can find and prints its support) and `estimate` (weight only).
- **`verify/ris_gpu.py`** — wrapper on numpy + stdlib: reads board code JSON, derives logical bases via the audited `verify/gf2.py`, feeds packed matrices to the binary, and **re-verifies every GPU-reported operator on the CPU** (in `ker(H_check)`, anticommutes with an opposite-type logical, weight recounted) before writing a `.witness.json`. The GPU is never trusted directly.
- **`verify/test_ris_gpu.py`** — picked up automatically by `run_tests.py`: packing layout, input-file format, output parsing, and the CPU verification gate run everywhere; the GPU end-to-end check (Steane code, expects exactly d=3) auto-skips when there is no device, so CI is unaffected.

## Why

Deep RIS refutation of high-n board claims was bottlenecked on CPU budgets (2M-trial convergence checks, minutes to hours). The sqetch stack runs ~3.2M trials/s on an L40S but its torch `load_inline` JIT needed four environment fixes on a fresh cloud box before it ran once. This tool keeps the kernel and drops the fragile stack: one `nvcc` invocation at build time, zero runtime dependencies.

Like `gf2_fast.cpp`, this is an optional accelerator — the pure-Python verifier remains the reference and nothing in the trusted validation path changes. All reported weights are upper bounds; witnesses, not runs, are the evidence.

## Validation

On an L40S (nvcc 13.0, `-arch=native`):
- **[[112,6,7]]** (exact DRAT-certified d=7): witnessed dX=7, dZ=8 — matches the certificate exactly.
- **[[674,128,87]]** at a 50M-trial budget per side, versus sqetch at the same budget: X witness 86 (sqetch: 87), Z witness 90 (sqetch seeds ranged 84–87) — seed-level parity, all witnesses CPU-verified.
- Full local test suite: 10/10 files pass; the new GPU test passes on-device and skips cleanly off-device.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,13 @@
-.PHONY: build verify example test test-fast fast
+.PHONY: build verify example test test-fast fast ris
 
 build:
 	uv run python site/build.py
+
+# Optional GPU RIS deep-search binary (verify/ris_gpu.cu); needs nvcc + an
+# NVIDIA GPU. Driven by verify/ris_gpu.py, which CPU-verifies its output.
+ris:
+	mkdir -p build
+	nvcc -O3 -arch=$(or $(RIS_ARCH),native) -o build/ris_gpu verify/ris_gpu.cu
 
 # Optional C++ RIS accelerator (verify/gf2_fast.cpp); pure Python is the fallback.
 fast:
--- a/verify/ris_gpu.cu
+++ b/verify/ris_gpu.cu
@@ -1,0 +1,573 @@
+/*
+ * ris_gpu -- standalone GPU random-information-set distance search for CSS
+ * codes. OPTIONAL deep-search accelerator, like gf2_fast.cpp: the pure-Python
+ * verifier stays the reference, and every operator this tool reports is
+ * re-verified on the CPU by verify/ris_gpu.py before anyone believes it.
+ *
+ * The two device kernels are ported verbatim from sqetch by Muzhou Ma
+ * (https://github.com/Muzhou-Ma/sqetch, MIT License, Copyright (c) 2026
+ * Muzhou Ma). The host side replaces sqetch's torch JIT wrappers with plain
+ * CUDA runtime calls so the binary has no Python or torch dependency.
+ *
+ * Build:   make ris          (nvcc -O3 -arch=native -o build/ris_gpu ...)
+ * Input:   packed matrix file written by verify/ris_gpu.py ("RISGPU01"
+ *          header; W_null = basis of ker(H_check), W_logical = opposite-type
+ *          logical basis; rows bit-packed LSB-first into uint64 words).
+ * Output:  key=value lines on stdout; recover mode adds "support=..." with
+ *          the qubit indices of the best logical operator found.
+ *
+ * Each trial draws a random column permutation and a k_sub-row random sketch
+ * of W_null, row-reduces the sketch in permuted column order, and keeps the
+ * lightest row that anticommutes with a logical -- i.e. a random-ISD trial.
+ * Weights found are upper bounds on d; nothing here is a proof.
+ */
+
+#include <cuda_runtime.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <string>
+#include <vector>
+
+#define BLOCK_SIZE 128
+
+#define CUDA_CHECK(call)                                                    \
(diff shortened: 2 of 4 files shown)


vprusso (commented):
The trust boundary is the right shape and I checked it holds: `cpu_verify` accepts an operator only if it is in `ker(H_check)` and anticommutes with at least one opposite-type logical. That second condition is what rules out stabilizers, since a stabilizer commutes with every logical, so it is equivalent to the usual "not in the rowspace" test and it errs safe if `L_opp` is ever an incomplete basis: a real logical could be rejected, a stabilizer can never be accepted. Never trusting the GPU output directly is exactly right, and porting the kernels with attribution rather than taking a torch/ninja dependency is a good trade for this repo.

`witness_weight` is recorded as `len(support)` rather than the weight of the vector that was actually verified, which is a bug:

    if cpu_verify(support, H_check.shape[1], H_check, L_opp):
        entry["witness_weight"] = len(support)

`cpu_verify` builds `v[support] = 1`, so a duplicated index collapses. Checked it: support `[3,3,5]` gives `len(support) = 3` but `int(v.sum()) = 2`. The module docstring says "weight recounted", which is what it should be doing, and the fix is one line: recompute `int(v.sum())` and use that. The direction is conservative, since an overcount weakens the bound rather than inflating it, but the board verifier requires the witness weight to equal the claimed value exactly, so a `.witness.json` carrying a padded weight would be rejected downstream with a confusing message.

`v[support] = 1` also raises `IndexError` on an out-of-range index rather than returning False, so a malformed GPU line crashes the tool instead of being discarded like a failed verification. Worth folding a bounds check into `cpu_verify` next to the emptiness check, since it is the same class of defensive input handling.

`subprocess.run(cmd, capture_output=True, text=True, check=True)` has no `timeout=`. A wedged GPU binary hangs the wrapper indefinitely with no output, which on a long board sweep is indistinguishable from slow progress. The temp-file cleanup in the `finally` is correctly placed.

`test_cpu_verify_gate` covers the stabilizer-rejection case, which is the important one. The duplicate-index and out-of-range cases are worth adding alongside it once the weight fix lands, since both are things a GPU kernel can plausibly emit and neither is currently exercised.

For context on why this matters beyond convenience: five codes in the `n = 666, k = 150` family have now been measured at two budgets here, and all of them report `d <= 81` or `82` from a 2M-sample CPU screen while landing at 76 to 78 under 20M or more. A tool that makes deep budgets cheap is the direct answer to that, so this is worth landing.

FarLab:
Thanks for the careful read — all three landed in 63f1567:

1. **Weight recount**: `cpu_verify` now returns the recounted weight of the operator it actually verified (or `None`), and `witness_weight` uses that return value — `len(support)` is gone. I went one step stricter than the suggested fix: a support with duplicate indices is *rejected* rather than deduplicated-and-recounted, since the GPU's bit-packed vector can never legitimately produce duplicates, so their presence means the line is corrupt and shouldn't be trusted at any weight.
2. **Bounds check**: out-of-range (including negative) and empty supports are rejected in the same guard, next to the duplicate check — malformed output is discarded like any failed verification instead of crashing the sweep.
3. **Timeout**: the GPU subprocess now runs under `--proc-timeout` (default 7200s), so a wedged binary fails loudly instead of impersonating slow progress.

Tests now exercise the recount (`== len(found)` on a clean support), duplicate, out-of-range, negative, and empty cases alongside the existing stabilizer rejection.

And yes — the n=666 family numbers you cite are exactly the motivation. This week's sweep of everything above kd²/n = 500 (#605–#610) found six of seven claims inflated, all at n ≥ 480 where the CI screen can't converge; #611 has a related process proposal.

vprusso (approved):

Issue #611, opened 2026-08-19
RFC: credit refutations on the witness, not in the code's author list


FarLab:
Current precedent (the [[666,150]] correction, followed by #605–#610) appends the refuter to `provenance.authors`. After today's sweep — one person refuting five codes by four different constructors — the problems are concrete:

- **Semantics**: `authors` should answer "who constructed this code?" A refuter corrected one metadata field, yet now appears as an author on codes they didn't design — credited, oddly, for *lowering* their scores.
- **Scaling**: distances get revised repeatedly ([[666,150]]: 82→79→76 in three days). Every revision appending an author turns the list into an indistinguishable mix of construction and correction credit.
- **Incentives**: refuting a rival's code putting your name on their entry is a convention that gets weirder the more it's used.

**Proposal**: attach attribution to the witness itself:

```json
"Z": {
  "value": 36,
  "confidence": "upper_bound",
  "witness": [...],
  "witness_provenance": {"found_by": ["@FarLab"], "date": "2026-08-19", "method": "ris_gpu 1B-trial RIS"}
}
```

This credits exactly what was contributed (that operator, that method, that date), survives repeated revisions cleanly, and leaves the constructor sole authorship. The site can render "distance tightened by …" separately — arguably better visibility for refutation work than mid-author-list placement.

Cost: `schema_version` bump, validator tweak, site rendering change. If accepted, the six recent refutation PRs can be migrated in one follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

FarLab:
Correction after reviewing #602 (h/t @FarLab): appending the refuter to `authors` isn't only a credit convention — it's load-bearing for CI. `verify/check_authorship.py` requires the PR author to be listed in `provenance.authors` of every code they add or change (anti-impersonation binding), which is exactly why #602 was closed and why #605–#610 add the refuter. Moving credit to `witness_provenance` alone would make third-party correction PRs fail CI again.

So the proposal needs two halves:

1. **Credit**: `witness_provenance.found_by` on the per-side distance entry, as above.
2. **Authorship binding**: extend `check_authorship.py` so the PR author may alternatively appear in the `found_by` of a witness added/changed by the PR. The anti-impersonation property is preserved — you can only touch someone else's entry while claiming a witness, and the witness itself is machine-verified, so a bogus claim can't pass. (Optionally, such PRs could be restricted to touching only the `distance` block and `notes`.)

[...]

PR #610, opened 2026-08-19
Refute [[514,162,41]]: witnesses at weight 29 and 29, score 529.83 to 265.06

FarLab:
## Summary

GPU deep refutation of the [[514,162,41]] prime-index cyclic generalized-bicycle code on Z_257 (2026-08-11) — the largest correction of the sweep:

| side | claimed | witnessed |
|---|---|---|
| X | 41 | **29** |
| Z | 44 | **29** |

d<=41 tightens to **d<=29**; score 162·41²/514 = **529.83** → 162·29²/514 = **265.06**. Both weight-29 witnesses are recorded in the JSON.

## Method

Standalone `ris_gpu` tool (#604) on an NVIDIA L40S: 1B-trial random-ISD recovery ladder per side (seed 88000007). Both witnesses CPU re-verified with `verify/gf2.py` on the box and again locally; `verify/qldpc_verify.py` passes all checks on the updated JSON.

Completes the 500 < kd²/n < 1000 sweep (#608, #609): all three band codes were refuted, and only [[502,102,50]] at 507.97 remains in the band afterward. Distances remain witness-backed upper bounds. @FarLab added as an author for the refutation per board precedent; original constructor retained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #612, opened 2026-08-19
schema 0.2: witness_provenance — refutation credit lives on the witness

FarLab:
## What

First half of #611 (PR A of two). Adds an **optional** `witness_provenance` block to each per-side distance entry:

```json
"Z": {
  "value": 36,
  "confidence": "upper_bound",
  "witness": [...],
  "witness_provenance": {
    "found_by": ["@FarLab"],
    "date": "2026-08-19",
    "samples": 1000000000,
    "tool": "ris_gpu",
    "seeds": [88000007]
  }
}
```

- `schema_version` becomes `enum ["0.1", "0.2"]` — every existing code remains valid unchanged.
- `found_by`, `date`, `samples` required within the block; `tool`/`seeds` optional. Per the discussion, `samples` is deliberately mandatory: the budget a bound has survived is the load-bearing evidence (the n=666 family reports d≤81 at 2M trials regardless of which code you hold), and it tells the next refuter the budget to beat. `tool` stays free-text but optional.
- Detail pages render `witness found by @x · ris_gpu · survived 10⁹ trials · 2026-08-19` under the side's distance line.
- Tests: acceptance of a 0.2 block, rejection of a block missing `samples`, rejection of a handle without `@`.

## What this does NOT do

- No change to `check_authorship.py` — third-party refutation PRs still bind via the author-append precedent for now. The scoped `found_by` binding (diff confined to the distance block + strictly-lighter witness, per @vprusso's conditions) is PR B.
- No migration of #605–#610 — that follows once PR B lands, so migrated entries don't fail the authorship gate in the meantime.

### DIFF
--- a/schema/SCHEMA.md
+++ b/schema/SCHEMA.md
@@ -42,14 +42,20 @@ Two principles drive the format:
   - `witness`: support of a logical operator of that Pauli type and weight
     `value`. An X-witness must lie in `ker(H_Z)` and outside `rowspace(H_X)`;
     the Z-witness mirrors it. This is what makes the upper bound trustless.
-  - `witness_provenance` (optional, schema 0.2, issue #611): who found this
-    witness, when, and at what budget — `found_by` (list of `@handles`),
-    `date`, `samples` (total search trials), optional `tool` and `seeds`.
-    Refutation credit lives here, attached to the operator contributed,
-    rather than in `provenance.authors`, which stays reserved for the code's
-    constructors. `samples` is the load-bearing number: a bound that has
-    survived 10^9 trials is materially stronger evidence than the same value
-    backed by 10^6, and it tells the next refuter the budget to beat.
+  - `witness_provenance` (optional; requires `schema_version: "0.2"`, and the
+    schema enforces that; issue #611): who found this witness, when, and at
+    what budget — `found_by` (list of `@handles`), `date`, `found_at_samples`,
+    optional `survived_samples`, `tool`, and `seeds`. Refutation credit lives
+    here, attached to the operator contributed, rather than in
+    `provenance.authors`, which stays reserved for the code's constructors.
+    The two budgets are deliberately separate because they are opposite kinds
+    of evidence: `found_at_samples` is the budget the witness turned up at and
+    says nothing about whether something lighter exists, while
+    `survived_samples` is the largest budget a deep search has spent on this
+    side without finding anything lighter — only a null result may write it.
+    `survived_samples` is what actually constrains the distance and tells the
+    next refuter the budget they need to beat; a deep sweep that finds nothing
+    records its null result by raising it.
 - `locality` (optional): provide a layout and the verifier derives the locality
   class (`local-2d-single`, `local-2d-bilayer`, or `unrestricted`); omit it and
   the code is `unrestricted`.
--- a/schema/SCHEMA.md
+++ b/schema/SCHEMA.md
@@ -24,7 +24,8 @@ Two principles drive the format:
 
 ## Fields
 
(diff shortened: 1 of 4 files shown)


vprusso (commented):
The shape is right and I am glad `samples` is mandatory. One substantive problem with it, plus two smaller notes.

The `samples` field currently means two different things and the schema description says so explicitly: "the budget it was found at, or the largest budget the bound has survived". Those are opposite kinds of evidence. "Found at 2e6" says someone searched a little and got lucky; it implies nothing about whether something lighter exists. "Survived 2e7 without falling" says someone searched hard and failed to beat it, which is the claim that actually constrains the distance. Collapsing both into one integer loses precisely the distinction that motivated the field.

Our own data makes the difference concrete. The `[[666,150,81]]` witness was found at roughly 2e6 trials. We then spent about 1.05e8 on that code and found a weight-77 operator, so the honest record for the original witness is "found at 2e6, did not survive 2e7", and under this schema it would be written as `samples: 2000000` and rendered as though that were its strength.

The rendering makes the conflation user-visible, since `detail_page` hardcodes the surviving reading:

    parts.append(f'survived {sci_int(wp["samples"])} trials')

A witness that was merely found at that budget gets displayed as having survived it, which overstates the evidence in exactly the direction the field exists to prevent. Two fields would fix it (`found_at_samples` required, `survived_samples` optional and only written by a deep pass that found nothing lighter), and it gives the weekly sweep somewhere to record a null result, which today is invisible. If you would rather keep one field, then define it as the surviving budget only, make the found-at case omit it, and render accordingly.

Smaller: `schema_version` becomes `enum ["0.1", "0.2"]` but nothing gates the new block on the version, so a document declaring `0.1` can carry `witness_provenance` and still validate. That makes the version field decorative from here on. Either add a conditional so `0.2` is required when the block is present, or say in SCHEMA.md that the version is informational and not enforced, so the next person adding a field knows which convention they are following.

`sci_int` behaves correctly on the values that matter (2e6, 2e7, 1e9 render as expected, 1.05e8 as 1.1x10^8, sub-10^4 verbatim). The only wrinkle is that a value like 1000000001 prints as `1x10^9` rather than `10^9` because the mantissa check is exact float equality, which is cosmetic and probably not worth code.

MathysRennela (approved):
LGTM!

FarLab:
All three addressed in c83f92c:

1. **Split budgets**: `found_at_samples` (required) and `survived_samples` (optional). The schema descriptions carry your distinction verbatim in spirit — found-at "says nothing about whether something lighter exists"; survived may only be written by a null result and "is what a later refuter needs to beat". The detail page now renders `found at 5×10⁷ trials` always and appends `survived 10⁹ trials` only when earned, so the [[666,150,81]]-style case can no longer display its finding budget as strength. A weekly sweep records a null result by raising `survived_samples` — the previously-invisible outcome now has a home.
2. **Version gating**: a root-level conditional requires `schema_version: "0.2"` whenever either side carries `witness_provenance`, so the version field stays load-bearing. Test added: a 0.1 document with the block is rejected.
3. **`sci_int`**: took the cosmetic fix anyway — it was one line moved (compare the *rounded* mantissa against "1"), so 1000000001 renders as 10⁹.

Tests: block accepted with/without `survived_samples`, rejected without `found_at_samples`, rejected on 0.1, handle-without-@ rejected. Full suite green, site builds.

One knock-on worth noting for PR B (#613): the binding there reads only `found_by`, so the field split doesn't touch it, but its test fixtures used `samples` — I'll rename them for consistency when addressing any review over there.

vprusso (approved):

PR #613, opened 2026-08-19
authorship gate: witness_provenance refutation binding (#611 PR B)

FarLab:
## What

Second half of #611 (PR A: #612). `check_authorship.py` gains an alternative binding for third-party refutations, scoped exactly as @vprusso specified:

A PR author who is **not** in a code's `authors` binds iff:
1. **Confinement**: every difference from the base version is inside the `distance` block, plus the `name` string, `schema_version`, and an *append* to `provenance.notes`. Checks, n, k, authors, construction, model, family — byte-identical or rejected.
2. **Strict tightening**: at least one side's `value` strictly decreases, and every side the PR touches carries a `witness_provenance` whose `found_by` lists the PR author. A side may alternatively carry a survival stamp (provenance recorded, value and witness untouched) riding along with a real refutation.

Renamed files (d is in the filename) are paired with the deleted `codes/<n>-<k>-*.json` from the same diff. The witness itself is machine-verified by the distance checks, so the binding never has to trust the claim — together the conditions mean this path can only tighten a bound, never alter a construction or its authorship.

## Also: closes an author-swap hole

While wiring this up, the tests surfaced a pre-existing gap the new binding would have inherited: the gate only ever read the **new** file's author list, so modifying someone's code and swapping your own handle into `authors` passed. Now, on a change to an existing code, the author list may only be edited by someone already on it (new submissions unaffected; git/parse failures still fail open per the module's philosophy). Note this intentionally retires the append-yourself-as-author refutation path (#602 discussion, used by #605–#610) — `witness_provenance` becomes the sanctioned route, which is the point of #611.

## Tests

New `verify/test_check_authorship.py` (auto-discovered by `run_tests.py`) builds throwaway git repos and exercises 15 cases: clean rename + in-place refutations bind; missing/wrong credit, checks edits, author appends/swaps, notes rewrites, loosened bounds, and stamp-only changes are all rejected; owner flows are unchanged. Full suite 10/10.

Sequencing: independent of #612 code-wise (the gate only reads `found_by`), but both should land before the #605–#610 entries are migrated off the author-append style.

### DIFF
--- a/verify/check_authorship.py
+++ b/verify/check_authorship.py
@@ -6,6 +6,21 @@ GitHub user opening the PR is listed among the @handle authors of every code
 they add or change. Literature baselines (no @handle authors) are exempt, and a
 maintainer submitting on someone's behalf just adds themselves as a co-author.
 
+Refutation binding (issue #611): a PR author who is NOT an author of a code may
+still tighten its distance claim, provided the change is exactly a refutation
+and nothing else. The alternative binding holds when every difference from the
+base version is confined to the distance block (plus the name string and an
+append to provenance.notes), every witness_provenance the PR adds lists the PR
+author in found_by, and at least one side's value strictly decreases. Together
+these mean the path can only ever be used to tighten a claim, never to alter a
+construction or its authorship -- the witness itself is machine-verified by the
+distance checks, so a bogus claim cannot pass regardless.
+
+Relatedly, on a change to an existing code the author list itself may only be
+edited by someone already on it: without that, adding (or swapping in) your own
+handle would grant edit rights over anyone's entry, which is the hole the
+refutation binding exists to avoid. New submissions are unaffected.
+
 Fails CLOSED only on a definite author/PR-author mismatch. Anything ambiguous
 (no author info, git/parse error) fails OPEN with a warning -- a bug here must
 not block legitimate contributions, since impersonation is lower-stakes than the
@@ -26,15 +41,20 @@ HANDLE = re.compile(r"^@([A-Za-z0-9-]+)$")
 
 
 def changed_codes(base, root=ROOT):
+    """{path: status} for codes changed vs base; status is A, M, or D."""
     try:
         out = subprocess.check_output(
-            ["git", "diff", "--name-only", "--diff-filter=AM",
-             f"{base}...HEAD", "--", "codes"],
+            ["git", "diff", "--name-status", f"{base}...HEAD", "--", "codes"],
             cwd=root, text=True)
     except Exception as e:
         print(f"(could not diff vs {base}: {e}); skipping authorship check")
         return None
-    return [f for f in out.split() if f.endswith(".json")]
+    changes = {}
+    for line in out.splitlines():
+        parts = line.split("\t")
+        if len(parts) >= 2 and parts[1].endswith(".json"):
+            changes[parts[1]] = parts[0][:1]
+    return changes
 
 
 def handles(doc):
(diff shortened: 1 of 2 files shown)


vprusso (commented):
The confinement plus strict-tightening structure is what I had in mind and the implementation is tighter than my sketch: protecting every provenance key except an append-only `notes`, rejecting value increases, requiring `found_by` on any side that moves, and pairing renames through the deleted counterpart all close the paths I would have worried about. Catching the pre-existing author-swap hole while wiring it up is a good find, and I confirmed the fix works: adding your handle to a code that already has `@handle` authors now falls through to the refutation binding, where the unchanged-`provenance` rule rejects it.

Two gaps, both of which I checked by calling `refutation_binding` directly rather than reading.

Refuting an `exact` claim is blocked, and it is the case that matters most. `if bs.get("confidence") != ns.get("confidence"): return False` means a correction that lowers the value must keep the old confidence string. Concretely, on a code claiming `d = 4 (exact)` where a weight-3 operator turns up:

    refute an exact claim, confidence -> upper_bound : (False, 'distance.X.confidence changed')
    same but confidence left as 'exact'             : (True, '')

So the honest correction is rejected and the dishonest one binds. An exact claim that has just been falsified is exactly the claim whose tier must change, since the lower-bound proof it asserted is now known to be wrong. Suggest allowing the transition `exact -> upper_bound` on a side whose value strictly decreases, while continuing to reject every other confidence change, in particular any upgrade toward `exact`.

The author-swap fix does not cover baselines with no `@handle` authors. `base_hs` is empty for a literature entry, so `if (base_hs and hs != base_hs and author not in base_hs)` is false and the else branch accepts. Checked:

    base handles: []      new handles: ['bob']
    adding @bob to a no-handle baseline and editing n -> PASSES (treated as listed author)

That edit changed `n` and never reached `refutation_binding`. The behaviour predates this PR, since baselines were exempt before, but the PR's stated purpose is closing exactly this class, and the residual path grants full edit rights rather than the scoped ones. Treating an empty base author list as "no existing author to authorise the change" and falling through to the refutation binding would close it, at the cost that a genuine first `@handle` claim on a baseline needs a maintainer, which seems like the right side to err on.

Smaller: the `d` consistency check reads `min(nd.get(s, {}).get("value", 0) for s in ("X", "Z"))`, so a document missing a side yields `min(..., 0) = 0` and requires `d == 0`. Every board code carries both sides today, so this is latent rather than live, but it fails with a misleading message rather than saying a side is missing.

MathysRennela (approved):
LGTM!

vprusso (approved):

FarLab:
All three fixed in cb93db4:

1. **Exact-claim refutation**: a side whose value strictly decreases must now carry `confidence: upper_bound`. Your two probe cases flip to the right answers — the honest demotion binds, keeping `exact` at the corrected value is rejected — and the general rule falls out: certification is the only route back to `exact`, so upgrades are rejected whether they ride on a correction or on a survival stamp (both tested).
2. **Baseline handle claims**: the author-list guard now reads an empty base handle list as "no existing author can authorise the change" rather than as an exemption, so your probe (add `@bob` to a no-handle baseline and edit `n`) falls through to the refutation binding and is rejected there. A pure distance refutation of a baseline that leaves its authors untouched still passes via the no-handle exemption, so refuting literature entries stays possible. First `@handle` claims on baselines now need a maintainer — agreed that's the right side to err on.
3. **Missing side**: named explicitly (`distance.Z is missing`) before the d-consistency check can produce its misleading message.

Test suite grew from 15 to 22 cases, including both of your probes verbatim. Thanks for checking by execution rather than by reading — both gaps were invisible in the diff.

PR #615, opened 2026-08-19
Test scaffolding: adopt pytest, centralise import paths (#614)

### LINKED ISSUE #614: Test scaffolding: 10 files hand-roll what pytest provides, and it forces the sys.path hack
Splitting this out of #595, where the visible symptom was two `sys.path.insert` lines in a new test.

## The state

10 test files under `verify/` and `research/`, discovered by glob in a 63-line `run_tests.py` that spawns each as a subprocess. Across them:

- 14 `sys.path.insert` lines
- 8 files defining their own `check(label, cond)` helper
- 6 files carrying their own `_fail` accumulator
- 10 `if __name__ == "__main__"` guards

That is roughly 53 lines of scaffolding reimplementing discovery, per-test isolation, assertion reporting and failure accumulation. pytest provides all four, and `conftest.py` provides the path setup that is currently pasted into every file.

## Why the path hack is not the test's fault

`pyproject.toml` sets `[tool.uv] package = false` ("scripts-only repo: manage the environment, do not try to build/install it"), so nothing is importable by name. `run_tests.py` runs `subprocess.run([sys.executable, ROOT/test], cwd=ROOT)`, which puts the *test file's own directory* on `sys.path[0]`, not the repo root. So `from bb import ...` and `import gf2_fast` cannot resolve without help.

Verified both directions: stripping the inserts from `research/test_fast_backend.py` fails with `ModuleNotFoundError: No module named 'bb'`, and the same stripped file passes when `PYTHONPATH` supplies `verify/` and `research/kit`.

So the options are real ones, not cosmetic:

1. **Centralize the path setup** in `run_tests.py` via `PYTHONPATH` and delete the boilerplate. Cheapest, but running one test directly (`python research/test_fast_backend.py`, which is how you debug a single failure) then breaks.
2. **Adopt pytest** with a `conftest.py` doing the path setup once. Kills all four categories of boilerplate, gives assertion introspection so `check()` and `_fail` disappear, keeps single-test invocation (`pytest research/test_fast_backend.py::test_x`), and is what every contributor already knows. Cost: one dev dependency, which does not touch the verifier's runtime dependency surface, and that surface is the one the trustless-verification promise actually rests on.
3. **Make the repo installable**, reversing `package = false`. Solves imports everywhere, but it is a decision about what this repo is rather than a cleanup.

## Suggestion

Option 2, keeping `run_tests.py` as a thin wrapper so CI and the documented `uv run --frozen python run_tests.py` entry point keep working. Migration can be incremental since pytest collects plain `test_*` functions, so files can move over one at a time without a flag day.

Worth deciding before the hackathon rather than after: item 3 of the hackathon plan is a fresh-eyes submission sprint, and "how do I run and debug one test" is exactly the kind of friction that surfaces there.

vprusso:
Closes #614.

## The problem

11 test files carried 14 `sys.path.insert` lines between them, plus 8 private `check()` helpers, 6 `_fail` accumulators and a 63-line hand-rolled runner. The path boilerplate is not the tests' fault: `[tool.uv] package = false` means `verify/` and `research/kit/` are plain directories, and `run_tests.py` spawned each test with the test's own directory on `sys.path`, so `from bb import ...` could not resolve without help.

## What changed

- `conftest.py` at the root does the path setup once, before collection. The 14 inserts are gone.
- pytest added as a dev dependency group, so the verifier's runtime dependencies (numpy, jsonschema) are untouched. That surface is the one the trustless-verification promise rests on.
- `run_tests.py` is now a thin wrapper over pytest with the same CLI, so CI and the documented command do not change. Discovery is still by convention.
- Running or debugging a single test is now `pytest verify/test_verifier.py::test_main`, with `-k`, `-x` and `--pdb` available. This replaces `python verify/test_x.py`, which is a deliberate trade: the standalone-script property is dropped in exchange for a real runner.

## Two things worth knowing

`verify/test_gf2_fast.py` executed at module level and called `sys.exit()` twice. pytest hits that during collection, so it would have aborted the whole session. It now uses `pytest.skip(allow_module_level=True)` when the extension is unbuilt, and asserts its accumulated verdict.

Two entry points did not match the contract I first assumed. `test_receipt.main()` returns `None` and signals failure by raising; `test_heuristic_distance.main(argv)` takes arguments and calls `sys.exit` even on success. Both are now asserted against what they actually do. Both had been "passing" purely because the old runner checked the process exit code and never observed `main()`'s return value, so a `main()` that returned `None` looked identical to one that returned 0.

## Deliberately not in scope

The 8 duplicated `check()` helpers and 6 `_fail` accumulators stay. Sharing them is a trap now that all files run in one process: a shared mutable `_fail` would leak failures between files. Removing them properly means splitting each `main()` into granular `test_*` functions so pytest's assertion introspection does the job, which is per-file work better reviewed as separate diffs.

## Verification

Full suite run locally through the unchanged entry point (`uv run --frozen python run_tests.py`). CI runs the same command.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,17 @@
+"""Import paths for the test suite.
+
+This repo is deliberately not installable (`[tool.uv] package = false`), so
+`verify/` and `research/kit/` are plain directories rather than packages and
+nothing in them is importable by name. Every test used to open with its own
+`sys.path.insert` pair to work around that. pytest imports this file before
+collecting anything, so the fix belongs here once instead of in every test.
+"""
+import os
+import sys
+
+ROOT = os.path.dirname(os.path.abspath(__file__))
+
+for _rel in ("verify", os.path.join("research", "kit"), "research", "cli", "site"):
+    _p = os.path.join(ROOT, _rel)
+    if os.path.isdir(_p) and _p not in sys.path:
+        sys.path.insert(0, _p)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,3 +21,8 @@ research = ["scipy>=1.11", "ldpc>=2.0"]
 [tool.uv]
 # scripts-only repo: manage the environment, do not try to build/install it
 package = false
+
+[dependency-groups]
+dev = [
+    "pytest>=8",
+]
--- a/research/test_smoke.py
+++ b/research/test_smoke.py
@@ -13,9 +13,6 @@ Run: uv run python research/test_smoke.py   (exit 0 = pass)
 import os
 import sys
 
-_HERE = os.path.dirname(os.path.abspath(__file__))
-sys.path.insert(0, os.path.join(_HERE, "kit"))              # research/kit modules
-sys.path.insert(0, os.path.join(_HERE, "..", "verify"))    # the verifier
 
 from bb import build_bb, KNOWN
 from css import compute_k, verify_css
@@ -67,6 +64,11 @@ def main():
 
     print("PASS" if not _fail else "FAIL: " + ", ".join(_fail))
     return 0 if not _fail else 1
+
+
+def test_main():
+    """pytest entry point; the suite body lives in main()."""
+    assert main() == 0
 
 
 if __name__ == "__main__":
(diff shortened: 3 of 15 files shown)


FarLab (approved):

MathysRennela (approved):
LGTM!

PR #608, opened 2026-08-19
Refute [[502,102,56]]: witnesses at weight 50 and 50, score 637.15 to 507.97

FarLab:
## Summary

GPU deep refutation of the [[502,102,d<=56]] designed-divisor generalized-bicycle code (2026-08-13). Both sides fall to weight 50:

| side | claimed | witnessed |
|---|---|---|
| X | 56 | **50** |
| Z | 57 | **50** |

d<=56 tightens to **d<=50**; score 102·56²/502 = **637.15** → 102·50²/502 = **507.97**.

## Method

First deployment of the standalone `ris_gpu` tool (#604) end-to-end: 1B-trial random-ISD recovery ladder per side (seed 88000007) on an NVIDIA L40S, ~2-minute box setup (nvcc + numpy, no torch). Witnesses CPU re-verified with `verify/gf2.py` on the box and again locally; `verify/qldpc_verify.py` passes all checks on the updated JSON.

Part of the sweep of all three 500 < kd²/n < 1000 codes (with #609-to-follow for the other two). Distances remain witness-backed upper bounds. @FarLab added as an author for the refutation per board precedent; original constructor retained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #616, opened 2026-08-19
authorship gate: renames must not bypass the check

FarLab:
## The bug

Found while migrating #608–#610 to the new schema: the gate merged in #613 **fails open on renamed code files** — the exact case it was built for.

`changed_codes` parses `git diff --name-status` for `A`/`M` lines, but git folds a rename plus a small edit into a single `R` line when similarity is high. A real refutation renames `codes/<n>-<k>-<d>.json` to the new d while touching only the distance block — `R098` (98% similar) on an actual board file. The parser recognizes neither field of the `R` line, so the gate sees no changed files, prints "no submissions to check", and passes vacuously. Anyone could rename-and-edit another person's entry and skip authorship checking entirely (schema and distance checks still apply, but the authorship property is gone).

#613's 22 test cases missed it because the fixtures were tiny single-line JSONs — below git's 50% similarity threshold, so their renames decomposed into `D`+`A` naturally and the pairing logic worked. Live demonstration: the authorship step on #608–#610 is currently passing without inspecting them.

## The fix

- `--no-renames` on the diff, so renames always decompose into `D` + `A` — which `base_counterpart` already pairs correctly.
- Defensive parsing of `R old new` lines anyway (old → `D`, new → `A`), in case the flag is ever lost.
- Two new test cases on a padded, multi-line fixture (500 checks, `json.dump(indent=1)`) that genuinely triggers git's rename folding: the clean refutation binds, the tampered one is caught. 24/24; full suite 11/11.

cc @vprusso — smallest possible follow-up to your #613 review; the binding logic itself is untouched.

### DIFF
--- a/verify/check_authorship.py
+++ b/verify/check_authorship.py
@@ -44,10 +44,18 @@ HANDLE = re.compile(r"^@([A-Za-z0-9-]+)$")
 
 
 def changed_codes(base, root=ROOT):
-    """{path: status} for codes changed vs base; status is A, M, or D."""
+    """{path: status} for codes changed vs base; status is A, M, or D.
+
+    --no-renames matters: a refutation renames codes/<n>-<k>-<d>.json to the
+    new d while changing only the distance block, which git otherwise folds
+    into a single R line (98% similar on a real board file) that used to
+    parse as neither added nor modified -- the gate then checked nothing at
+    all. Splitting renames back into D + A keeps every path visible; the R
+    branch below is belt-and-braces in case the flag is ever lost."""
     try:
         out = subprocess.check_output(
-            ["git", "diff", "--name-status", f"{base}...HEAD", "--", "codes"],
+            ["git", "diff", "--name-status", "--no-renames",
+             f"{base}...HEAD", "--", "codes"],
             cwd=root, text=True)
     except Exception as e:
         print(f"(could not diff vs {base}: {e}); skipping authorship check")
@@ -55,8 +63,16 @@ def changed_codes(base, root=ROOT):
     changes = {}
     for line in out.splitlines():
         parts = line.split("\t")
-        if len(parts) >= 2 and parts[1].endswith(".json"):
-            changes[parts[1]] = parts[0][:1]
+        if len(parts) < 2:
+            continue
+        status = parts[0][:1]
+        if status == "R" and len(parts) >= 3:
+            if parts[1].endswith(".json"):
+                changes[parts[1]] = "D"
+            if parts[2].endswith(".json"):
+                changes[parts[2]] = "A"
+        elif parts[1].endswith(".json"):
+            changes[parts[1]] = status
     return changes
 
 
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Reproduced against `codes/666-150-76.json` in a throwaway repo: rename it to `666-150-73.json`, attach a lighter Z witness credited to `@eve`, who is on nobody's author list. git folds it to `R099`, and main's gate exits 0 with "no added/changed code submissions to check" even when the same commit rewrites `provenance.authors` to `["@eve"]` and replaces `provenance.construction`. So the consequence is not just that authorship goes unchecked, it is that an entry can be taken over outright. This branch exits 1 on that case and binds the clean refutation correctly.

Merges cleanly on top of #615 (pytest scaffolding), and the `test_main` entry point survives the merge.

The `indent=1` line is doing more work than the two new cases suggest, and I would put it in the commit message rather than in a trailing comment. It retrofits realistic representation onto every fixture, so the 22 existing cases exercise the rename path now too. Reverting `check_authorship.py` to main while keeping the new test file fails 10 cases, not 2: nine of those are pre-existing rejection tests that were only meaningful because compact JSON fell below git's similarity threshold. That is the stronger argument for the change than the two cases you added on top.

Of those two, only the tampered one discriminates. "clean refutation still binds" passes against pre-fix code as well, since the pre-fix gate passes everything vacuously. Worth knowing which of the pair is the actual regression test if one ever starts failing.

The `R` branch is unreachable while `--no-renames` is passed, and it is the one part of the diff with no coverage. It also does not handle `C`, which would be the same shape of hole under `diff.renames = copies`. I checked that case: `--no-renames` on the command line beats `diff.renames = copies` in config, so both are theoretical. Keeping the fallback is fine, it just is not load-bearing.

I audited the other places in the repo that parse a diff for the same assumption, and the fix does not need to extend to any of them:

- `.github/workflows/verify.yml:67`, `git diff --quiet --diff-filter=AM`, is unaffected. `--quiet` short-circuits before rename detection runs, so it exits 1 on a pure rename. Confirmed against the live run on #608: the scope check, trusted `verify_all`, and the authorship step all executed, which matches your observation that authorship was passing rather than being skipped. Had this one been broken it would have been the more serious bug, since `codes_changed=false` also moves `verify_all` from the trusted tree to the submitted one.
- `check_prose.py:181` uses `AMR` and gets the new path.
- `check_submission_scope.py:37` uses `AMR`. Line 49 uses `A` and does miss renames, but its only consumer is the one-new-code cap, where a rename should not count as new. I tried to evade the cap by pairing a deleted entry against a genuinely new one, and two real board codes at the same (n, k) are not 50% similar, so git does not pair them.
- `cli/qldpc.py:477` is reporting, not a gate.

FarLab:
Thanks — the takeover reproduction is the right severity framing (authors + construction rewritten, gate exits 0), and the audit of the other diff parsers answers the question I should have asked next.

Two points taken, recorded here rather than force-pushed into the commit message to avoid dismissing the review:

- **`indent=1` is the load-bearing test change.** It retrofits realistic multi-line fixtures onto all 22 pre-existing cases, so the whole suite exercises the rename-folding path — your revert experiment (10 failures, not 2) is the stronger argument for it than the two named cases.
- **Of the two new cases, `high-similarity rename: tampered refutation still caught` is the discriminating regression test**; the "clean refutation binds" twin also passes against pre-fix code (which passed everything vacuously) and is there as the positive control.

Agreed the `R` branch is belt-and-braces, not load-bearing, and that `C` is theoretical while `--no-renames` is on the command line. Ready to merge from my side.

PR #622, opened 2026-08-19
verify: deep RIS kernel for ris_gpu — full-basis RREF + pair stage (--pair-depth)

FarLab:
## What

Adds a third kernel to `verify/ris_gpu.cu` behind a new `--pair-depth P` flag: per trial, the **full kernel basis** is copied into shared memory (with-replacement sketch draw only when an explicit smaller `--k-sub` is given), the usual collaborative RREF runs, and then — new — the XOR of every pair among the `P` lightest rows is scanned as well (Leon/Stern-style depth 2). Recover mode works unchanged: the winning row or row-pair XOR flows through the existing done-flag ladder to the CPU re-verifier.

The two sqetch-ported kernels are untouched (the "ported verbatim" attribution stays true).

## Why

The sketch kernels are depth-1 Prange: a trial succeeds only when the random column cut isolates the target's **entire** support at once — combinatorially unlucky as n and d grow. During the two-orbit GB campaign we found 1e5 CPU information sets (full kernel + pair search in `gf2_fast`) beating 4e7 sketched GPU trials at n ≥ 500. This kernel closes that gap: CPU per-trial power at GPU throughput.

Benchmark on the `[[682,20,22]]` GB candidate (X side, NVIDIA L4):

| mode | trials | wall | best found |
|---|---|---|---|
| sketch, `k_sub=64` | 40,000,000 | 39 s | 28 (misses) |
| deep, `--pair-depth 24` | 2,000,000 | 34 s | **22, CPU-verified witness** |

## Changes

- `verify/ris_gpu.cu`: `deep_pair_kernel` (+ shmem sizing, dispatch, usage); in deep mode `k_sub` defaults to the full kernel basis.
- `verify/ris_gpu.py`: `--pair-depth` passthrough; `--k-sub` default now defers to the binary; `parse_output` int-coerces `k_sub`/`k_null`/`k_logical`.
- `verify/test_ris_gpu.py`: four new tests (CPU-side flag surface; deep-mode exactness on Steane incl. the full-basis default; sketch+pairs hybrid on `[[186,10,14]]`; mid-size deep exactness). GPU-guarded tests now `pytest.skip()` with a reason instead of silently passing — CPU-only CI reports them as skipped; the `__main__` runner catches the skip exception so direct invocation on GPU boxes still works.

## Test plan

- CPU-only (CI situation): `pytest verify/test_ris_gpu.py` → 5 passed, 4 skipped.
- NVIDIA L4 (`make ris`, CUDA 12.9): 9 passed — deep mode recovers exact distances with CPU-verified witnesses on `[[7,1,3]]`, `[[186,10,14]]`; benchmark above.

### DIFF
--- a/verify/ris_gpu.cu
+++ b/verify/ris_gpu.cu
@@ -645,7 +645,8 @@ static void usage(const char* argv0) {
         "  is committed. --pair-depth P > 0 switches to the deep kernel:\n"
         "  full-basis RREF (k_sub defaults to the whole kernel) plus XOR\n"
         "  combinations of the P lightest rows per trial -- far stronger per\n"
-        "  trial at large n, at lower trial throughput.\n", argv0);
+        "  trial at large n, at lower trial throughput. P is capped at %d;\n"
+        "  larger values are rejected.\n", argv0, MAX_PAIR_TOP);
     exit(2);
 }
 
@@ -688,6 +689,11 @@ int main(int argc, char** argv) {
         }
     }
     if (!path || trials <= 0 || batch <= 0 || k_sub <= 0) usage(argv[0]);
+    if (pair_depth < 0 || pair_depth > MAX_PAIR_TOP) {
+        fprintf(stderr, "--pair-depth %d out of range (max %d)\n",
+                pair_depth, MAX_PAIR_TOP);
+        exit(2);
+    }
 
     Input in = read_input(path);
     int n = in.n, nw = in.nw;
@@ -806,8 +812,8 @@ int main(int argc, char** argv) {
     }
 
     printf("mode=%s\n", recover ? "recover" : "estimate");
-    printf("n=%d\nk_null=%d\nk_logical=%d\nk_sub=%d\n", n, in.k_null,
-           in.k_logical, k_sub_eff);
+    printf("n=%d\nk_null=%d\nk_logical=%d\nk_sub=%d\npair_depth=%d\n",
+           n, in.k_null, in.k_logical, k_sub_eff, pair_depth);
     printf("seed=%llu\ntrials=%lld\n", seed, trials_done);
     printf("best_weight=%d\n", best_overall <= n ? best_overall : -1);
     if (recover && !best_vec.empty()) {
--- a/verify/ris_gpu.cu
+++ b/verify/ris_gpu.cu
@@ -355,6 +355,224 @@ sqetch_ksub_recover_kernel(
 }
 
 /* ------------------------------------------------------------------ */
+/*  Deep kernel: full-basis RREF + depth-2 pair stage (this repo).     */
+/*                                                                     */
+/*  Motivation: the sketch kernels above are depth-1 Prange -- a trial */
+/*  succeeds only if the permutation isolates the target's ENTIRE      */
+/*  support. Empirically (two-orbit GB campaign, 2026-08-19) 1e5 CPU   */
+/*  information sets with a pair stage beat 4e7 sketched GPU trials at */
+/*  n >= 500. This kernel gives GPU trials the same per-trial power:   */
+/*  when k_sub >= k_null the whole kernel basis is copied (no          */
+/*  with-replacement rank loss), and after the RREF the lightest       */
+/*  pair_top rows are XORed pairwise (Leon/Stern-style depth 2).       */
+/*  Recover-capable: the winning vector (row or row-pair XOR) is       */
+/*  written back for CPU re-verification.                              */
+/* ------------------------------------------------------------------ */
+
+#define MAX_PAIR_TOP 32
+
+__global__ void __launch_bounds__(BLOCK_SIZE)
+deep_pair_kernel(
+    const uint64_t* __restrict__ W_null,
+    const uint64_t* __restrict__ W_logical,
+    int k, int kx, int nw, int n, int k_sub, int pair_top,
+    int* global_best,
+    unsigned long long base_seed,
+    int d_target,
+    int* found_flag,
+    uint64_t* out_vec,
+    int* done_flag
+) {
+    const int bid = blockIdx.x;
+    const int tid = threadIdx.x;
+
+    extern __shared__ char raw_shmem[];
+
+    uint16_t* perm = (uint16_t*)raw_shmem;
+    int perm_bytes = (n * 2 + 7) & ~7;
+    uint64_t* W_sub = (uint64_t*)(raw_shmem + perm_bytes);
+    int wsub_bytes = k_sub * nw * 8;
+    uint64_t* pivot_row_shmem = (uint64_t*)(raw_shmem + perm_bytes + wsub_bytes);
+    int pivot_bytes = nw * 8;
+    int16_t* wts = (int16_t*)(raw_shmem + perm_bytes + wsub_bytes + pivot_bytes);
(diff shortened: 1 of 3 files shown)


FarLab:
Deep-kernel vs CPU benchmark (identical algorithm both sides: full-kernel RREF + pair_depth 8; both find the exact known distances):

| code | CPU 8-thread (trials/s) | GPU L4 (trials/s) | speedup |
|---|---|---|---|
| [[126,12,10]] | 485k | 3.79M | 7.8× |
| [[254,14,16]] | 122k | 1.52M | 12.5× |
| [[682,20,22]] | 15.3k | 76.5k | 5.0× |

The n=682 dip is shared-memory occupancy (~30 KB/trial resident), as expected for the shmem-bound design. L4 is the budget GPU; an L40S roughly doubles these.

vprusso (approved):

vprusso (changes requested):
I could not compile this. There is no nvcc on my machine or anywhere in the fleet, and CI never builds it either: `make ris` appears in no workflow, only `make fast`. So 258 lines of CUDA would land with zero automated execution behind them. To review it anyway I transliterated `deep_pair_kernel` into a single-threaded CPU emulator, splitting phases at the `__syncthreads()` boundaries and running each phase for tid 0..127 in order, which is one legal execution of a correctly barrier-synchronised block. That exercises the algorithm, not the races and not the shared-memory sizing.

The algorithm is right. Under emulation it recovers exact distances with CPU-verifiable supports: weight 3 on Steane, weight 14 on `[[186,10,14]]` at 2000 full-basis trials. The barrier structure matches the existing sketch kernel's idiom, two syncs on every path through the column loop including the `pr >= rows` early-out, the triangular pair-index decode is correct, and the inlined parity fold in the pair stage agrees with `gf2_dot`.

What I would like changed before merge:

1. The benchmark does not isolate the pair stage. The deep kernel changes two things at once, the full kernel basis replacing a 64-row with-replacement sketch and depth-2 replacing depth-1, and the table compares the pair only against the sketch kernel. The missing row is `--pair-depth 1`, which keeps the full-basis RREF and makes `npairs` zero. On my emulation, 200 trials on `codes/666-150-76.json`, seed 1:

   | config | best weight |
   |---|---|
   | `k_sub=64`, pairs off (sketch-like) | 98 |
   | full basis, `--pair-depth 1` | 99 |
   | full basis, `--pair-depth 24` | 97 |

   The pair stage is doing something real, but at matched trial counts the full basis alone is not better than the sketch on this code. Please run the `[[682,20,22]]` benchmark once more at `--pair-depth 1` and the same 2M trials, and put the number in the table. If it also lands at or near 22, the win is the full basis and the pair bookkeeping is not earning its hundred lines.

2. None of the four new tests exercises the pair stage. I re-ran all of them under emulation with the stage inert (`--pair-depth 1`) and every assertion still holds: Steane still reaches 3, `[[186,10,14]]` still reaches 14, and the hybrid test only asserts `k_sub == 8` plus CPU-verify consistency. The cause is regime. In my sweep, pairs change nothing at n <= 360 (186, 248 and 360 all saturate at the exact distance with the stage off) and only start mattering at n >= 500, where `666-150-76` goes 99 to 97 and `514-162-41` goes 60 to 57 at 200 trials. So the tests cover the flag surface and the full-basis default, which is worth having, but the mechanism the PR is named after has no test that would fail if you deleted it. Either add one at n >= 500 with a fixed seed and a trial budget between the two regimes, or say in the docstrings that these cover deep-mode plumbing rather than the pair stage.

3. `--pair-depth` above 32 is silently clamped to `MAX_PAIR_TOP`. `--pair-depth 64` and `--pair-depth 32` produce identical output, and neither the usage text nor the argparse help mentions a cap. Someone running a long campaign at `--pair-depth 48` gets 32 and no diagnostic. Print the effective value or reject the argument, and put the cap in the usage string.

The trust posture is unaffected, which I checked rather than assumed. `ris_gpu.cu` is not in the trusted closure in `check_validator_integrity.py`, and it should not be: a recovered vector goes through `cpu_verify` and then the real verifier, so a wrong kernel loses results rather than admitting bad ones. The `parse_output` int-coercion of `k_sub`, `k_null` and `k_logical` has no consumers outside the test file, so it cannot change an existing witness JSON.

Separately from this PR, nothing in CI compiles this file at all. A CPU-only syntax gate would at least catch a build break on a machine none of us has in front of us. I will open that as its own issue.

FarLab:
@vprusso Its true that the current CI cannot test the new GPU code, the GPU tests should be skipped. Once we add GPU in the CI, then this could be tested. Did you try on an actual GPU if you have access?

FarLab:
Ablation as requested (all at 2M trials, seed 3, `[[682,20,22]]` X side, L4):

| config | best weight |
|---|---|
| sketch `k_sub=64`, no pairs | 28 |
| full basis, `--pair-depth 1` | **22** |
| full basis, `--pair-depth 24` | 22 |

You called it: **on this benchmark the full basis is the win**, and the pair stage doesn't move the headline number. The pair stage's contribution shows at matched small budgets on larger/denser kernels — on `[[514,162,41]]` (X side, seed 11, 20k trials): pair-depth 1 → 56, pair-depth 24 → **54**. That comparison is now `test_gpu_pair_stage_matters` (estimate mode is a pure min over a seed-determined trial set, so the strict inequality is deterministic and the test fails if the pair stage is deleted).

Also addressed:
- `--pair-depth > 32` is now rejected with a diagnostic (cap documented in usage; effective value printed in the output).
- Deep-mode test docstrings state explicitly that they cover plumbing/full-basis behavior, not the pair stage.

Suite: 5 passed / 5 skipped CPU-only; 10 passed on an L4. Happy to retitle/reframe the PR toward "full-basis deep mode (+ pair stage for large n)" if you prefer the honest emphasis.

vprusso (approved):
I reproduced the ablation on the CPU emulator and it matches your L4 numbers exactly: `514-162-41` X side, seed 11, 20000 trials, full basis, gives 56 at `--pair-depth 1` and 54 at `--pair-depth 24`. Same integers, from a single-threaded transliteration with a completely different execution model, which is the strongest confirmation I can get without hardware.

That also settles the determinism argument, and it is a better argument than I gave you credit for. `batch` is a host-side constant of 50000, so a 20000-trial run is a single batch, `batch_seed` never advances, and the block seeds are `seed ^ (bid*A + C)` for `bid` in `[0, 20000)`. Nothing in the trial set depends on the device: no grid sizing from SM count, no floating point anywhere in the kernel, and `atomicMin` is order-independent. So `test_gpu_pair_stage_matters` is exact across GPUs, not just repeatable on yours. One fragility to be aware of rather than fix: it inherits the default `--batch`, so if that 50000 ever changes and the test's trial count crosses it, the trial set changes with it.

Yes to the retitle. The ablation says the full basis is the headline result and the pair stage is a second-order effect that appears at large n, so "full-basis deep mode (+ pair stage for large n)" describes what this actually does. Worth carrying the ablation table into the file comment above `deep_pair_kernel` too, since the motivation paragraph there currently attributes the win to the pair stage.

`pair_depth` is now printed but is not in the int-coercion tuple in `parse_output`, so it comes back as a string while its neighbours come back as ints. Nobody consumes it yet, so it can ride along in your next touch of the file rather than another round trip.

Approving. The cap now errors instead of silently clamping, the docstrings say what they cover, and the pair stage has a test that fails if you delete it.

PR #632, opened 2026-08-20
Add [[177,9,5]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[177,9,5]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/177-9-5.json` passes locally

Pareto frontier in all 12 track cells — `local-2d-single`, `local-2d-bilayer`
and `unrestricted`, each at weight <= 4, <= 6, <= 8 and any weight. Score
`kd^2/n = 1.271`, max check weight 4, locality class `local-2d-single`.

Two things worth stating plainly, since a 12-of-12 listing overstates itself if
read as a record claim:

- This is a **frontier position, not a density record.** The cell it targets is
  thin and its moderate-`k` interior is empty — nothing on the board reaches
  `k >= 9` at `d >= 5`. The board's best `kd^2/n` is 1.564, well above this
  entry's 1.271. The family's own ceiling is `4d^2/(3d^2+1) -> 4/3`.
- The distance is a **witness-backed upper bound**, from 20,000 RIS trials per
  side, and is submitted as `upper_bound` on both sides. No certificate
  accompanies it and no exact claim is made.

This is the `m = 5` rung of a one-parameter family whose `m = 3` case is the
board's existing dense-packing column: `[[101,5,5]]`, `[[197,5,7]]`,
`[[325,5,9]]`, `[[485,5,11]]`, `[[677,5,13]]`. The parameterization reproduces
that whole column exactly, which is the main evidence that this generalizes the
published construction rather than resembling it. This entry is not equivalent
to any of them — `n` and `k` both differ. Recorded in `provenance.notes`.

Research note: `notes/177-9-5.md` — including the band-pitch threshold that
makes the multi-band variants collapse to a flat distance of 6 below
`pitch_min(d)`, which is the part of the search most likely to save someone else
time.


FarLab:
Hi @Xo1otl, thank you for the code submission! If you could star the repo, that would be much appreciated :)

Xo1otl:
Thanks for merging! Just starred the repo.

PR #634, opened 2026-08-20
Add [[202,10,5]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[202,10,5]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/202-10-5.json` passes locally

Pareto frontier in all 12 track cells. Score `kd^2/n = 1.238`, max check
weight 4, locality class `local-2d-single`.

claim:

  `kd^2/n` is 1.564, above this entry's 1.238.
  and no exact claim is made.

### The interesting part: a band-pitch threshold

This entry is a *multi-band* packing — 4 bands rather than the published 2 — and
it only works above a pitch threshold. That threshold is the finding, and the
research note is mostly about it:

- At the **published** vertical pitch `d - 1`, adding bands adds qubits and **no
  logicals at all**. `k` does not move. This is the natural-looking
  generalization, and it returns nothing.
- Above a threshold, `k` becomes the full patch count
  `rows * m - floor(rows / 2)`, with distance preserved. Measured thresholds:
  `pitch_min = 6` at `d = 5`, `10` at `d = 7`, `12` at `d = 9`.
- **Below** the threshold the distance collapses to a flat **6, regardless of
  `d` and `n`** — a floor that is easy to mistake for a working code if only
  `d = 5` is examined.
- At `pitch >= 2d` the bands stop sharing checks and the distance collapses to
  1. Odd pitch breaks CSS outright.

Reviewer-relevant caveat, stated rather than buried: **this code sits exactly at
its threshold.** `pitch = 6` is `pitch_min(5) = 6`, the boundary and not the
interior, so it is the least-margin point of the multi-band family. The witness
confirms `d <= 5`, but anyone building on this should take a larger pitch for
margin. Also, `pitch = 6 = d + 1` here is a coincidence of `d = 5`: the working
pitches at `d = 7` and `d = 9` are 10 and 12, which are `d + 3`.

The two-band, pitch `d - 1` case of this construction is the board's existing
dense-packing column `[[101,5,5]]` .. `[[677,5,13]]` (arXiv:2511.06758; Fujiu,
Nagayama, Nishio, Kawaguchi, Satoh), which the parameterization reproduces
exactly. The band stagger into a second dimension is not in that paper. Not
equivalent to any board entry. Mutually non-dominated with the two-band rungs
`[[177,9,5]]` and `[[215,11,5]]`, submitted separately. Recorded in
`provenance.notes`.

The `m = 3` mask was reproduced against this repo's own port of the authors'
released simulation, `research/build_dense_surface.py`.

Research note: `notes/202-10-5.md`.


vprusso (approved):

FarLab:
@Xo1otl Looks like there is an issue with code we just merged in to main. Can you try merging with main?

Xo1otl:
Done — merged main into the branch. Thanks!

PR #635, opened 2026-08-20
Add [[271,7,7]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[271,7,7]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/271-7-7.json` passes locally

Pareto frontier in all 12 track cells. Score `kd^2/n = 1.266`, max check
weight 4, locality class `local-2d-single`.

claim:

- This is a **frontier position, not a density record.** The targeted cell is
  thin — nothing on the board reaches `k >= 6` at `d >= 5` below `n = 368`. The
  board's best `kd^2/n` is 1.564, above this entry's 1.266.
  and no exact claim is made. Worth noting that RIS screening gets harder as `d`
  grows, so a `d = 7` bound is weaker evidence per trial than a `d = 5` bound at
  the same trial count.

This is the `m = 4` rung of a one-parameter family whose `m = 3` case is the
board's existing dense-packing column: `[[101,5,5]]`, `[[197,5,7]]`,
`[[325,5,9]]`, `[[485,5,11]]`, `[[677,5,13]]`. The `d = 7` member of that column,
`[[197,5,7]]`, is this code's direct `m = 3` neighbour. The parameterization
reproduces that whole column exactly, which is the main evidence it generalizes
the published construction rather than resembling it. Not equivalent to any of
them — `n` and `k` both differ. Mutually non-dominated with the `d = 5` rungs
`[[177,9,5]]` and `[[215,11,5]]`, submitted separately. Recorded in
`provenance.notes`.

Research note: `notes/271-7-7.md` — including the band-pitch threshold, which at
`d = 7` sits at pitch 10 rather than the `d + 1` the `d = 5` case might suggest.


vprusso (approved):

codes/271-7-7.json
```diff
@@ -155,1541 +169,2466 @@
...
+    56,
+    61,
+    62
+   ]
+  }
+ },
+ "provenance": {
+  "authors": [
+   "@Xo1otl",
+   "QunaSys"
```
FarLab: is QunaSys a github user handle?

FarLab:
@Xo1otl can you also add the number of layers in the json file? Similarly for the other open PRs. Thanks!

codes/271-7-7.json
Xo1otl: Ah, my mistake — it's not a GitHub handle. Removing it from all three open PRs, sorry!

Xo1otl:
Sure — adding `"layers": 1` here and to the other open PRs. Thanks!

FarLab:
Thanks @Xo1otl. Looks like there is an issue with code we just merged in to main. Can you try merging with main?

Xo1otl:
Done — merged main into the branch (and #645 as well). Thanks!

PR #641, opened 2026-08-20
Fix layers display for [[177,9,5]] and require layers with layouts

FarLab:
`codes/177-9-5.json` omitted the optional `locality.layers` field, so the code page showed `?` for layers while the verifier and `g` score defaulted to 1.

- Add `"layers": 1` to the entry (verifier still passes, same fingerprint).
- `site/build.py`: default the layers tile to 1 instead of `?`, and the stacked-legend default to 1 instead of 2, matching `qldpc_verify.py`.
- Schema: `locality.layers` is now required alongside `coordinates` (also in the verifier's no-jsonschema fallback). `qldpc submit --layers` defaults to 1 and is always written. Docs (CONTRIBUTING, TRACKS) updated; new test `layout without layers rejected`.

All 339 existing codes validate against the tightened schema; `run_tests.py` passes.


vprusso (changes requested):
The data fix and the params-tile default are right, and I checked both rather than assuming. `codes/177-9-5.json` has `max_qubits_per_site = 1`, so `layers: 1` is the true value and not just a convenient one, and the verifier already reported `computed.locality.layers = 1` before the change, so nothing scored moves. It also closes the set: after this, all 78 layouts on the board declare `layers` explicitly, and the `detail_page` default becomes unreachable for current data while still being the correct fallback going forward.

The legend default is wrong, though, and it is the one line I would not merge.

```python
if stacked:
    legend.append('<span><span class=rg></span>2 qubits stacked '
                  f'({loc.get("layers", 1)} layers)</span>')
```

`stacked` is set exactly when some site holds more than one qubit, so that branch only runs on a layout that is drawn with a stack. Meanwhile `qldpc_verify.py` computes `honest = max_mult <= layers`, so a stacked layout with `layers = 1` is by definition not honest. The default can therefore only fire on a code that is stacked and undeclared, and for that code it renders `2 qubits stacked (1 layers)`, which contradicts both the drawing above it and the verifier's own rule.

I confirmed the reachable case by deleting `layers` from a stacked entry and running the verifier: it stays `ok: true` with `layers: 1, max_qubits_per_site: 2`, fails the honesty test, and drops to `unrestricted`. The site still has a `locality` block for it, so it still draws the layout and still hits this branch.

The reasoning that produced it is worth naming, because it will come up again. The verifier's default of 1 is conservative for classification: assume the fewest layers so that cramming qubits onto a point cannot buy a locality class. Copying that number into a description of an already-drawn stack inverts what the conservatism was for. Same constant, opposite meaning.

Either revert that hunk, or derive it from what is actually being drawn:

```python
f'({loc.get("layers", max(mult.values()))} layers)'
```

`mult` is already in scope a few lines up. That also fixes `1 layers` reading ungrammatically, since the value can no longer be 1 inside this branch.

Nothing on the board hits this today (0 stacked layouts with `layers` absent, and this PR removes the last absent one), so it is latent rather than live. Still worth getting right while the line is open.

One thing I would leave alone: showing `1` in the params tile where the field is absent does present a computed default as though it were declared. `?` was arguably more honest about that. But the verifier and the `g` score both treat absent as 1, so displaying 1 matches what is actually scored, and the inconsistency you are fixing is the worse of the two.

FarLab:
Agreed on the legend default — it copied the classification-conservative 1 into a description of a drawn stack. Switched to `loc.get("layers", max(mult.values()))` as suggested (94d1bcb); checked on 72-6-6 with `layers` deleted that the legend renders `(2 layers)`. Params-tile default left at 1 per your last paragraph.

FarLab:
CI is expected to fail here. Will merge.

Xo1otl:
Heads up — this change edited `verify/qldpc_verify.py` but `verify/validator_manifest.json` still pins the old hash (pinned `00490f5f…`, now `1281f4ae…`), so `check_validator_integrity.py` now fails CI on open submission PRs (e.g. #634, #635). Re-pinning with `python verify/check_validator_integrity.py --update` should fix it.

PR #642, opened 2026-08-20
Add two-layer 2D-local layout for [[84,6,10]] (r = 5.0, weight-6 bilayer record)

willzeng:
## Layout submission

- Parameters: [[n, k, d]] = [[84, 6, 10]] (existing literature baseline, code unchanged)
- Tracks: local-2d-bilayer / weight-6 (newly earned by this layout)
- Contribution: the verified LAYOUT, following the `codes/16-6-4.json` precedent — the code stays a Liang–Liu–Song–Chen baseline (arXiv:2503.03827); this PR adds `locality` only, plus provenance notes crediting the layout.

Weight-6 bilayer cell. The layout has measured interaction radius **exactly 5.0** with 2 layers, min site spacing 1.0, and at most 2 qubits per site — verifier-computed class `local-2d-bilayer`. That places kd²/n = 7.14 at the top of the weight-6 bilayer cell (current best: `codes/72-12-6.json` at 6.00; no strict domination — [[72,12,6]] keeps higher k at lower n — so both remain on the cell frontier). Geometric efficiency g = 0.0114.

### Method

Two independent approaches, the generic one winning: (1) a structured product-fold layout derived from the code's Z₄₂ translation symmetry (recovered from the Tanner graph by automorphism search) bottoms out at r = √52 ≈ 7.21 — provably optimal within linear deformations once the unit-spacing constraint is priced, and any 1D arrangement of the 7-cycle axis forces an end-to-end pair since its check offsets reach ±3; (2) simulated annealing over site assignments (spectral initialization, Metropolis with softmax cost, max-focused hill-climb polish) breaks the product structure and reaches r = 5.0. Twenty-two runs from varied initializations all converged to 5.0 or 5.099, and continuous off-grid refinement could not improve on the grid optimum, so 5.0 is likely at or near this code's floor.

`python verify/qldpc_verify.py codes/84-6-10.json` passes locally, including the distance refutation pass (the checks and distance claims are untouched). Layout search used Claude Fable 5; `provenance.model` is set accordingly.


willzeng:
Reworked to the #643 layout-binding form (review round 2 design): `provenance.authors` and `model` are untouched — layout credit and method now live in `locality.contributed_by` (`@willzeng`, 2026-08-20, annealing method string), with `schema_version: 0.2` per the schema gate. Branch rebased on main, squashed to a single codes-only commit.

Verified locally against #643's schema and verifier (full pass, `local-2d-bilayer`, r = 5.0), and #643's `check_authorship.py` accepts the diff. **Depends on #643**: CI here will fail `schema_valid` until the `contributed_by` schema field merges — re-run checks after #643 lands.

vprusso (approved):
Verified rather than read. The +1084/-735 shape looked alarming for something described as adding one block, but structurally the diff is clean: `checks`, `distance`, `n`, `k`, `family` and `code_type` are byte-identical to the base and the deletions are reformatting. `locality` is added, `schema_version` goes 0.1 to 0.2 as `contributed_by` requires, `name` gains its suffix, and `provenance.notes` is a strict append. `authors` and `model` are untouched.

The verifier agrees with every number in the description:

```
ok: True
locality: {"interaction_radius": 5.0, "layers": 2, "max_qubits_per_site": 2,
           "min_site_spacing": 1.0, "qubits_per_unit_area": 2.0, "bbox": [6.0, 7.0]}
failing checks: none
```

Radius exactly 5.0 against the 7.0 bilayer cap, 2 layers, no cramming, unit spacing, so `local-2d-bilayer` is earned rather than asserted. `g = 4·6·100/(84·2²·5⁴) = 0.01143` matches the 0.0114 you quote.

The frontier claim holds and the caveat you attached to it is the right one. Recomputing the weight-6 bilayer cell from the layouts that actually pass the honesty test:

| code | params | kd²/n |
|---|---|---|
| 84-6-10 | [[84,6,10]] | 7.14 |
| 72-12-6 | [[72,12,6]] | 6.00 |
| 62-10-6 | [[62,10,6]] | 5.81 |
| 70-6-8 | [[70,6,8]] | 5.49 |

Top of the cell on the headline figure, and no strict domination over [[72,12,6]], which keeps higher k at lower n. Both stay on the frontier, exactly as you wrote it.

## One thing worth knowing, which is not a defect in this PR

The #643 layout binding never runs on this submission. `84-6-10.json` credits its authors as plain names (`Zijian Liang`, `Ke Liu`, `Hao Song`, `Yu-An Chen`), so `handles()` returns empty and `main()` takes the earlier exit:

```
ok    codes/84-6-10.json: no @handle authors (baseline/anonymous), exempt
```

I checked both directions. The original form of this PR, appending `@willzeng` to `provenance.authors`, is rejected on current main. The submitted form passes, but through the baseline exemption rather than through the binding built for it. The same diff from `@mallory` also passes, which I ran to be sure.

So #643 earned its keep by making the first approach impossible, not by admitting the second. The binding will do real work on the 185 layout-less entries that do have `@handle` owners; this one was never gated.

The consequence is worth carrying into the layout-filling effort: your layout lands with no authorship protection. Anyone can later replace or delete it and the gate will not object, because the entry still has no `@handle` authors after this merges and `contributed_by` confers no edit rights by design. That is the correct design, but it means layout work on plain-name literature baselines is unprotected in a way layout work on handle-owned entries is not. Might be worth a follow-up on whether `contributed_by` should also bind the exempt path.

Minor: the description says "Layout search used Claude Fable 5; `provenance.model` is set accordingly." `model` is unset, correctly, since the binding forbids setting it. The attribution now lives in `contributed_by.method`, which reads "Search driven by Claude Fable 5." The data is right and the sentence is stale.

Approving.

PR #643, opened 2026-08-20
check_authorship: layout binding — first-layout additions may append their contributor

willzeng:
## Policy extension: the layout binding

PR #642 (adding a verified 2-layer layout to the `codes/84-6-10.json` literature baseline) failed the authorship gate: a modification to an existing entry currently has only two doors — be a listed `@handle` author, or be exactly a refutation credited in `witness_provenance.found_by`. Retrofitting a layout onto a baseline fits neither, and first-claiming a handle on a no-handle baseline was reserved to maintainers. With the layout-filling effort ahead (most board entries lack a layout and therefore a g score), that path needs to be CI-passable.

This adds `layout_binding()` alongside `refutation_binding()` in `verify/check_authorship.py`. A modified entry passes when the diff is **exactly** a first-layout addition:

- adds `locality` where none existed (replacing an existing layout stays reserved to the entry's listed authors);
- `provenance.authors` keeps the original list and **appends** the contributor (the PR author must be among the appended handles);
- `provenance.notes` append-only; `provenance.model` may be set only if previously absent;
- `name` may change (e.g. a layout suffix);
- **everything else — checks, distance, and all other provenance — must be byte-identical.**

The layout itself is a separately verifiable artifact (the verifier recomputes radius, spacing, per-site occupancy, and locality class from the coordinates), so crediting its contributor grants no edit rights over the code: any later edit by them must still pass one of the bindings. The issue #611 concerns (author swaps, credit grabs, ride-along edits) are each covered by a test.

Tests: 10 new cases in `verify/test_check_authorship.py` (baseline and handle-owned targets, missing self-credit, ride-along check/distance/notes edits, author swaps, layout replacement, model overwrite); all 20 existing cases pass unchanged; `run_tests.py --skip-slow` green. `check_authorship.py` is not in `verify/validator_manifest.json`, so no manifest update is required.

Once merged, #642 can be re-run against the updated base and should pass as-is.

### DIFF
--- a/schema/SCHEMA.md
+++ b/schema/SCHEMA.md
@@ -64,6 +64,11 @@ Two principles drive the format:
   - `interaction_radius`: claimed max check diameter in the layout; the
     verifier recomputes the true max check diameter and requires
     `measured <= claim`.
+  - `contributed_by` (optional, schema 0.2): `{by, date, method?}` crediting
+    who contributed this layout. Layout credit lives here, beside the
+    artifact, not in `provenance.authors` — the same separation
+    `witness_provenance` uses for refutation credit — so adding a layout to
+    an existing entry never changes its author list.
 - `provenance`: `authors`, `construction` (how it was built), optional
   `references`, `date`, `notes`, `model`.
   - `origin`: `"baseline"` for a literature seed or `"submission"` for a code
--- a/schema/code.schema.json
+++ b/schema/code.schema.json
@@ -113,6 +113,35 @@
           "type": "number",
           "minimum": 0,
           "description": "claimed max check diameter in the layout; verifier recomputes"
+        },
+        "contributed_by": {
+          "type": "object",
+          "description": "who contributed this layout, and when. Layout credit lives here, on the artifact contributed, not in provenance.authors -- the same separation witness_provenance uses for refutation credit, so contributing a layout never widens edit rights over the code (issue #611).",
+          "required": [
+            "by",
+            "date"
+          ],
+          "additionalProperties": false,
+          "properties": {
+            "by": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "pattern": "^@[A-Za-z0-9-]+$"
+              },
+              "minItems": 1,
+              "description": "@handles of the people who contributed this layout"
+            },
+            "date": {
+              "type": "string",
+              "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
+            },
+            "method": {
+              "type": "string",
+              "maxLength": 400,
+              "description": "optionally, how the layout was found (tool, model, search method)"
+            }
+          }
         }
       }
     },
@@ -241,6 +270,28 @@
         },
         "required": [
           "distance"
+        ]
+      },
+      "then": {
+        "properties": {
+          "schema_version": {
+            "const": "0.2"
+          }
+        }
+      }
+    },
+    {
+      "$comment": "locality.contributed_by (a 0.2 feature) likewise requires schema_version 0.2",
+      "if": {
+        "properties": {
+          "locality": {
+            "required": [
+              "contributed_by"
+            ]
+          }
+        },
+        "required": [
+          "locality"
         ]
       },
       "then": {
--- a/verify/check_authorship.py
+++ b/verify/check_authorship.py
@@ -23,14 +23,14 @@ refutation binding exists to avoid. A refuted 'exact' claim must demote to
 upper_bound, and no correction may claim 'exact' at the new value without going
 through certification. New submissions are unaffected.
 
-One more binding exists for layouts: a change that ONLY adds a first `locality`
-block (the code had none) may also append the contributor to `authors` and set
-a previously-absent `model`, with everything else -- checks, distance, and the
-original provenance -- byte-identical (notes append-only). The layout is a
-separate verifiable artifact (the verifier recomputes the radius and locality
-class from the coordinates), so crediting its contributor does not grant edit
-rights over the code itself: any later change by them still has to pass one of
-these bindings or be a pure layout replacement they authored.
+One more binding exists for layouts, and it mirrors the refutation binding's
+credit model exactly: a change that ONLY adds a first `locality` block (the
+code had none) binds when the block's `contributed_by.by` lists the PR author.
+`provenance` must be byte-identical apart from an append to notes -- in
+particular `authors` and `model` never change, so layout credit lives beside
+the artifact (like witness_provenance.found_by) and can never widen the
+contributor's edit rights over the code: they are still not a listed author
+on the next PR, and any later change must itself pass a binding.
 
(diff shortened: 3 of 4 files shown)


vprusso (changes requested):
The binding itself is written carefully and the field-by-field restrictions do what they say. The problem is one step later: appending the contributor to `provenance.authors` hands them the first door, and the first door has no restrictions at all.

`main()` short-circuits before any binding runs:

```python
if author in hs:
    base_hs = handles(base_doc) if base_doc is not None else None
    if (base_hs and hs != base_hs and author not in base_hs):
        pass                      # fall through to the bindings
    else:
        print(f"ok    {f}: PR author @{author} is listed")
        continue
```

After a layout binding merges, the contributor is in the base author list, so on the next PR `hs == base_hs` and the guard never engages. Two PRs against your branch, no maintainer in the loop:

```
PR1  @eve bolts a junk layout onto @npdeep's code -> exit 0  @eve binds via layout addition
PR2  @eve rewrites checks/distance/construction   -> exit 0  PR author @eve is listed
```

PR2 replaced every X check, moved `d` from 4 to 9, and rewrote `provenance.construction` to "eve's construction". It passes on the same code both against a no-handle literature baseline and against an entry owned by `@npdeep`.

So the docstring's claim inverts:

> crediting its contributor does not grant edit rights over the code itself: any later change by them still has to pass one of these bindings

It does grant them, on the very next PR, and that is the exact hole #611 exists to close ("adding your own handle would grant edit rights over anyone's entry").

The first step is cheap and needs no real work. A layout does not have to be any good to land: the verifier assigns `unrestricted` to a bad one rather than rejecting it. I bolted a collinear layout onto `codes/100-20-8.json` (owned by `@yifanhong`) and it verifies clean:

```
verifier ok: True
computed.locality: {'interaction_radius': 99.0, 'layers': 1, 'max_qubits_per_site': 1, ...}
failing checks: none
```

261 of the 339 board entries have no layout, 185 of them owned by an `@handle`. That is the attack surface, and it is the same set your motivation section is about.

## The fix, and it is already in this file

`refutation_binding` had this identical problem and solved it the other way: it refuses any change to `provenance.authors` at all ("authors and construction are the constructor's") and credits the refuter through `witness_provenance.found_by` instead. Credit lives beside the artifact, not in the author list, so the refuter never widens their own access.

Do the same here. Put the layout contributor in something like `locality.contributed_by`, leave `provenance.authors` byte-identical, and the escalation disappears without any change to `main()`. It also makes the two bindings symmetric, which is worth something on its own for a file this security-sensitive.

If credit really has to appear in `authors`, then the first door needs to distinguish a layout-only author from a constructor before it waves anyone through, which is more machinery for a weaker guarantee.

## On the tests

The 10 new cases are thorough about what PR1 may contain, and I could not get any of the single-PR abuses past them. They all share one blind spot: every case verifies a diff against a fixed base and stops. Nothing exercises the state the repo is left in once a binding has merged.

Worth adding as a permanent regression, since this is the second time the author list has turned out to be the privilege boundary: apply a binding, commit it to `main`, then re-run the gate on a follow-up edit by the same author and assert it is refused.

Two small things while you are in here. `appended = [a for a in na[len(ba):]]` is just `na[len(ba):]`. And the failure message concatenates both binding reasons, which for a plain unauthorized edit prints two rejections where one is the relevant one; worth leading with whichever binding got furthest.

willzeng:
Updated per review — the escalation is closed by adopting exactly the symmetric design you suggested:

- **Credit moved to `locality.contributed_by`** (`{by, date, method?}`, schema-0.2-gated like `witness_provenance`): the binding now requires `provenance` to be byte-identical apart from an appended note — `authors` and `model` never change, so a layout contributor is not a listed author on their next PR and the first door never opens for them. `main()` is untouched.
- **Merged-state regression added**: the new test commits a layout binding to main, then asserts the same contributor's follow-up rewrite (checks + distance + construction) is refused, and that a clean refutation by them still binds. This encodes 'the author list is the privilege boundary' as a permanent check, per your suggestion.
- **Failure message** now leads with the binding the diff was evidently aiming for (layout-shaped diffs report the layout reason first), with the other in parentheses; the redundant comprehension is gone with the rewrite.
- Schema + SCHEMA.md document the new field; `schema/code.schema.json` change is additive and version-gated (validated: 0.1 + `contributed_by` rejected, 0.2 accepted, plain layouts unaffected).

Branch is rebased on main (post-#644 re-pin), integrity check green, all 24 authorship cases + `run_tests.py --skip-slow` pass. Note for after merge: #642 will be reworked to the `contributed_by` form before re-running its CI.

vprusso (approved):
The escalation is closed. I re-ran the same two-step attack against this head, adapted to the `contributed_by` form, and PR2 is now refused:

```
PR1  @eve adds layout via contributed_by   -> exit 0  binds via layout addition
PR2  @eve rewrites the code                -> exit 1  refused
```

The two adjacent paths are shut as well: appending to `provenance.authors` no longer binds at all, and a `contributed_by` block crediting someone other than the PR author is rejected. Omitting credit entirely is rejected. A base entry with no `provenance` key still works. Nothing outside `check_authorship.py` reads `contributed_by`, so it confers no authority anywhere else, which is the property that makes this safe.

The merged-state pair is the right test and it is asserting the right thing in both directions: that a merged binding does not let the contributor edit the code, and that it also does not *cost* them the refutation path they would otherwise have had. The second half is easy to leave out and its absence would have been a slow regression.

The schema change is tight. `contributed_by` is `additionalProperties: false`, `by` is pattern-matched to `^@[A-Za-z0-9-]+$` so a free-text name cannot masquerade as a handle, and the 0.2 gate is a separate `if/then` that leaves plain layouts alone.

## One thing to fix before merging

The guidance message still describes the design you replaced:

```python
print("\nAuthorship mismatch: the PR author must be one of a code's "
      "@handle authors, or the change must be exactly a refutation "
      "credited to them in witness_provenance.found_by (issue #611), "
      "or exactly a first-layout addition that appends them to "
      "provenance.authors.")
```

That last clause now names the exact action the gate rejects. It is the only instruction a blocked contributor gets, so following it verbatim gets them blocked a second time with no hint why. Should read that the layout addition must credit them in `locality.contributed_by.by`.

## Follow-up, not for this PR

Layout credit is currently invisible to anyone reading the board. `site/build.py` has no reference to `contributed_by`, so a contributor who retrofits a layout onto someone else's entry gets a JSON field and nothing on the page. The mechanism exists to give credit, and right now it records it without showing it. Worth an issue against the layout-filling effort, since that effort is the reason this binding exists and the credit is most of the incentive.

Approving on the fix above; the security property is what I was blocking on and it is correct now.

willzeng:
Guidance message fixed — the mismatch help text now points at `locality.contributed_by.by` instead of the retired authors-append instruction. Follow-up issue opened for surfacing layout credit on the site.

vprusso (approved):

Issue #505, opened 2026-08-14
RFC 0001: A certified circuit tier for the board (d_circ)


FarLab:
# RFC 0001: A certified circuit tier for the board

**Status:** draft · **Author:** @FarLab (with Claude Fable 5) · **Date:** 2026-08-14

**Revision 2 (2026-08-14):** certification method changed. Rev 1 used stim's
`shortest_graphlike_error`; that is inapplicable to most of the board (see
[Certification method](#certification-method-why-not-graphlike)). d_circ is now
a witness-backed upper bound obtained by the board's own search on the DEM's
parity-check matrix, with the same `≤` semantics `d` already carries.

## Summary

Add a *circuit tier* to the board: submissions may include syndrome-extraction
circuits, and the verifier establishes their **circuit-level distance d_circ**
per basis. No new metric is introduced — the existing scores are additionally
evaluated with `d → d_circ`. The tier is penalty-only (it can never inflate a
score), its claims are witness-backed and checkable in milliseconds by anyone,
and it turns "find a better syndrome schedule for code X" into a refutable
research task across the entire existing board. Measured logical-error-rate
scoring (Monte Carlo) is explicitly out of scope for this RFC and sketched only
as future work.

## Motivation

Code distance is not what a machine experiences. Between `d` and real
performance sit the syndrome-extraction circuit's fault mechanisms: an
ancilla fault mid-extraction propagates onto multiple data qubits (a hook
error), and measurement outcomes can themselves be faulty. Whether these
mechanisms cost distance depends entirely on the extraction schedule — the
per-check CX orderings — and a poorly scheduled circuit can realize as
little as half the code distance, while a well-scheduled one realizes all
of it. The schedule is therefore part of the design, and today the board
cannot see it: two submissions with identical [[n,k,d]] can differ by a
factor ~2 in the distance a device would actually enjoy.

Circuit-level distance is also the natural next rung on the board's
existing ladder of certified truth (exact k by rank, witness-backed d,
exact d by certificate): it is a combinatorial property of a concrete
artifact — established by the same witness-and-refutation discipline the
board already applies to `d`, not by simulation. The wider field is moving
toward circuit-level evaluation — state-of-the-art dense-code proposals
optimize measured error rates directly — but no public leaderboard
certifies circuit distance. This tier occupies that ground at a fraction
of the cost of full error-rate benchmarking.

## Design principles

1. **A parameter, not a metric.** d_circ joins n, k, d as a certified
[...]

PR #646, opened 2026-08-20
Circuit tier Phase A (RFC 0001, #505): schema, verifier fast path, toolkit

FarLab:
Implements the trusted-stack half of RFC 0001's Phase A (#505), rev 2 semantics: d_circ as a witness-backed upper bound established on the DEM's parity-check matrix by the board's own search-and-refute discipline. The seed submission ([[25,1,5]] circuits + claim) follows in a stacked PR so this one stays on the trusted side of the submission-scope boundary.

## What's in

- **Schema**: optional `circuit` block (schema 0.2) — per-basis `d_circ` `{value, confidence, witness}`, `rounds`, `stim_version`, `ancilla_coordinates` (Phase B). Two spec decisions the RFC left implicit, now pinned:
  - **witness indices** = 0-based `error(...)` instruction indices in the committed flattened `.dem`, in file order — canonical and checkable by parsing the `.dem` alone;
  - `confidence` admits only `upper_bound` (the exact/MILP tier is deferred, so the schema can't invite unverifiable claims).
- **`verify/circuit_tools.py`**: the canonical noise recipe implemented as an *applier* — a circuit conforms iff it equals `apply_noise(strip_noise(circuit))` at the reference rate, making the recipe definitionally mechanical and giving submitters the exact tool to produce conforming circuits; DEM derivation; sparse GF(2) witness check; `ris_dem` (the existing RIS loop on `H_dem`, gf2_fast-accelerated, with witness extraction); the reference CSS memory-circuit builder with explicit-schedule support.
- **`verify/circuit_verify.py`**: RFC verification steps 1–3 + 5 — detector determinism, noise-recipe conformance, code binding (k independent declared-basis logicals on the final transversal readout, closure detectors matching the declared check rows exactly, declared round count = max mid-circuit measurement count of any qubit), byte-for-byte `.dem` re-derivation under the pinned stim, and the witness check with the penalty-only `d_circ <= d` clamp. All deterministic and cheap; step 6 (refutation search) stays in the gate and lands with the next PR.
- **`verify_all.py`** runs the fast path on any entry declaring a circuit block, so a merged claim is never CI-unchecked.
- **stim pinned `==1.16.0`**: the byte-for-byte `.dem` comparison makes the stim version part of the spec, so this is an exact pin; a bump must re-check the committed `.dem` artifacts and shows up as a reviewable diff.
- **15 adversarial tests**: the known-good artifact is generated in-test with the toolkit's own builder + searcher; each tamper (witness weight/detectability, stale DEM, missing or wrong-rate noise, TICK coarsening, non-deterministic detector, stabilizer-as-observable, inflated round count, unpinned stim) must be caught by its specific check.
- **`research/circuit_seed.py`**: seed-artifact generator — geometric zigzag plaquette schedules read off the 2D layout, pattern scan + deep RIS confirmation, writes `circuits/<slug>/` and the JSON block.

## Validation

- Full suite: 31 passed (18 existing + 13 new) with gf2_fast built; board sweep 341/341.
- On the [[25,1,5]] surface entry, greedy random schedules realize only d_circ = 3 on the Z-memory (a severe hook), and the generator's pattern scan independently recovers the standard perpendicular-zigzag schedule (Tomita–Svore) restoring 5/5 — deep-checked at 3×1500 RIS trials per basis. That artifact is the stacked seed PR.

## Revisions after review

- **DEM comparison is mechanism-for-mechanism, not byte-for-byte** (e455591): the seed PR's CI showed DEM probabilities differ in their last ulps across architectures (FMA contraction, arm64 vs x86). Structure, targets and file order must match exactly; probabilities to rtol 1e-9.
- **TICK layers must be genuinely parallel** (e16599b, closes vprusso's blocking finding): no qubit operated on twice per layer, so deleting TICKs can no longer shed idle-noise mechanisms and inflate d_circ; disjoint-layer merges (real pipelining) stay legal. Subsumes the within-layer parallelism gap.

## Known Phase-A limits (tracked for follow-ups)

- Refutation on `H_dem` in `gate_changed.py` + budget measurement: next PR.
- Fresh-ancilla measurement gadgets fall outside the round-count binding and would need a spec extension.


vprusso (changes requested):
The engineering here is strong and I want to say what I checked before the objection. The 13 adversarial tests pass and each one does fail for its own reason when I looked at what it tampers with. Defining the recipe as an applier (`circuit == apply_noise(strip_noise(circuit))`) is a genuinely better idea than a checklist of placement rules, because it makes conformance decidable rather than argued. `derive_dem` correctly leaves `allow_gauge_detectors` at its default, so a non-deterministic detector raises rather than silently producing a DEM. And pinning stim to `==1.16.0` is the right response to the byte-for-byte concern I raised on #505, though I could not test whether the same version formats those 20-digit probability literals identically across architectures, so I would treat that as reduced rather than closed.

The blocking problem is the TICK granularity item in your known-limits list. It is not a fidelity caveat. It is a score lever pointing the submitter's way, and at the limit it cancels the penalty-only property the tier rests on.

`apply_noise` puts `DEPOLARIZE1` on idle data at the end of each TICK-delimited segment, so segment count controls how much idle noise exists. Nothing in `circuit_verify.py` constrains TICK structure (no occurrence of TICK, segment or layer in the file), and `noise_recipe_errors` compares against the applier's output on the submitter's own skeleton, so a coarse skeleton is conforming by construction.

Measured on `codes/7-1-3.json`, rounds=3, same schedule and same seed, varying only TICK density:

| basis | ticks | mechanisms | conforming | d_circ |
|---|---|---|---|---|
| X | 40 | 277 | yes | 2 |
| X | 20 | 304 | yes | 2 |
| X | 10 | 263 | yes | 2 |
| X | 0 | 202 | yes | **3** |
| Z | 40 | 283 | yes | 2 |
| Z | 0 | 202 | yes | **3** |

Every row passes `noise_recipe_errors`. Deleting TICKs sheds about a quarter of the fault mechanisms and moves d_circ from 2 to 3.

Two consequences worth stating separately.

The scored quantity is `k·d_circ²/n`, so 2 to 3 is a 2.25x improvement bought by deleting annotations. And on this code `d = 3`, so the coarsened claim lands exactly on the clamp: `d_circ = d`. The penalty-only guarantee is not violated, it is neutralized. A circuit that a device would run worse than the honest one scores as though scheduling cost nothing.

The direction is the bad one. `build_css_memory` gives "every gate layer in its own TICK segment", which is the maximum-idle-noise, minimum-d_circ choice. So a submitter who copies the reference implementation you ship scores strictly worse than one who strips the ticks out. The tier would be rewarding the opposite of the behaviour it exists to encourage.

None of the 13 tests touch TICK structure (zero occurrences in the test file), which is consistent: the hole is in what conformance means, not in the checks written against it.

## What I would want before this merges

The recipe needs to pin layer structure, not just channel placement, because layer structure is what determines idle time. Options, roughly in order of how much I like them:

- Derive segmentation instead of reading it: apply idle noise per gate layer computed from the skeleton's data dependencies, ignoring submitted TICKs entirely. Then TICK becomes a comment and the lever disappears. This also subsumes the "within-layer gate parallelism is not yet enforced" item, since you would be computing the parallelism rather than trusting it.
- Or require one TICK per gate layer and reject circuits that pack multiple non-commuting gate layers into a segment, which is the rule `build_css_memory` already follows.
- Or, if this genuinely has to be Phase B, gate the tier on it: refuse `circuit` blocks until layer structure is pinned, rather than merging a scoring path with a known 2.25x lever in it. Shipping the verifier without the seed submission is already the shape of this PR, so holding the scoring semantics one more PR is cheap.

Whichever way, a test in the same adversarial style would nail it: same skeleton at two TICK densities must produce the same d_circ, or the coarser one must be rejected.

The rest of the slice I would take as-is.

FarLab:
Thanks — the demonstration was exactly right, and the lever is now closed in e16599b. I took your second option rather than derivation, deliberately:

**Conformance now requires every TICK layer to be genuinely parallel: no qubit operated on more than once per layer** (`layer_conflict_errors`, folded into `noise_recipe_errors` so a submitter's local self-check remains the gate they must pass). This rejects exactly the coarsenings that only exist on paper — your 0-TICK variant reuses every ancilla within one "layer" and fails immediately — while merges of genuinely *disjoint* layers stay legal. I think that boundary is the semantically right one: merging disjoint layers is real pipelining a device could execute, with a genuine idle-time reduction, i.e. schedule optimization this tier exists to score rather than gaming. It also subsumes the within-layer-parallelism item from the known-limits list in the same stroke.

On why not the derivation option, which I agree reads strongest: computing layers from data dependencies means re-timing operation atoms, and any reordering across measurement instructions renumbers the measurement record — which detectors and observables reference by relative index — so the canonicalized circuit is no longer the committed one, or the verifier must prove record-order preservation through the rescheduling. It would also force maximal pipelining on everyone, removing pipelining as a submitter degree of freedom rather than making it honest. The parallel-layer rule gets the same guarantee (claimed layer structure must be executable) without touching record order.

Your requested test is in as `test_tick_coarsening_rejected`: the TICK-stripped skeleton must be rejected (with an assertion that it demonstrably had fewer mechanisms — the lever), and the disjoint-merge variant (MX on X-ancillas fused with R on Z-ancillas) must stay conformant. The reference builder emits one gate layer per TICK and is unaffected; the seed artifact passes unchanged.

One note on your "reduced rather than closed" caveat about cross-architecture probability formatting: it materialized immediately — the seed PR's CI (x86) rejected the arm64-generated .dem on ulp-level probability differences (FMA contraction). Fixed in e455591: the comparison is now mechanism-for-mechanism (count, types, targets, file order all exact) with probabilities compared to rtol 1e-9, which ulp drift passes and any real tamper still exceeds. Unit-tested on both sides of the tolerance.

vprusso (approved):
Re-verified against e16599b rather than taking the summary's word. My original 0-TICK coarsening is now rejected with the layer-conflict message ("TICK layer 0 operates on qubit 9 more than once"), and the boundary you drew is the right one: a merge of genuinely disjoint layers is real pipelining with a real idle-time reduction, which is exactly what the tier should score. The argument against derivation also holds up; re-timing across measurement instructions renumbers the record that detectors reference by relative index, and I had not thought that through when I suggested it.

I also probed the adjacent lever the parallel-layer rule does not obviously cover: identity-padding idle data qubits so they count as touched (single-qubit gates carry no noise channel in the recipe, so an I per idle qubit per layer looked like it would shed all idle DEPOLARIZE1 while staying conformant). It does not fire: mechanisms and d_circ are identical to the reference on the same skeleton, seed and budget. So the recipe as implemented does not let a no-op gate shield a qubit from idle noise, and the one other way I could see to fake "not idle" is closed.

The rtol fix is the right response to the FMA drift, and the bounds are sane: 1e-9 relative passes ulp noise while any tamper that moves a probability by a schema-relevant amount is orders of magnitude above it, and the structural comparison (count, types, targets, order) stays exact, which is the part d_circ actually depends on.

15/15 on the suite here. Approving.

PR #650, opened 2026-08-20
Add [[345,9,7]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[345,9,7]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/345-9-7.json` passes locally

Pareto frontier in all 12 track cells. Score `kd^2/n = 1.278`, max check
(`locality.layers = 1`). This entry reaches `k = 9` at `d = 7` in the weight-4,
single-layer class.

claim:

  `kd^2/n` is 1.564, above this entry's 1.278; this ladder's ceiling at `d = 7`
  is `196/148 = 1.324`.
  and no exact claim is made. RIS evidence per trial weakens as `d` grows, so
  this `d = 7` bound is softer than the family's `d = 5` bounds at the same
  trial count.

This is the `m = 5`, `d = 7` rung of the one-parameter family whose `m = 3`
case is the board's dense-packing column `[[101,5,5]]` .. `[[677,5,13]]` — its
direct `m = 3` neighbour is `[[197,5,7]]`. Board rungs `[[177,9,5]]` (#632),
`[[215,11,5]]` (#633) and the open `d = 7`, `m = 4` rung `[[271,7,7]]` (#635)
trade `n` against `k` or `d` and are all mutually non-dominated with this
entry. Not equivalent to any board entry — `n` and `k` both differ. Recorded in
`provenance.notes`.

Research note: `notes/345-9-7.md`.


vprusso (approved):
🚀  Once verification succeeds

PR #651, opened 2026-08-20
Add [[447,7,9]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[447,7,9]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/447-7-9.json` passes locally

Pareto frontier in all 12 track cells. Score `kd^2/n = 1.268`, max check
(`locality.layers = 1`). This extends the dense-packing ladder to `d = 9`.

Stated plainly:

  `kd^2/n` is 1.564; this ladder's ceiling at `d = 9` is `324/244 = 1.328`.
  side, submitted as `upper_bound` on both sides. RIS evidence per trial
  weakens as `d` grows, so this is the softest bound among the ladder's
  submitted rungs; the note spells out the structural argument (identical
  fusion seams to the `d = 5` and `d = 7` rungs, where witnessed distance
  equals patch distance throughout) that carries the corresponding weight.

This is the `m = 4`, `d = 9` rung of the family whose `m = 3` case is the
board's dense-packing column `[[101,5,5]]` .. `[[677,5,13]]` — its direct
`m = 3` neighbour is `[[325,5,9]]`. The board rungs `[[177,9,5]]` (#632),
`[[215,11,5]]` (#633), `[[253,13,5]]` (#645), `[[271,7,7]]` (#635) and
`[[345,9,7]]` (#650) trade `n` against `k` or `d` and are all mutually
non-dominated with this entry. Not equivalent to any board entry — `n` and `k`
both differ. Recorded in `provenance.notes`.

Research note: `notes/447-7-9.md`.


vprusso (approved):
🚀  Once verification succeeds

PR #652, opened 2026-08-20
Add [[569,9,9]]

Xo1otl:
## Code submission

- Parameters: [[n, k, d]] = [[569,9,9]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/569-9-9.json` passes locally

Pareto frontier in all 12 track cells. Score `kd^2/n = 1.281`, max check
(`locality.layers = 1`). This entry reaches `k = 9` at `d = 9` in the weight-4
single-layer class.

Stated plainly:

  `kd^2/n` is 1.564; this ladder's ceiling at `d = 9` is `324/244 = 1.328`.
  side, submitted as `upper_bound` on both sides. At `n = 569` this is the
  largest rung submitted so far, so per-trial RIS evidence is at its softest;
  the note spells out the structural argument (identical fusion seams to every
  other rung, where witnessed distance equals patch distance throughout) that
  carries the corresponding weight.

This is the `m = 5`, `d = 9` rung of the family whose `m = 3` case is the
board's dense-packing column `[[101,5,5]]` .. `[[677,5,13]]`. The board rungs
`[[177,9,5]]` (#632), `[[215,11,5]]` (#633), `[[253,13,5]]` (#645),
`[[271,7,7]]` (#635), `[[345,9,7]]` (#650) and the open `[[447,7,9]]` (#651)
trade `n` against `k` or `d` and are all mutually non-dominated with this
entry. Not equivalent to any board entry — `n` and `k` both differ. Recorded in
`provenance.notes`.

Research note: `notes/569-9-9.md`.


vprusso:
Thanks for this, @Xo1otl . Looks like the verify check is failing, but might be worth fixing and re-running?

Xo1otl:
Thanks for the look! The code is unchanged since the earlier green run, and the failure is the job hitting its 45-minute cap — the deep refutation on this entry (n=569, the largest in this batch) just ran long on that seed, not a refutation. Retriggering CI once; if it times out again I will withdraw this one as too heavy for the CI budget rather than keep burning runner time.

Xo1otl:
Re-run is green — the earlier failure was the 45-minute cap on that seed, as suspected. No changes to the submission.

PR #653, opened 2026-08-20
Harden external path checks in prose

cosenal:


cosenal:
Closing because an AI-generated over-engineered collection of regex patterns

Issue #654, opened 2026-08-20
Gate the refutation budget by diff class: skip layout-only, shallow for tightenings


willzeng:
The CI refutation gate (`verify/gate_changed.py`) prices the code, not the diff: any change to a code file triggers the full budget, including the deep battery for record holders. That is overkill for two diff shapes that cannot introduce a new over-claim:

- **Layout-only** (checks and distance byte-identical to base): the claim already survived the gate at merge and the weekly `refute_board.py` sweep keeps attacking it; everything a layout adds is recomputed deterministically. With the layout-filling effort (~261 entries missing g scores) each 15-entry batch PR currently costs ~45 min of redundant gate time.
- **Tightening** (a refutation-shaped correction: values strictly down, new witnesses of exactly the new weights, upper_bound confidence, nothing else touched): the entry just became strictly less wrong; a shallow pass catches gross remaining over-claims and the weekly sweep provides depth.

One diff shape needs the opposite treatment: adding or raising `witness_provenance.survived_samples` claims a deep null result (and deters future refuters per SCHEMA.md), so it should get the deep battery regardless of record status.

Everything else — new codes, edited checks, raised values, mixed or unclassifiable diffs — keeps today's behavior, fail-closed. Classification must be recomputed from the git diff, never taken from the PR's framing, and skips must print receipts naming the weekly sweep as the backstop.

FarLab:
Completed by #655.

PR #655, opened 2026-08-20
gate_changed: price the refutation budget by diff class (issue #654)

willzeng:
## What

The refutation gate re-ran its full search on every changed code file, pricing the code rather than the diff. This classifies each changed file against its base-revision counterpart (rename-aware, same pairing logic as `check_authorship.py`) and budgets accordingly:

| Diff class | Condition (structural, recomputed from the git diff) | Budget |
|---|---|---|
| new | no base counterpart | unchanged (deep for cell records, standard otherwise) |
| layout-only | `checks` and `distance` byte-identical to base | **skip**, with a receipt naming the weekly sweep as backstop |
| tightening | values strictly down, new witnesses of exactly the new weights, `upper_bound` at corrected values, `d = min(dX,dZ)`, checks/n/k/locality untouched | **standard shallow pass** |
| survival stamp | `witness_provenance.survived_samples` added or raised on any side | **deep battery** regardless of record status |
| anything else | edited checks, raised values, mixed layout+distance, unclassifiable, unreadable base | full existing behavior (fail closed) |

## Why each is sound

- **Layout-only**: the distance claim already survived this gate at merge and the weekly `refute_board.py` sweep keeps attacking it with fresh seeds; everything a layout adds (radius, spacing, class, g) is recomputed deterministically by the cheap verifier that runs earlier in the same job. A layout cannot make a distance claim more wrong.
- **Tightening**: the new witness proves `d <= value` trustlessly in the verify_all step; the entry just became strictly *less* wrong than what is on the board. The shallow n-scaled pass catches gross remaining over-claims; the weekly sweep provides depth.
- **Survival stamps**: per SCHEMA.md, `survived_samples` is the budget that tells the next refuter what to beat — a deterrent claim. It gets *more* scrutiny than before (deep even for non-record codes), closing a gap rather than opening one.

Every skip and reclassification prints its reason and lands in the gate receipts (`diff_class` / `diff_reason` fields), so the audit trail records a decision, never an omission.

## Impact

A 15-layout batch PR (the mass g-score-filling effort, ~261 entries) drops from ~45+ minutes of redundant gate time to seconds; refutation corrections drop from ~10 minutes to under one. New-code submissions keep exactly today's scrutiny.

## Tests

`verify/test_gate_diff_classes.py`: 18 cases — the classifier table row by row, the adversarial shapes (stamp riding on a tightening, mixed layout+distance, witness swap at unchanged value, exact-at-new-value, weight/value mismatch, garbage input), and a git-fixture integration test of the base-doc pairing across in-place edits, refutation renames, and new files. Full `run_tests.py --skip-slow` green; `gate_changed.py` is not hash-pinned, so no manifest change.

### DIFF
--- a/verify/gate_changed.py
+++ b/verify/gate_changed.py
@@ -17,6 +17,23 @@ size, so a small code gets near-exhaustive coverage and a large one a
 proportionate search; frontier-advancing codes get several independent deep
 seeds (tens of thousands of trials each) before they earn the record.
 
+The gate prices the DIFF, not just the code (issue #654): each changed file
+is classified against its base-revision counterpart, because what changed
+determines what could newly be over-claimed. A layout-only diff (checks and
+distance byte-identical to base) skips refutation entirely -- the claim
+already survived the gate when it merged and the weekly board sweep keeps
+attacking it; everything a layout adds is recomputed deterministically by the
+verifier. A tightening diff (a refutation-shaped correction: values strictly
+down, new witnesses of exactly the new weights, upper_bound confidence,
+nothing else touched) gets the standard shallow pass -- the entry just became
+strictly less wrong, and the weekly sweep provides depth. A diff that adds or
+raises witness_provenance.survived_samples gets the deep battery regardless
+of record status: a survival stamp claims a deep null result and deters
+future refuters, so it is itself the claim being vetted. Anything else --
+new codes, edited checks, raised values, unclassifiable diffs -- fails closed
+to the full existing behavior. Classification is recomputed from the git
+diff, never taken from the PR's framing.
+
 Usage:
   python verify/gate_changed.py [--code-root PATH] [BASE] [files...]
     --code-root PATH  repository tree containing the submitted codes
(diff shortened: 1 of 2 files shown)


vprusso (changes requested):
The design is sound and I checked the pricing table against the actual attack surfaces rather than the description. Classification is recomputed from the git diff, layout-only requires checks and distance byte-identical (so it composes correctly with the #643 binding), the tightening class re-derives every refutation-shaped condition instead of trusting the PR, and a survival stamp goes deep even when it rides on a tightening, which is the case where a submitter would most like it to hide. The stamp check running before the frozen-field check is also right, since deep is the strictest budget on every path, and a stamp-plus-checks-edit dies at the authorship gate anyway. The 18 unit tests pin each class from structure alone, and the git-repo pairing test covers the rename path with real diffs. All pass here.

Three changes requested, none touching the classification logic.

1. Drop the `sys.path.insert` from `verify/test_gate_diff_classes.py`. #615's root `conftest.py` puts `verify/` on the path before collection, and this file is exactly the boilerplate that PR removed from the other eleven test files. (The pre-existing path lines in `gate_changed.py` itself are a different matter: runtime scripts still need them because the repo is deliberately not installable, so those stay.)

2. Narrow the exception handling. Three places catch `Exception`:
   - `changed_codes_status` and `load_base_doc`: the failures these can legitimately see are `subprocess.CalledProcessError`, `OSError` and (for the loader) `json.JSONDecodeError`. Catch those; anything else is a bug that should surface as a traceback in CI, not as a silent full-budget run.
   - the blanket `try/except` around the whole of `classify_diff`: fail-closed is the right default, but catching `Exception` turns a future typo in this function into "every diff prices full" with only a receipt string to notice it by. The unit tests do pin the classes, so a regression would fail loudly there, which is why this is a request and not a blocker; still, `(KeyError, TypeError, AttributeError, ValueError)` covers every malformed-document shape the docstring is guarding against, and letting the rest raise costs nothing.

3. Add type annotations to the new public functions (`changed_codes_status`, `load_base_doc`, `base_doc_for`, `classify_diff`). The repo has no annotations today, so this is new-code-leads rather than matching the neighbours; a repo-wide linting and typing standard is about to land as its own issue, and these functions are the kind it will ask for.

One observation, no action needed: `base_doc_for` re-runs the `git diff` subprocess once per changed file. PRs carry one code, so this costs nothing today; if the gate ever batches, hoist the call.

FarLab:
Checked that layout validation still runs on a layout-only PR under this change: it does. Layout checks live in `qldpc_verify.verify` step 9, which `verify_all.py` runs from the trusted tree on every `codes/` change, one CI step before the gate — this PR only skips the RIS distance search. The layout-only branch also calls `validate_candidate(doc, refute=False)`, so the receipt still records the structural verdict. Confirmed by checking out the branch and pushing three layout-only diffs through `verify_all` + the new gate:

| Test diff | `verify_all` | gate | receipt |
|---|---|---|---|
| honest grid layout on `40-10-4` | PASS (schema enforces `schema_version: "0.2"`) | layout-only, skipped | `structural_result: passed` |
| 39 coords for n=40 on `40-12-4` | **FAIL** `coordinates_cover_all_qubits` → job fails | layout-only, skipped, prints `ok` | `structural_result: failed` |
| spacing-0.1 layout on `40-6-4` | PASS, class computed `unrestricted` | layout-only, skipped | `structural_result: passed` |

Classifier looks tight (`checks`/`n`/`k`/`code_type` frozen, whole `distance` compared, stamps go deep, everything else fails closed); 20/20 tests pass locally; branch is up to date with main.

Three observations. **The first two are pre-existing behavior, not introduced by this PR** — noting them here because they bear on what "layout checks" actually guarantee:

1. **Pre-existing: dishonest layouts merge green.** The verifier only hard-fails on coordinate count, schema, and `interaction_radius_within_claim` (if the submitter claims one). A layout that violates spacing ≥ 1 / qubits-per-site ≤ layers / radius cap is silently demoted to `unrestricted` (`verify/qldpc_verify.py` step 9, by design). For a mass layout-filling batch, a botched layout merges and earns nothing with no signal. Worth a follow-up issue: fail when `locality` is present but earns no class.

2. **Pre-existing (but now more visible): gate prints `ok` on a file the verifier rejected** (`40-12-4` above). Harmless in CI since `verify_all` already failed the job, but the gate's stdout and its receipt disagree. One-line fix in the layout-only branch: count `verdict["passed"] == False` toward `failed` (the normal path has the same gap).

3. **New in this PR — the one real scrutiny regression.** Adding a layout can move a code into a 2d-local cell where it becomes a *new cell record*. Before this change `board_record_slugs` read the submitted layout (via `_locality_rank`) and ran the deep battery on it; now it's skipped. So a code that only ever had the shallow pass (dominated in unrestricted) can become a local-2d record with no deep pass ever, and the weekly sweep is bounded single-seed RIS, not the deep battery. Cheap fix: if `cls == "layout-only"` and `slug in records` but the slug was *not* a record at base, run deep instead of skipping. I'd ask for this one before merge.

Reviewed with Claude Fable 5.

willzeng:
Fixes #654

willzeng:
Both reviews addressed:

**@FarLab's item 3 (the scrutiny regression)** — fixed as requested, slightly generalized: a layout-only diff that (a) moves the code into a tighter locality class than base (`layout_entered_tighter_class`, computed with the gate's own `_locality_rank`) and (b) lands in `board_record_slugs` now reclassifies to `layout-record` and runs the **deep battery** instead of skipping. Since n/k/d/w are unchanged by a layout-only diff, record status in cells the code already occupied cannot change — entering a tighter class is the only path to a new record, so the rank comparison is equivalent to the was-not-a-record-at-base check without reconstructing the base board. Covered by three new guard tests (honest layout fires it; dishonest layout that earns no class does not; unchanged class does not).

**Item 2** — the layout-only branch now always runs `validate_candidate(refute=False)` and counts a failed structural verdict toward the gate's exit code with a FAIL line, so stdout and the receipt can no longer disagree. (The same gap on the normal path is pre-existing and left for a follow-up, per your scoping.)

**Item 1** — opened as a follow-up issue (dishonest layouts silently demoting to `unrestricted`).

**@vprusso's three**: `sys.path` boilerplate dropped from the test file (conftest covers it); exceptions narrowed (`CalledProcessError`/`OSError`/`JSONDecodeError` at the subprocess/loader sites, `(KeyError, TypeError, AttributeError, ValueError)` around `classify_diff` so real bugs raise in CI); type annotations added to the new public functions. The per-file `git diff` in `base_doc_for` is noted for hoisting if the gate ever batches.

21 classifier tests + full `run_tests.py --skip-slow` (39 passed) green; integrity pin untouched.

vprusso (approved):

FarLab (approved):

PR #662, opened 2026-08-20
Harden generated site against stored XSS

### LINKED ISSUE #649: Harden generated-site encoding for untrusted submission metadata
## Summary

`codes/*.json` is untrusted input, but `site/build.py` currently relies on manual, per-use escaping. Some attribute and inline-JSON contexts are not encoded consistently, allowing crafted metadata to alter generated markup.

This is a focused follow-up to #594.

## Proposed work

- Validate submission-derived URLs before rendering links.
- Use one HTML-safe JSON serializer for every embedded data block.
- Audit current free-text metadata across text, attribute, URL, SVG, and client-rendered contexts.
- Add adversarial regression tests for every unrestricted string field.
- Make CI flag newly added unrestricted string fields until their rendering behavior is covered.

## Acceptance criteria

- Crafted metadata renders only as inert content.
- Embedded JSON remains parseable and cannot terminate its containing element.
- Valid references and existing site behavior remain unchanged.
- The complete board builds and the regression suite passes.

Broader CSP or template-engine changes are defense-in-depth follow-ups, not required for this fix.

bachase:
## Summary

- validate submission-derived reference links and code slugs before placing them in HTML attributes
- serialize every embedded JSON data block through one HTML-safe helper and escape model labels inserted by the client-side chart
- add adversarial coverage for every free-form schema string, with a schema inventory that fails when future fields are added without classification
- include site security tests in the documented test runner

This intentionally keeps the current string-based generator. Template-engine and broader CSP changes remain separate follow-ups.

## Verification

- `uv run --frozen python run_tests.py` (37 passed, 6 skipped)
- clean temporary site build (345 detail pages)

Closes #649

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,8 +1,8 @@
 """Run the test suite. A thin wrapper over pytest, kept as the documented entry
 point so CI and the README command do not change.
 
-Discovery is still by convention -- pytest collects verify/test_*.py and
-research/test_*.py -- so a new test joins the suite the moment it exists and
+Discovery is still by convention -- pytest collects test_*.py under verify/,
+research/, and site/ -- so a new test joins the suite the moment it exists and
 cannot be silently left out of CI the way a hand-listed step can (which is how
 two of five tests once went unrun).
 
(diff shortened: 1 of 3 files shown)


FarLab (approved):

MathysRennela (approved):
LGTM!

PR #663, opened 2026-08-20
Circuit-tier refutation in the gate: bounded RIS on H_dem (RFC 0001 step 6)

FarLab:
Completes Phase A of #505: the adversarial half of the circuit tier. The fast path (#646) checks that a claimed witness is real; this PR makes the claim *contested* — `gate_changed.py` now independently searches every changed entry's memory DEMs for an undetected logical fault set lighter than the claimed d_circ, with the same trust layering as the code-distance gate:

- The DEM is **re-derived from the committed `.stim`** with the pinned stim — the search substrate is never the committed `.dem` taken on faith.
- A find counts **only after the pinned GF(2) witness check confirms it**; the printed witness is itself the corrected claim a follow-up submission can adopt.
- **Fails closed**: unreadable or over-cap artifacts are a gate failure, never a silent pass.
- **Schedule swaps can't dodge review**: `map_changed` maps `circuits/<slug>/*` diffs back to `codes/<slug>.json`, so a PR touching only the `.stim` files still pays the full gate.

## Budget (the RFC's "measure before inviting submissions" instruction)

Measured per-trial RIS cost on `ker(H_dem)` fits ~`1.2e-12 · m³` seconds with gf2_fast (8.4 ms at m=1651, 26 ms at m=2845, 3.9 s at m=15800, 25 s at m=34530; python fallback ~10×). The gate targets **120 s wall-clock per basis**, converted to trials through that fit (floor 12, cap 2000), with a hard time cap inside `ris_dem` for when the fit is off.

That measurement also sized the tier's new submission cap: **`MAX_DEM_MECHANISMS = 25 000`** (enforced in `circuit_verify`, raise-only) — at the cap the trial floor costs ~2 min/basis, keeping a circuit entry inside the same ~10-minute budget a deep code claim gets. This is the RFC's anticipated "cap the tier by n·rounds", enforced on the actual cost driver. A weight-6 BB entry at rounds=6 sits at m≈16k — inside the cap, ~30 gate trials/basis; the known routes up are GPU RIS and sparse RREF.

On the seed entry the gate runs 2000 trials/basis in ~17 s each and finds nothing below the claimed 5/5 (which incidentally deepens that claim's survived budget).

## Tests (5 new, suite 38 passed)

The over-claim fixture is organic rather than synthetic: the greedy schedule on [[25,1,5]] is hook-limited to a true bound of 3, and a 1-trial search returns a heavier valid witness — exactly the under-searched submission the RFC's convergence risk describes. The gate must refute it with a validated witness; claiming the deep-searched bound must survive the same search; circuits/-only diffs must map to the owning code; missing artifacts must raise; the budget curve must have full depth at small m, a feasible floor at the cap, and a strictly shallower python fallback.

### DIFF
--- a/schema/SCHEMA.md
+++ b/schema/SCHEMA.md
@@ -141,6 +141,9 @@ limits are:
 - At most 200000 total support entries across all checks.
 - At most 600 locality coordinate entries.
 - Dense verifier intermediates capped at 50000000 cells.
+- Circuit tier: at most 25000 DEM error mechanisms per memory circuit
+  (verification-budget rule: the refutation gate must be able to search the
+  DEM; raise-only as the search stack improves).
 
 These are far above the current board entries. A larger code should be handled
 through a maintainer-run path until the verifier is sparse end-to-end.
--- a/verify/circuit_tools.py
+++ b/verify/circuit_tools.py
@@ -322,11 +322,14 @@ def _rref(M):
     return gf2.rref(np.ascontiguousarray(M))[0]
 
 
-def ris_dem(H, L, trials, seed=0, pair_top=24):
+def ris_dem(H, L, trials, seed=0, pair_top=24, max_seconds=None):
     """RIS upper-bound search for the lightest undetected logical fault set:
     min |e| with H e = 0, L e != 0. The code-tier searcher on the DEM's
     parity-check matrix -- hyperedge degree is irrelevant to it. Returns
-    (weight, sorted column indices) or (None, None)."""
+    (weight, sorted column indices) or (None, None). Stops after `trials`
+    permutations or `max_seconds` wall-clock, whichever comes first (the time
+    cap keeps the CI gate bounded when the size estimate is off)."""
+    import time
     K = _kernel_basis(np.asarray(H, dtype=np.int8))
     m = K.shape[1]
     if K.shape[0] == 0:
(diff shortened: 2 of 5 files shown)


vprusso (approved):
merge conflicts, but LGTM!

FarLab:
Rebased onto main to resolve the conflict with #655 (diff-priced refutation). The resolution is more than mechanical — the two features had to be composed, and the naive merge had two holes worth recording:

1. **A circuit-block-only edit classifies as `layout-only`** under #655 (checks and distance are unchanged — correctly so for the code tier), but the layout-only path `continue`d before the circuit search: a changed d_circ claim would have merged unsearched. The layout-only branch now runs the circuit gate before skipping, and `test_circuit_block_edit_skips_code_tier_but_is_searched` pins the conjunction so a future change to either half resurfaces it.
2. **A circuits/-files-only diff** (JSON untouched, mapped in by `map_changed`) had no base counterpart in #655's codes/-scoped status and would have been priced `new` — the deepest code battery for an unchanged code claim. The gate now loads the base doc directly in that case, so the code tier classifies as unchanged (skip) while the changed circuit surface still pays the full circuit search.

The circuit search itself is now diff-priced by the same principle (`circuit_gate_needed`): it runs iff the circuit claim surface changed (JSON block or `circuits/<slug>/` artifacts) or the entry is new; an untouched claim — e.g. a layout tweak on an entry that ships circuits — skips it, exactly as #655 skips the unchanged distance claim. Receipts record `gate.circuit.searched` alongside the verdict.

Suite after the merge: 67 passed, including #655's classification tests unchanged.

PR #667, opened 2026-08-20
Pin the production validator execution closure

cosenal:
## What was wrong

`verify/validator_manifest.json` claimed to pin the trusted validation stack, but it contained hashes for only five files. Several production entrypoints and dependencies that can change a CI verdict were outside that set, including `verify/gate_changed.py`, `verify/refute_board.py`, the authorship/prose/scope checks, the schema, the circuit verifier, the optional C++ refuter, and its build path.

This was not a way for a code submission to weaken its own gate in the same PR: submission PRs are checked with the base branch's trusted tree, and mixed `codes/` plus verifier changes are rejected. The gap was a quieter two-stage problem. A verifier-only change could alter an unpinned gate without requiring a manifest update; after merge, later submissions would inherit that changed file as trusted code without the expected re-pin signal.

## Example

Before this change, a PR could modify `verify/gate_changed.py` to reduce the refutation search while leaving `verify/validator_manifest.json` untouched. The integrity check would still report that all five pinned files matched.

After this change, `verify/gate_changed.py` and the rest of the current repo-local production verdict/policy closure are hashed. Changing one makes the integrity check fail until the manifest is deliberately regenerated, leaving a small, explicit hash diff for maintainers to review.

## What changed

- Expanded the manifest from 5 files to 25 repo-local production files: CI entrypoints, local Python dependencies, workflow definitions, dependency configuration, schema data, the trusted fixture, the circuit-tier verifier modules, and the C++ accelerator source/build path.
- Switched manifest paths to repository-relative names and added `--root` so the checker can validate another checkout.
- Made missing trusted files and malformed manifests fail closed.
- Made newly added JSON verifier fixtures require a pin automatically.
- Added `Makefile` and `decode/` to the verify workflow triggers, submission-critical scope, and CODEOWNERS coverage so their hashes are actually enforced.
- Corrected stale accelerator comments that said CI did not build or use `verify/gf2_fast.cpp`.
- Added regression tests that tamper with every pinned file individually, remove a trusted gate file, add an unpinned fixture, and assert that every manifest path is submission-critical.

The boundary is intentionally the production validation and policy-gate implementation. Variable submissions and PR inputs, post-merge site generation, unused offline research and certification tools, the separate pytest and research self-test harness, and external tools such as GitHub Actions and the compiler are not described as hash-pinned by this manifest.

This PR is limited to the integrity-manifest coverage gap and the wiring needed to enforce it; it does not attempt a broader security audit.

## Testing

- `python3 verify/check_validator_integrity.py` - 25 files match
- `uv run --frozen pytest verify/test_validator_integrity.py` - 29 passed
- `uv run --frozen python run_tests.py --skip-slow` - 87 passed, 6 skipped
- workflow YAML parse check
- `git diff --check`


cosenal:
This comes from looking into Track 2. of @FarLab's hackday email

vprusso (approved):
Verified rather than read, since this file is the trust anchor for everything else.

The manifest matches the PR tree (25 files, checker ok) and all 29 tests pass here. I checked the closure claim mechanically: an AST walk over every pinned Python file, resolving each repo-local import against the manifest, finds no unpinned dependency, so the closure is actually closed rather than just larger. And I demonstrated the gap in both directions: a tampered `verify/gate_changed.py` sails through main's 5-file checker ("ok: trusted validation stack matches the pin") and is caught by this PR's ("changed (pinned 744286a97863..., now ...)"). That tamper is exactly the example the PR body gives, and it is real.

The `decode/` catch is the part I would have missed. `gate_changed.py` uses `decode/distance.py` as the second refutation mechanism, so it has been verdict-relevant all along while sitting outside CODEOWNERS, outside the workflow triggers, and outside the scope check's critical set. All three are fixed here together with the pin, which is the right shape: a pin without the CODEOWNERS line would be a tripwire nobody is required to look at.

Worth stating plainly what this does and does not enforce, because the docstring gets it right and the "Example" section could be over-read. On pull_request events GitHub runs the PR's own workflow copy, and for verifier-change PRs the integrity check runs from the submitted tree, where the same PR can regenerate the manifest. So the pin is not self-enforcing against a malicious verifier PR; it makes the change loud and un-forgettable in one CODEOWNERS-gated file. The enforcement is the owner review of that diff. The rewritten docstring says exactly this ("the manifest is the durable review signal for deliberate changes to that checkout"), which is the honest claim, and the fail-closed handling of missing files and malformed manifests removes the quiet exits the old checker had.

Two operational notes, neither blocking:

- #655 modifies `gate_changed.py` and is mid-review; whichever of these merges second needs a re-pin, and that manifest diff will now appear in the later PR by design. Same for any circuit-tier follow-up touching `circuit_tools.py`/`circuit_verify.py`. That is the mechanism working, but the first few contributors to hit it will need the `--update` incantation pointed out, so expect a little friction.
- `test_execution_closure_is_submission_critical` is a quietly important test: it welds the manifest to `check_submission_scope.py`'s critical set so the two lists cannot drift apart. If either list is ever restructured, keep that weld.

The fixture auto-discovery (a new `verify/fixtures/*.json` forces a pin) closes a door I had not thought about: `verify_all.py` validates every fixture in that directory, so an unpinned fixture would be unreviewed trusted input.

Approving.

FarLab:
Needs a re-pin before merging: `verify/gate_changed.py` changed on main via #655/#678, so the integrity check fails on this branch until `python verify/check_validator_integrity.py --update` is run after merging main (I tested it — it's a one-line manifest diff and everything else is green).

vprusso:
Has some merge conflicts as well, but LGTM @cosenal once those are resolved.

MathysRennela (approved):

PR #668, opened 2026-08-20
Add [[64,18,d<=8]] 2BGA from Lin-Pryadko database: GAP SmallGroup(32,21)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[64,18,d<=8]]
- Tracks: unrestricted / weight-8 (computed by the verifier from H and the layout)

### Checklist
- [x] `python verify/qldpc_verify.py codes/64-18-8.json` passes locally
- [x] Model version specified: Solar Pro 4 (Upstage AI)

- **unrestricted / weight ≤ 8**: on the Pareto frontier — advances the cell with kd²/n = 18.0
- **unrestricted / any weight**: on the Pareto frontier

Score kd²/n = 18.0, max check weight 8, locality class unrestricted.

Construction: 2BGA from Lin-Pryadko database: GAP SmallGroup(32,21), a=1+g2+g3+g17, b=1+g7+g22+g31

Research note: `notes/64-18-8.md`


vprusso (changes requested):
Verified the reconstruction at the row level rather than trusting the citation: the Lin-Pryadko database's `diswtabelian_wt8wtL4_order32_k18.txt` contains exactly `32 21 18 8 [2,3,17] [7,22,31]`, which is this construction generator for generator, with their exact d = 8. The file lives in the abelian archive, so SmallGroup(32,21) is abelian and the generalized-bicycle family tag is defensible. I also attacked the distance independently: 2M RIS trials per side (400x the gate's budget) found nothing below 8, and at n = 64 that budget is far past where small-code searches converge. The frontier claim holds on my own recomputation: nothing on the board dominates [[64,18,8]] at w = 8, and the nearest entries ([[64,16,6]] w=10, [[65,17,3]] w=8) are strictly behind it.

Two metadata fields need filling before this merges, both one-line edits:

- `provenance.novelty` is unset. The schema has `known_parameters` for exactly this case, and the note itself says the entry is a direct reproduction. Without the tag the board renders this as an undistinguished submission; with it, the row carries the "known params" chip that tells a reader the parameter set is from the literature and the contribution is the reconstruction. That chip is the honest framing of what this PR is, and it exists so that reconstructions do not read as discoveries.
- `provenance.references` is empty. The source is named only inside the free-text construction string; `references` is where it belongs so the detail page links it. `["github.com/QEC-pages/2BGA-codes @ 403d194c", "arXiv:2306.16400"]` covers the database and the paper behind it.

Keeping the confidence at `upper_bound` despite the database's exact value is the right call and the note explains it well.

PR #669, opened 2026-08-20
Add [[66,20,d<=7]] cyclic bicycle from arXiv:2608.09115

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[66,20,d<=7]]
- Tracks: unrestricted / any weight (computed by the verifier from H and the layout)

### Checklist
- [x] Model version specified: Solar Pro 4 (Upstage AI)

- **unrestricted / any weight**: on the Pareto frontier — advances the cell with kd²/n = 14.848

Score kd²/n = 14.848, max check weight 19, locality class unrestricted.

Construction: Cyclic bicycle over F_2[Z_33], from arXiv:2608.09115v1 Table II/III

Research note: `notes/66-20-7.md`


vprusso (changes requested):
Verified independently rather than from the citation: arXiv:2608.09115's abstract itself names [[66,20,7]] at kd²/n = 14.85 with the distance certified exactly through the Calderbank correspondence, so the source supports both the parameters and the value. My own attack found nothing lighter at 2M RIS trials per side, 400x the gate's budget and well past convergence at n = 66. The frontier claim holds on recomputation: nothing on the board dominates [[66,20,7]] at w = 19, and kd²/n = 14.85 is a real advance for the any-weight cell at small n.

Same two metadata gaps as #668, both one-line edits:

- `provenance.novelty` should be `known_parameters`: this is a direct reproduction of a published table entry, the note says so plainly, and the chip that field drives is the board's designed way of saying it.
- `provenance.references` should carry `["arXiv:2608.09115"]` rather than leaving the source only in the construction string.

One thing worth stating for the record, not a change request: at w = 19 this is a dense code by LDPC lights, and it lands only in the any-weight cell, which the PR body correctly reflects. The conservative `upper_bound` confidence despite the paper's exact certification is again the right call.

PR #678, opened 2026-08-20
gate_changed: layout-only structural verdict must come from the verify gate (fixes #674 CI)

willzeng:
## The bug

#655's layout-only branch runs `validate_candidate(doc, refute=False)` and treated the overall `passed` as the structural verdict (the receipt-honesty fix from review). But `validate_candidate` is the **new-candidate** pipeline: on an in-place edit of a file already on the board, its dedup gate flags the code as an exact duplicate *of itself* (`exact_duplicate_of: '<slug>.json'`) and novelty reports no board advancement — so `passed` is `False` for every modified code regardless of correctness. Result: all ten layouts in #674 FAILed with 'structural checks failed' while `verify_all` passed them in the same job. The normal gate path never surfaced this because it never counted the verdict toward the exit code (the pre-existing gap noted in #655's review).

## The fix

New `structural_ok(verdict)` helper: the authoritative structural signal for an in-place edit is `gates.verify.ok` — the thing the gate must not say ok about is a file the VERIFIER rejects; dedup/novelty are meaningless for a file that is on the board by definition. The layout-only branch uses it for the FAIL decision, and the receipt's headline `passed` is set from it with the dedup self-match noted as expected in the refute detail.

Regression test added (verify-ok-but-dedup-self-dup must pass; verifier-rejected must fail; empty verdict fails). 22 classifier tests + full `run_tests.py --skip-slow` green.

After merge: re-run CI on #674 and #675 — their layouts are unaffected (all pass `qldpc_verify` locally, including layout honesty); only this gate misjudgment blocked them.

### DIFF
--- a/verify/gate_changed.py
+++ b/verify/gate_changed.py
@@ -144,6 +144,17 @@ def base_doc_for(f: str, doc: dict, base: str, code_root: str) -> dict | None:
     return None
 
 
+def structural_ok(verdict: dict) -> bool:
+    """The structural soundness of a board edit, from a validate_candidate
+    verdict. validate_candidate is the NEW-candidate pipeline: its overall
+    'passed' includes gates that are meaningless for an in-place edit of a
+    file already on the board -- dedup flags the code as an exact duplicate
+    of itself, and novelty judges board advancement. What the gate must not
+    say ok about is a file the VERIFIER rejects, so only gates.verify.ok is
+    authoritative here."""
+    return bool(((verdict.get("gates") or {}).get("verify") or {}).get("ok"))
+
+
 def layout_entered_tighter_class(base_doc: dict | None, new_doc: dict) -> bool:
     """True when the diff's layout moves the code into a tighter locality
     class than it had at base. For a layout-only diff this is the only way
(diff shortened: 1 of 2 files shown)


vprusso (approved):
Reproduced the premise directly rather than from the CI log: `validate_candidate` on `codes/40-10-4.json` (one of #674's ten) returns `passed: False` with `verify.ok: True` and `dedup.exact_duplicate_of: "40-10-4.json"`. The file fails the new-candidate pipeline purely by being its own duplicate, which is definitionally true of every in-place edit. So the bug is exactly as described, and it is a consequence of the receipt-honesty change I asked for in the #655 review: I asked for the verdict to count, and the verdict being counted was the wrong one for this diff class. Fair cop.

The fix draws the right line. For a layout-only diff the classifier has already proven checks and distance byte-identical to a merged entry, so the only new content is the layout, and `qldpc_verify` is precisely the check that vets it (including the honesty test and class derivation). Dedup and novelty are statements about candidacy, not soundness, and they cannot be meaningful for a file whose presence on the board is the premise of the diff class. `structural_ok` reads only `gates.verify.ok`, fails closed on an empty or malformed verdict (the chained `or {}` handles a missing gates block, and the test pins it), and the receipt keeps the full verdict with the self-match explained in the refute detail rather than scrubbed, so an auditor reading the receipt sees what dedup said and why it was overridden.

Checked the obvious follow-on worry: nothing else in the gate consumes `verdict["passed"]` for a decision; the only other consumer is the receipt itself. And the mutation happens before `make_receipt`, so receipt and exit code agree.

22/22 on the classifier tests here. Approving; re-running #674 and #675 after merge is the right sequencing, and their layouts pass `qldpc_verify` locally so they should go green on the re-run.

Issue #460, opened 2026-08-11
Add circuit-level logical error rate as a board metric


vprusso:
Every number on the board today comes from the parity check matrix alone: n, k, d, check weight, plus a layout when the submitter ships one. None of it says how the code performs when you actually run it, which is the number most people want when picking a code for hardware.

TRACKS.md already concedes the gap. The mitten entry in Reference bars says the paper evaluates those codes by logical error rate under noise, not by kd^2/n. So we are carrying codes whose own authors judged them on a metric we do not measure.

## Proposal

Add logical error rate under circuit-level noise as a per-code number.

This changes what a submission is. A row today is a code. Under this it becomes a code plus a syndrome extraction circuit plus a decoder plus a noise model. The same code with a better schedule earns a better number, which is the point: the circuit and the layout are part of the design, not packaging around it.

arXiv:2607.28795 runs a pipeline close to what we would need. They build syndrome extraction circuits for reconfigurable atom arrays with a greedy scheduler, decode with Relay-BP, and report logical error per round at 0.1% circuit noise. We could copy that shape.

## Naive first

The first versions do not need to be good. A dumb scheduler (measure every check in sequence, no parallelism) and stock BP+OSD would produce a defensible number, and a defensible number beats no number. The bar is reproducibility: fixed noise model, fixed decoder, fixed shot count, circuit shipped with the submission so anyone can rerun it.

If someone later submits the same code with a smarter schedule and gets a better rate, that is the system working.

## Open questions

1. Which noise model do we fix? Uniform depolarizing at p = 1e-3 is the obvious default. Idling error included, or idling-free as in arXiv:2607.27644?
2. Logical error per round, or per code cycle out to some target fidelity? Per round is easier to define and to compare across n.
3. Decoder: pin one so numbers are comparable, or let submitters choose and record which?
4. Where does it live? A separate track, or a column beside kd^2/n?
5. What compute budget can a submission cost at PR time? The distance gate already runs about ten minutes per code.

## Why this also helps the geometry question

A trapped ion or photonic proposal has no meaningful 2D layout, so the geometric score either ignores it or penalizes it for a plane it never claimed to live in. But it does have a circuit, and the circuit is where the connectivity constraints actually bite. Circuit-level numbers give those architectures somewhere to be measured.

Tagging @mathysrennela @FarLab @willzeng.

MathysRennela:
[...]

PR #680, opened 2026-08-20
Measured logical-error-rate tier for circuit entries (#460)

vprusso:
Implements the measurement half of #460 as a vertical slice on top of the merged circuit tier (#505/#646). The committed memory circuits and the canonical noise recipe are reused unchanged, so one artifact serves both the d_circ floor and the measured rate.

## Design, against the parameters agreed on the issue

- Per round: reported as `ler_per_round` via the parity-aware conversion from failures/shots and `circuit.rounds`.
- Confidence interval: `ci95` (Wilson) is a required field, and both it and the rate must recompute exactly from the raw counts or the claim fails.
- Uniform depolarizing at p = 1e-3: `p` is a schema constant equal to the circuit tier's `P_REF`.
- Decoder: one pinned decoder, `bposd-cs-10`, which is BP+OSD exactly as `decode/distance.py` already pins it, with the DEM's own per-mechanism probabilities as the channel prior. MWPM is deliberately not offered: matching needs a decomposable DEM, and weight-6+ codes produce mechanisms that do not decompose, the same fact that forced RFC 0001 rev 2 off `shortest_graphlike_error`. A second decoder can be added as an enum member later without breaking any stored claim.
- Ships with the submission and rerunnable: the block stores (shots, failures, seed); the stim sampler is seed-deterministic, so the sample is fully determined by the claim plus the pinned stim version.

One deliberate divergence from the issue thread: the noise model includes idle noise, because that is the canonical recipe the circuit tier already enforces mechanically. Two recipes would mean a circuit cannot carry between tiers; flagging this explicitly for review since idle-free was suggested on the issue.

## Verification is statistical, on purpose

`ler_verify.py` re-measures with an independent seed (capped at 100k shots) and rejects claims outside 4 sigma of sampling error. Bit-exact replication is not the contract: BP is float arithmetic and the circuit tier's byte-for-byte `.dem` comparison already had to retreat to tolerance over FMA drift on arm64 vs x86. Re-measurement is also exactly the test that catches the gaming direction, a claimed rate lower than the circuit earns. The false-alarm rate on an honest claim at 4 sigma is per-mille; anything large enough to change a ranking is far outside the band.

An entry with an `ler` block whose environment lacks the decoder fails verification rather than skipping it, so an unverifiable number cannot merge. The two `verify_all` CI steps and the self-tests gain `--extra research` (versions from the existing lock). The refutation-gate step deliberately does not: giving it ldpc would silently activate `gate_changed`'s second refutation mechanism, which should be its own decision.

## Not in this PR

- Site display and the separate-track presentation (follow-up; same pattern as the d_circ chip).
- A seed entry with a measured rate: submission-scope keeps code data out of a verifier PR, so it follows stacked, as #646's did.
- Coordination with #667: if the manifest expansion lands first, `ler_tools.py` and `ler_verify.py` belong in its TRUSTED tuple with a re-pin; whichever merges second reconciles.

## Verification

8 new tests (honest claim verifies; under-reported failures caught by replication; broken arithmetic, unpinned decoder, and sub-floor shots each caught by their check; same-seed determinism; conversion sanity). Full board sweep 358/358 unchanged, since no entry carries the block yet.

## What the budget admits today

With the failure-count floors and LER_SECONDS=120 per basis, the pinned
decoder's throughput decides which codes can carry the block at all
(numbers from FarLab's arm64 runs; x86 CI will differ by a small factor):

- `25-1-5` at 1.1 ms/shot: claim ~110 s/basis for ~140 failures, replica
  self-sizes to ~105k shots. Verifiable with margin.
- `49-1-7` at 16.8 ms/shot: 120 s buys ~3 expected replica failures.
  Unverifiable within budget; fails closed. Reaching 100 failures would
  cost ~67 min/basis.
- `81-1-9` at 132.7 ms/shot: ~0 expected replica failures in budget;
  ~30 h/basis to 100 failures. Fails closed.

So as merged the tier admits d=5 and nothing better. That is a stated
choice: a statistically meaningful gate inside a 15-minute CI budget.
Batching the per-shot decode loop (#702) is the ~10x that re-admits
d=7; raising LER_SECONDS is the fallback if batching stalls. Under-reports
up to the factor-2 admissibility bound can survive when the budget
truncates the replica; SCHEMA.md names that gray zone rather than
pretending the gate is sharper than it is.


schema/code.schema.json
```diff
@@ -135,6 +135,133 @@
         "notes": {
           "type": "string",
           "description": "schedule provenance, optional"
+        },
+        "ler": {
+          "type": "object",
+          "description": "optional measured logical-error-rate tier (issue #460): the same committed memory circuits, sampled under the canonical noise recipe and decoded with the pinned decoder. Distance is a floor; this is the measured rate a device-like simulation actually sees, prefactors included.",
```
MathysRennela: description field should probably not reference an issue

schema/SCHEMA.md
```diff
@@ -89,6 +89,30 @@ Two principles drive the format:
     Every TICK layer must be genuinely parallel (no qubit operated on twice
     in a layer): layer count controls idle-data noise, so this is what makes
     a schedule's claimed parallelism -- and the resulting d_circ -- honest.
+  - `ler` (optional; issue #460): the measured logical-error-rate tier, on
+    the same committed circuits. `d_circ` is a floor; this is the rate a
```
MathysRennela, line 93: referencing issue #460 is bad form imo

verify/ler_verify.py
```diff
@@ -0,0 +1,160 @@
+"""Verifier for measured logical-error-rate claims (issue #460).
```
MathysRennela: same as comments above about issue being mentioned

verify/verify_all.py
```diff
@@ -52,13 +53,27 @@
...
-            crep = verify_circuit(doc, os.path.join(code_root, "circuits", slug))
+            circuits_dir = os.path.join(code_root, "circuits", slug)
+            crep = verify_circuit(doc, circuits_dir)
             if crep["ok"]:
                 circ = (f", d_circ<="
                         f"{crep['earned_d_circ']['d_circ']['value']}")
             else:
                 rep["ok"] = False
                 rep["checks"] += [c for c in crep["checks"] if not c["ok"]]
+            # measured-rate tier (issue #460): an ler claim is re-measured,
```
MathysRennela: same as above

MathysRennela (commented):
overall good, but several files mention issue #460 which is probably not the best idea for code comments.

vprusso:
Addressed: all six issue-number references (module docstrings, the verify_all comment, the schema description, and SCHEMA.md) are gone; the wording is self-contained now. The remaining RFC 0001 mention in ler_tools.py stays deliberately, since it names the design fact (non-decomposable DEMs) that explains why MWPM is not the pin, not a tracker link. Say the word if you want that one out too.

FarLab (commented):
Checked out `feat/ler-tier` (fa3518d) and exercised the tier end-to-end on the one entry that already carries committed circuits, `[[25,1,5]]` (rotated surface code, 5 rounds). Summary of what I ran and what I think, in that order.

## What I ran

- `verify/test_ler.py`: 8/8 pass in 2.9 s (stim 1.16.0, ldpc 2.4.1, arm64).
- `ler_tools.ler_block` on `circuits/25-1-5/memory_{x,z}.stim` at 20k shots, seed 2026, then `ler_verify.py` on the result: `ok: true`, 42 s. Replica (seed `2026 ^ 0x5EED1E12`) got 27/20000 and 23/20000 against claims of 26 and 24. The doc validates against the new schema.
- Same block with `failures //= 5` and arithmetic kept consistent: both `*_ler_replicated` checks fail, `*_ler_arithmetic` pass, exit 1. The gaming direction is caught as designed.
- Out of scope for the PR but needed to judge the numbers: MWPM (pymatching 2.4.0) on the same circuits at p = 0.5e-3 / 1e-3 / 2e-3 with 400k shots, and the pinned BP+OSD at 1e-3 / 2e-3.

## Numbers

| p | MWPM, decomposed DEM | BP+OSD-CS-10 (pinned) |
|---|---|---|
| 0.5e-3 | 2.75e-4 /shot, 5.5e-5 /round | — |
| 1e-3 | 2.01e-3 /shot, 4.0e-4 /round | 1.30e-3 /shot, 2.6e-4 /round (X); 2.4e-4 (Z) |
| 2e-3 | 1.40e-2 /shot, 2.8e-3 /round | 9.85e-3 /shot, 2.0e-3 /round |

Log-log slopes per doubling of p: MWPM 2.6–2.9, BP+OSD 2.9.

These are consistent with what is known for d=5 under uniform circuit-level depolarizing noise:

- Absolute MWPM rate at p=1e-3 (4e-4/round) sits in the few-×1e-4 band the literature gives (Fowler-style estimate ~1.6e-4 with p_th≈0.57 %; stim-style all-noise circuits ~3–5e-4). This recipe (no 1q-gate noise, data-only idle) matches no published one exactly, so factor-2 agreement is the right bar and it is met.
- Slopes approaching (d+1)/2 = 3 from below are the d_circ = 5 signature; the rate is not just the right size, it scales like a distance-5 code.
- BP+OSD beats decomposed MWPM by 1.55×, the expected gain from decoding hyperedges rather than splitting them. Worth one sentence in SCHEMA.md: board `ler` values will read ~1.5× better than the MWPM numbers people know from papers, so nobody "refutes" an entry by running pymatching.
- X ≈ Z (26 vs 24; 805 vs 804 for MWPM), as a self-dual code with a hook-safe schedule should give.

So: implementation correct, physics right. My concerns are about statistical design.

## Concerns

**1. The shot floor is too low for the tier's stated purpose.** The PR motivates the tier by prefactor differences between schedules with equal d_circ. Those differences are typically 1.2–2×. At `MIN_SHOTS = 10000` and p_shot ~1e-3 an honest entry has ~13 failures (σ ≈ 28 %); at 20k I got 26 (CI spans a factor 2.1). Resolving a 30 % difference at 2σ needs ~200 failures, i.e. ~150k shots for this code and millions for a code at p_shot ~1e-4. As written the tier can certify an order of magnitude, not the comparison it exists to make.

**2. The 4σ replication gate passes a 2× under-report at the floor.** σ is computed under the claimed rate (correct direction), but with 13 expected failures a claim of 6 gives |diff| ≈ 7e-4 against 4σ ≈ 9.8e-4: it passes. At 20k shots a 2× under-report is borderline (7e-4 vs 7.2e-4). 5× is caught (I confirmed), 2× is not, and 2× reorders rankings.

**3. Verification is weakest for the best codes.** `REPLICA_SHOTS_CAP = 100000` means a code at p_shot = 1e-4 gets a replica of ~10 failures, so the gate only detects ≥3× misreports. The cap should scale with the claimed rate, not be a constant.

Suggested fix for 1–3 together: make the floors failure-count floors, e.g. require `failures >= 100` (equivalently `shots >= 100 / p_shot`) and size the replica to the same expected failure count. The counter-argument is cost: BP+OSD-CS-10 runs ~1.1 ms/shot on this 1687-column DEM, so the 100k replica is ~2 min/basis here and will be 10–50× that on an n≈400 code. A statistically meaningful tier and a 15-minute gate are in direct tension; the PR should say which one it is choosing rather than inherit both defaults.

**4. Smaller points.**
- `rounds = d` is the minimum at which a per-round number means anything; the literature uses rounds = 3d or extrapolates, and the time-boundary effect shifts per-round rates by 10–30 %. Same size as current statistical error, so harmless now, but it caps the precision the conversion can ever deliver.
- A single fixed p cannot separate prefactor from exponent. A second point at p = 2e-3 (7× higher rate, 7× cheaper per failure) would make each entry self-certifying the way the slope check above was. Optional; comparability argues for one p, and this can be a later enum addition.
- pymatching silently accepts the board's undecomposed DEM and drops every 3- and 4-detector mechanism (it kept 572 of 1687 edges here) while still returning numbers. That is a concrete reason the MWPM exclusion is right, and worth recording in the `ler_tools` docstring as a "do not do this" rather than only the decomposability argument.
- Measurement and replica ran in one process on one arm64 box, so the FMA-drift story is untested locally; the x86 CI run on the first real seed entry will be the first test.

None of this blocks the vertical slice; it is about what the tier can claim once entries start carrying the block. I'd merge with the failure-count floor (or at least a stated plan for it) and the SCHEMA.md sentence on BP+OSD vs MWPM.

vprusso:
All four numbered points addressed in the latest push, taking your suggested fix for 1-3 in exactly the form you framed it: the floors are failure counts now.

1. `MIN_FAILURES = 100` on every claim, enforced twice: `failures.minimum: 100` in the schema and the arithmetic check, whose message tells the submitter to raise shots until the count is reached. Claims carry <= ~10% sigma, so the 1.2-2x prefactor comparisons the tier exists for are resolvable.

2. The replica is sized to discriminate rather than fixed: it targets 150 expected failures under the claimed rate, which puts 4-sigma detection near 1.33x independent of the rate. Your 2x-at-the-floor case is now the regression test (`test_underreported_failures_rejected` tampers by exactly 2x with consistent arithmetic above the floor, and the sized replica catches it).

3. Same mechanism covers the best-code case, since the replica scales as 1/p_claim. And the tension you asked the PR to resolve explicitly is resolved explicitly: `LER_SECONDS = 120` per basis bounds the gate (deadline checked every 1000 shots), and when truncation leaves the replica unable to catch even a 2x under-report, the claim fails as "unverifiable within budget" rather than merging weakly checked. The statistical meaning of the tier wins; the budget bounds what may merge, never how honestly it is checked. Stated in SCHEMA.md and pinned by a monkeypatched-budget test.

4. The pymatching silent-truncation warning is in the `ler_tools` docstring with your measured 572-of-1687 number, the ~1.5x BP+OSD-vs-decomposed-MWPM offset sentence is in SCHEMA.md so nobody refutes an entry with pymatching, and the rounds = d boundary-bias caveat is recorded next to the conversion. The second p point I agree belongs as a later enum addition; comparability wins for v1.

End-to-end on the merged `[[25,1,5]]` at 100k shots/basis with the new floors: X 127 failures (2.54e-4/round), Z 121 (2.42e-4/round), which agrees with your independent arm64 measurement (2.6e-4/2.4e-4) inside one sigma, so the cross-platform statistical story has now survived its first x86-vs-arm test. Verification passes at ~120s/basis: the budget truncates the replicas to 84k/80k shots, which still detect 1.63x/1.68x under-reports, comfortably inside the 2x admission bound.

On your cost observation, made concrete by the run above: the honest floor-compliant claim costs the gate ~4 minutes both bases on this 1687-mechanism DEM, and the unverifiable-within-budget failure is doing real work as the tier's admission rule for larger codes until the decoder path gets faster. That is the stated trade, not an accident.

FarLab (commented):
Re-verified at ea2900cd (arm64): 11/11 tests; honest 100k-shot claim on `[[25,1,5]]` verifies (X 147 / Z 138 failures, replica self-sizes to ~105k shots, 1.48x detection, ~110 s/basis); a genuine 2x under-report (true failures, doubled claimed shots) is caught on both bases with margin. All four points from my review are addressed for real.

One finding worth stating in the PR text: the budget excludes the d=7/d=9 seeds just merged in #682. Pinned decode runs 16.8 ms/shot on `49-1-7` and 132.7 ms/shot on `81-1-9`, so 120 s buys ~3 and ~0 expected replica failures — both unverifiable within budget — and the claim itself would cost ~67 min/basis (d=7) / ~30 h/basis (d=9) to reach 100 failures. The rule fails closed, which is right, but as merged the tier admits d=5 and nothing better; worth saying with numbers, plus an issue for batching the per-shot decode loop (~10x there re-admits d=7).

Two small fixes:
1. Sigma omits the claim's own variance: for a floor claim vs a 150-failure replica the "4 sigma" gate is ~2.5 sigma of the actual difference statistic, i.e. ~1% false-alarm per basis, not per-mille. One line: `sigma^2 = p(1-p) * (1/shots + 1/rep_done)`.
2. The `REPLICA_FAILURES` comment says ~1.33x but the formula prints 1.49x; pick one. And SCHEMA.md should say plainly that under-reports up to 2x can survive when the budget truncates (I measured a 1.47x tamper passing a 1.59x threshold — within spec, but name the gray zone).

Nit: dead `-(-lv.REPLICA_FAILURES // 1) and` in `test_replica_sized_to_discriminate`.

With the sigma fix and the admissibility numbers stated, mergeable as far as I'm concerned.

vprusso:
All addressed in the latest push, and the sigma point was a genuine error on my side, thank you for the arithmetic.

1. Sigma now carries both binomial terms: `sigma^2 = p(1-p)(1/shots + 1/rep_done)`, so the nominal 4 sigma is 4 sigma of the actual difference statistic and the per-mille false-alarm claim is true again. The correction had a consequence your review predicted in spirit: with the wider sigma, mean-crossing admissibility let a truncated replica squeak under the factor-2 bound while catching a boundary tamper at coin-flip odds (my own truncation test caught it). Admissibility therefore now demands power, not mean-crossing: `(f-1)p >= Z_GATE sigma + POWER_SIGMA sqrt(f) sigma` at f = 2, so the boundary under-report is caught at ~98% or the claim is unverifiable within budget.

2. The comment now derives from the corrected formula instead of contradicting it, the printed factor is documented as the 50%-power point, and SCHEMA.md names the gray zone between that point and the enforced bound, with the note that admissibility's power requirement is what bounds the zone. Your measured 1.47x-past-1.59x case is exactly the zone as now documented.

3. Dead expression gone.

4. The admissibility numbers are stated in SCHEMA.md (d = 5 in budget; d = 7 and d = 9 fail closed at the pinned decoder's speed), and the batching issue is #702 with your per-shot measurements as the baseline and "d = 7 verifies within the existing 120 s budget on the merged 49-1-7 seed" as the acceptance line.

Re-verified end to end on `[[25,1,5]]` at 100k shots/basis under the corrected statistics: X 127 / Z 121 failures, replicas truncated to 82k shots by the budget, detection at 1.53x/1.54x (was 1.63x/1.68x under the old sigma: honest widening), both bases pass in ~120 s each, and the power-based admissibility bound holds with margin. 11/11 tests.

vprusso:
Round-3 items are on the branch at 425098c: combined-variance sigma, power-based admissibility with the 50%-power point documented, the 1.4x comment reconciled with the formula, the gray zone named in SCHEMA.md, dead expression removed. The budget-admission numbers (d=5 in, d=7/d=9 fail closed, with per-shot costs) are now in the PR body, and the batching work that re-admits d=7 is filed as #702.

vprusso:
Added a decode-cost benchmark (`verify/bench_ler.py`) so the budget question from the review is measurable rather than discovered by a claim failing closed. It reports, per code and basis, setup cost, marginal ms/shot, the shots `LER_SECONDS` buys, and the per-shot failure rate a code must exceed to reach `MIN_FAILURES`.

Numbers from an x86 box (12 cores, one unrelated 4-thread job running, so read these as a loaded-machine floor rather than a clean best case):

| code | basis | DEM mech | ms/shot | shots in budget | p needed |
|---|---|---|---|---|---|
| 25-1-5 | X | 1687 | 1.22 | 97891 | 1.0e-3 |
| 25-1-5 | Z | 1687 | 2.81 | 42746 | 2.3e-3 |
| 49-1-7 | X | 5473 | 37.2 | 3227 | 3.1e-2 |
| 49-1-7 | Z | 5473 | 50.9 | 2358 | 4.2e-2 |
| 81-1-9 | X | 12691 | 326 | 353 | 2.8e-1 |
| 81-1-9 | Z | 12691 | 302 | 396 | 2.5e-1 |

This confirms the d=7 and d=9 conclusion with a second measurement, but it also says something the review has not yet: 25-1-5 is not comfortably admissible, it is marginal, and which side of the line it lands on depends on the machine. Its measured rate is about 1.3e-3 per shot on X and 1.2e-3 on Z. Against the numbers above that clears X by roughly 1.25x and misses Z by roughly 2x, whereas on the arm64 box in the earlier review pass the same code verified on both bases at about 1.1 ms/shot.

So as written the tier admits d=5 on fast hardware and nothing anywhere, on slow hardware. That is still fail-closed and still sound, but a submitter would see an unexplained refusal that depends on the runner. Three options, in the order I would take them: land the batching in #702, which moves everything at once; or raise `LER_SECONDS` enough that the d=5 class clears on a slow runner with margin; or state the throughput requirement in SCHEMA.md so the refusal is legible. I would not tighten `MIN_FAILURES` to paper over it, since the failure floor is what makes the tier able to compare prefactors at all.

The benchmark is deliberately a reporting tool and not a test, since a wall-clock assertion would be exactly the machine-dependent flake we hit in #684.

MathysRennela (approved):

PR #684, opened 2026-08-21
gf2_fast: in-C++ DEM trial loop for the circuit-tier search (10–18×)

FarLab:
Follow-up to #663/#682: ports the circuit tier's RIS trial loop into `gf2_fast`, removing the per-trial overhead that made the DEM search 10–100× more expensive than it needed to be.

## What was wrong

`ris_dem` orchestrated trials from numpy: each trial materialized an m×m column permutation (~130 MB allocation + cache-hostile gather at m≈12k) and re-packed it into bitsets before any elimination ran — and the loop was single-threaded. Invisible at code-tier sizes (kernels ≤ 0.5 MB); dominant at DEM widths.

## Why the port is small

The machinery already existed: `min_logical_weight_rand_core`'s CSS nontriviality test (odd inner product against any row of a basis) **is** the circuit tier's `L·e ≠ 0` predicate verbatim, and `rref_perm` already randomizes the information set by pivot *order* with zero data movement. `dem_rand_witness` packs `ker(H_dem)` once, fans out across 4 threads, and returns (weight, mechanism indices). `circuit_tools.ris_dem` consumes it in fixed 64-trial chunks so the wall cap still binds; deterministic given (seed, trials, threads).

Trust layering unchanged: the C++ side proposes, `witness_errors` validates, and the numpy loop remains the audited reference and fallback — with a parity test asserting both paths find the same hook-limited bound (3) on the greedy [[25,1,5]] DEM with valid witnesses. The gate's budget flag keys on the new symbol specifically, so a stale extension is budgeted as the fallback it would actually run.

## Measured effect (same machine as the #663 calibration)

| m | before | after | |
|---|---|---|---|
| 1,687 (d=5 seed) | 8.4 ms/trial | 0.76 ms | 120 s buys 20,000 trials (was 2,000) |
| 5,353 (d=7) | ~197 ms | 10.8 ms | ~11,000 trials (was 609) |
| 12,411 (d=9) | ~2.5 s | 330 ms | ~360 trials (was 48) |
| 25,000 (tier cap) | ~19 s (extrap.) | ~3.1 s | ~38 trials — the cap is now genuinely searchable |

Budget constants re-fit (`2e-13·m³` C++, `1.9e-11` numpy fallback → factor 100), trial ceiling 2,000 → 20,000. `MAX_DEM_MECHANISMS` stays 25,000 — raising it is a submission-surface decision for its own PR, now that the search could afford it. Verified end-to-end: the gate re-searched the merged 25-1-5 artifact at 20,000 trials/basis (~17 s each), nothing below the claimed 5/5.

Suite: 68 passed.

### DIFF
--- a/verify/circuit_tools.py
+++ b/verify/circuit_tools.py
@@ -332,11 +332,18 @@ def ris_dem(H, L, trials, seed=0, pair_top=24, max_seconds=None, threads=4):
 
     When gf2_fast provides dem_rand_witness, the whole trial loop runs in
     C++ (kernel packed once, pivot-order permutation instead of physical
-    column moves, `threads` workers) in fixed 64-trial chunks so the wall
-    cap is still enforced between chunks; results are deterministic given
-    (seed, trials, threads) whenever the cap does not truncate. Finds are
-    PROPOSALS either way -- callers validate with witness_errors. The numpy
-    loop below is the fallback and the audited reference implementation."""
+    column moves, `threads` workers), handed out in chunks. A chunk cannot
+    be aborted, so the wall cap is enforced by PROJECTION, not just by
+    checking between chunks (vprusso's #684 review: a blind 64-trial chunk
+    overshot a 0.5 s cap by ~140x at m~12k): the loop opens with a small
+    8-trial chunk to measure this machine's per-trial cost, then sizes every
+    later chunk to what the remaining budget can pay for (floor `threads`,
+    else stop), so the worst overshoot is one small chunk rather than one
+    arbitrarily expensive one. An unconstrained run always uses the fixed
+    (8, 64, 64, ...) sequence, so results stay deterministic given (seed,
+    trials, threads) whenever the cap does not truncate. Finds are PROPOSALS
+    either way -- callers validate with witness_errors. The numpy loop below
+    is the fallback and the audited reference implementation."""
     import time
     H = np.ascontiguousarray(np.asarray(H, dtype=np.int8))
     L = np.ascontiguousarray(np.asarray(L, dtype=np.int8))
@@ -344,12 +351,21 @@ def ris_dem(H, L, trials, seed=0, pair_top=24, max_seconds=None, threads=4):
         deadline = (time.monotonic() + max_seconds) if max_seconds else None
         best, wit = None, None
         done, chunk_i = 0, 0
+        per_trial = None            # measured; conservative running max
         while done < trials:
-            t = min(64, trials - done)
+            t = min(8 if chunk_i == 0 else 64, trials - done)
+            if deadline is not None and per_trial is not None:
+                afford = int((deadline - time.monotonic()) / per_trial)
+                if afford < max(threads, 1):
+                    break
+                t = min(t, afford)
+            t0 = time.monotonic()
             w, sup = _GF.dem_rand_witness(H, L, trials=t,
                                           seed=seed + 7919 * chunk_i,
                                           pair_depth=pair_top,
                                           threads=threads)
+            per_trial = max(per_trial or 0.0,
+                            (time.monotonic() - t0) / t, 1e-9)
             if w is not None and (best is None or w < best):
                 best, wit = int(w), [int(i) for i in sup]
             done += t
--- a/verify/circuit_tools.py
+++ b/verify/circuit_tools.py
@@ -322,15 +322,42 @@ def _rref(M):
     return gf2.rref(np.ascontiguousarray(M))[0]
 
 
-def ris_dem(H, L, trials, seed=0, pair_top=24, max_seconds=None):
+def ris_dem(H, L, trials, seed=0, pair_top=24, max_seconds=None, threads=4):
     """RIS upper-bound search for the lightest undetected logical fault set:
     min |e| with H e = 0, L e != 0. The code-tier searcher on the DEM's
     parity-check matrix -- hyperedge degree is irrelevant to it. Returns
     (weight, sorted column indices) or (None, None). Stops after `trials`
     permutations or `max_seconds` wall-clock, whichever comes first (the time
-    cap keeps the CI gate bounded when the size estimate is off)."""
+    cap keeps the CI gate bounded when the size estimate is off).
+
+    When gf2_fast provides dem_rand_witness, the whole trial loop runs in
+    C++ (kernel packed once, pivot-order permutation instead of physical
+    column moves, `threads` workers) in fixed 64-trial chunks so the wall
+    cap is still enforced between chunks; results are deterministic given
+    (seed, trials, threads) whenever the cap does not truncate. Finds are
+    PROPOSALS either way -- callers validate with witness_errors. The numpy
+    loop below is the fallback and the audited reference implementation."""
     import time
-    K = _kernel_basis(np.asarray(H, dtype=np.int8))
+    H = np.ascontiguousarray(np.asarray(H, dtype=np.int8))
+    L = np.ascontiguousarray(np.asarray(L, dtype=np.int8))
+    if _GF is not None and hasattr(_GF, "dem_rand_witness"):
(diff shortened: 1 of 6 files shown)


vprusso (changes requested):
Built the extension and verified the claims independently rather than from the table. On the merged `25-1-5` artifact (m=1687): C++ path 1.13 ms/trial, numpy fallback 55 ms/trial, a 49x per-trial speedup on this machine; both paths find w=5 with witnesses that pass `witness_errors`; two identical calls return identical results, so the determinism claim holds as stated. The parity test passes, the gate still validates every proposal through `witness_errors` before counting a refutation, and the reuse argument is genuinely right: `L·e != 0` is the CSS nontriviality predicate with L in the basis seat, so the trials core carries over without modification. My constants differ from yours (you measured 0.76 ms and a 100x factor; I get 1.13 ms and 49x under load), which is expected across machines and is exactly why the wall cap has to be the real safety net rather than the fit.

Which is the problem, and it is the one thing I would not merge as-is.

## The wall cap no longer binds where it matters

The chunked loop only checks the deadline BETWEEN 64-trial chunks, and the chunk runs to completion in C++ with no abort. Measured on the merged `81-1-9` DEM (m=12,691):

```
max_seconds=0.5 requested; actual wall 70.2s  (overshoot 140x)
```

The comment this PR keeps says the cap "bounds CI even where the fit is off", and that property is now false at exactly the sizes it was written for. At `MAX_DEM_MECHANISMS` a single chunk is ~3.3 minutes at your fit-correct cost, and if the fit is off by the same 5x my machine already disagrees with yours, one chunk is ~17 minutes with no abort. The pre-port loop checked per trial, so the cap degraded gracefully; this one cliff-edges.

The fix does not need to touch the C++. In the Python loop, track elapsed-per-trial from completed chunks and project before launching: if `done > 0` and `est_per_trial * next_chunk > remaining`, shrink the chunk to fit (floor at `threads`) or stop. To bound the FIRST chunk, which projection cannot help, open with a small fixed chunk (say 8 trials) before the 64s. A fixed chunk-size sequence keeps the uncapped run deterministic, which is the property you were protecting with fixed 64s; capped runs lose tail determinism, but your docstring already scopes determinism to "whenever the cap does not truncate". With that change the overshoot is bounded by one small chunk instead of one arbitrary one.

A regression test in the same style as the demo above would pin it: request a sub-second cap on the 81-1-9 DEM and assert wall time within a small multiple of it.

## Coordination, not blockers

- #667 (still open) also rewrites the `gf2_fast.cpp` header comment, so these two PRs conflict textually on those lines; whoever merges second resolves a trivial hunk. If #667 lands first, this PR also needs a manifest re-pin, since #667 pins `gf2_fast.cpp`.
- `CIRCUIT_PY_FACTOR = 100` vs my measured 49x means fallback budgets under-provision trials on a machine like mine. That errs shallow-and-warned, which is the safe direction, so leave it; just noting the constant is machine-specific.

Everything else I would take as-is: the header correction now matches what CI actually does, the budget key on `hasattr(_GF, "dem_rand_witness")` correctly treats a stale extension as the fallback it would actually run, and hoisting the kernel packing out of the trial loop is the whole win.

FarLab:
Fixed as you prescribed, in 15fb619's follow-up commit. Reproduced first: your 0.5 s request on the m=12411 DEM ran 21.6 s on my machine (43×; your 140× is the same bug at your trial cost) — confirming the cap had stopped being the thing that bounds CI when the fit is wrong.

The loop now opens with a fixed 8-trial chunk to measure the actual per-trial cost on the machine it's running on (kept as a conservative running max, which also absorbs the per-chunk kernel recompute), then sizes each subsequent chunk to what the remaining budget affords, flooring at `threads` and stopping below that. No C++ changes. Measured after: the same 0.5 s request completes in 2.8 s — one small opener chunk, as designed — and unconstrained runs keep the fixed (8, 64, …) sequence, so determinism is unchanged exactly where the docstring scopes it.

Your regression test is in as `test_dem_wall_cap_binds_on_large_dem`: sub-second cap on the 81-1-9-sized DEM, wall asserted `< 15 s` (generous for slow CI hardware; the pre-fix behavior was 21–70 s, so the bound still discriminates).

On coordination: #667 is still open, so I've left the header hunk as-is — happy to rebase whichever lands second, and I'll re-pin the manifest here if #667 goes first. Agreed on leaving `CIRCUIT_PY_FACTOR` alone; shallow-and-warned is the right failure direction for a machine-specific constant.

vprusso (changes requested):
The fix is correct and I verified it against the head: my original demo drops from 70.2 s to 12.7 s for a 0.5 s cap on the m=12,691 DEM, the uncapped (8, 64, ...) sequence is deterministic across identical calls, and the conservative running-max on per-trial cost is the right way to absorb the kernel recompute. The residual overshoot is one opener chunk, exactly as designed.

The regression test, though, fails on my machine: `wall 17.7s for a 0.5s cap` against the 15 s bound, with the projection working exactly as intended. The overshoot is bounded by one 8-trial opener chunk, and the opener's cost is machine- and load-dependent, so ANY absolute-seconds assertion is asserting the speed of the box, not the correctness of the loop. My box is faster than a CI runner and still tripped it under load; on shared CI hardware this will flake, and a flaky trust-path test ends up skipped or deleted, which un-pins the regression.

Assert the structure instead of the clock. The property the fix guarantees is "when the cap is already blown, the loop stops after the opener", and that is testable with a stub, deterministically, in milliseconds:

```python
def test_dem_wall_cap_stops_after_opener(monkeypatch):
    """#684 review regression, load-independent form: a chunk cannot be
    aborted, so when the opener reveals the cap is already unaffordable the
    loop must stop after that single small chunk -- no second chunk, ever."""
    calls = []

    def stub(H, L, trials, seed, pair_depth, threads):
        calls.append(trials)
        time.sleep(0.05)          # any nonzero cost blows a 1 ms cap
        return None, None

    fake = types.SimpleNamespace(dem_rand_witness=stub)
    monkeypatch.setattr(ct, "_GF", fake)
    H = np.zeros((2, 8), dtype=np.int8); H[0, 0] = H[1, 1] = 1
    L = np.zeros((1, 8), dtype=np.int8); L[0, 7] = 1
    ct.ris_dem(H, L, trials=20000, seed=3, max_seconds=0.001)
    assert calls == [8], f"expected only the opener chunk, got {calls}"
```

That pins the exact bug (the pre-fix loop would have queued a 64-trial second chunk) with no DEM build, no extension, and no dependence on what else the runner is doing. Keep the big-DEM run as the demo in the PR description if you like; it just should not be the assertion. The existing test's 23 s runtime also goes away, which the self-test suite will appreciate.

Everything else is done as far as I am concerned; this is the one change between here and approval.

FarLab:
Right on both counts — the absolute-seconds bound was asserting the box, not the loop, and your 17.7 s false failure with correct behavior is the proof. Swapped in your structural test essentially verbatim as `test_dem_wall_cap_stops_after_opener` (moved to `test_circuit_gate.py` so it runs even where the extension isn't built — it needs neither the extension nor a DEM). Verified it fails against the pre-fix loop (`calls == [64]`) and passes against head in milliseconds. The big-DEM timing run stays in the PR description as a demo only, and the suite sheds the ~20 s the old test cost.

vprusso (approved):
Verified on the head rather than trusted: the structural test passes in the suite without the extension built, and against the pre-fix loop it fails with exactly the right shape (`assert [64] == [8]`), which is the discriminating property the wall-clock version could not give. Moving it to `test_circuit_gate.py` so it runs extension-less was a better call than my sketch made, since the projection logic is pure Python and deserves coverage everywhere.

The cap-honoring loop, the demo staying descriptive rather than asserted, and the earlier verified items (49x-and-up speedup, determinism, witness validity through `witness_errors`, budget keyed on the DEM entry point) close everything I raised across the three rounds. The remaining #667 header-hunk coordination is bookkeeping for whoever merges second, as already agreed.

Approving.

PR #691, opened 2026-08-23
Fix leaderboard header layout: keep "best g" card in place (#666)

### LINKED ISSUE #666: Layout issue on desktop for the leaderboard
One of the recent changes introduced a layout issue on the board (see screenshot).
The issue is that the "best g" box is misplaced. (Likely cause: the explanation text "7 of 13 contributors have a verified layout and d ≥ 3 (their own code, or a layout they contributed)" is too long.)

Image

MathysRennela:
## Summary
Fixes the desktop layout issue on the leaderboard where the "best g" hero card was pushed out of place (issue #666).

**Root cause:** `.lbsub` contains two inline spans, and the long geo sentence ("7 of 13 contributors have a verified layout and d ≥ 3 …") made the title block a very wide flex item inside `.lbhead`, squeezing the "best g" card onto a second row.

**Fix:** cap `.lbsub` with `max-width:min(46ch,100%)`, `flex:1 1 auto`, `min-width:0`, so the sentence wraps within the title block instead of stretching it. The header stays on one row in both metric modes.

Closes #666

## Changes
- `site/build.py` — CSS rule for `.lbsub`
- `docs/style.css` — regenerated output

## Validation
- Ran `python site/build.py` — regenerated `docs/index.html` + 365 detail pages + `docs/style.css`; confirmed the new `.lbsub` rule landed in the output CSS.

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -93,9 +93,16 @@ padding:16px 20px;background:var(--soft);border-bottom:1px solid var(--ln)}
 .lbh{font-size:16px;margin:0;font-family:'Space Mono',ui-monospace,monospace;
 text-transform:uppercase;letter-spacing:.03em}
 .lbsub{margin:4px 0 0;font-size:13px;color:var(--mut);
-/* cap the title block so the long geo sentence can't stretch it and push the
-   hero cards ("best g") onto a second row on desktop */
-max-width:min(46ch,100%);flex:1 1 auto;min-width:0}
+/* cap the title block so the long geo sentence can't stretch it (issue #666).
+   A plain length, deliberately: a percentage arm (min(38ch,100%)) is a cyclic
+   reference during intrinsic sizing, so the whole max-width is IGNORED for
+   the flex line-breaking decision and the row sees the unwrapped sentence --
+   which is exactly the bug. 38ch + the .lbscore cap clear the row at 1024px;
+   below 38ch of viewport the block just fills the container. */
+max-width:38ch}
+/* the title wrapper is the actual flex item; without min-width:0 its
+   min-content locks the row even when .lbsub wraps */
+.lbhead>div{min-width:0}
 .lbcta{flex:0 0 auto;font-size:13px;font-weight:700;color:#fff;
 font-family:'Space Mono',ui-monospace,monospace;
 background:var(--ac);border:none;border-radius:8px;padding:8px 14px;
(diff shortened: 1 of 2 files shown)


vprusso (approved):

vprusso (changes requested):
First, the regression is mine: the long geo sentence is the #656 reword, so thanks for picking it up. I rendered the fix rather than reading it, though, and the measurements say it does not repair the reported layout, so I have to hold it.

## What I measured

Real board markup + real stylesheet (the built `docs/index.html` leaderboard section), section forced to `data-mode=geo` (where the long sentence and the best-g card actually coexist; in eff mode `.lbsgeo` is `display:none`, so the sentence cannot stretch anything there), headless Chromium at fixed widths:

| viewport | main CSS | this PR's CSS |
|---|---|---|
| 1058 (the issue's screenshot width) | card wrapped below title | card still wrapped; sentence now wraps to 2 lines |
| 1280 | wrapped | still wrapped |
| 1330 | one row | one row |
| 1440 | one row | one row |

The cap does what it says (the sentence wraps instead of stretching the title block), but the best-g card's placement is identical to main at every width I sampled. The row's binding constraint is not the sentence: it is the rigid content, dominated by the hero card's one-line subtitle (`[[656,114,3]] · d≤3 · r=1.4142 · ρ=1 · @npdeep`, ~330px of nowrap) plus the weight slider and toggle. Below ~1300px those cannot fit beside any title block, capped or not.

## What would fix 1058

Give the card a compressible subtitle rather than (or in addition to) capping the title: let `.lbsd` wrap, or cap `.lbscore`'s width so its min-content stops dictating the row. With the card compressible to ~180px the row fits at 1024 with margin. Whatever the chosen lever, please verify at 1058 before merging, since that is the width the issue documents.

## Smaller things

- `flex:1 1 auto` and `min-width:0` on `.lbsub` are no-ops: `.lbsub` is not a flex item (its anonymous parent `<div>` is). Either drop them or put `min-width:0` on the parent, where it would actually allow the title block to shrink.
- A reproduction gotcha that cost me an hour and will cost the next reviewer the same: the leaderboard's toggle script re-initializes `data-mode` on load, so a static render of the built page always snaps back to eff mode. Strip the scripts (or drive the toggle) before screenshotting, otherwise you are always looking at the mode where this bug cannot occur, and any CSS change here "verifies" trivially.
- The PR commits `docs/style.css` but not the rest of the regeneration it describes; convention elsewhere is to leave `docs/` to the site bot entirely. Suggest dropping that hunk.

Happy to re-run the width sweep on the next revision; the harness is sitting here ready.

vprusso:
Pushed the completing fix onto this branch rather than bouncing it back again, since the root cause turned out to be worth a write-up and the diff is small.

Why 46ch (and my suggested 34ch, and any `min(Xch,100%)`) could never move the card: the percentage arm of `min()` is a cyclic reference during intrinsic sizing, so the whole `max-width` is ignored exactly when the wrapping flex line is broken. The row sees the unwrapped ~590px sentence, evicts the hero, and only then applies the cap to the laid-out block. Measured directly: the title flex item was 593px wide holding 283px of visible content. That is also why the fix looked verified in the inspector, where the capped width is what you see, while the layout stayed broken.

Second, independent constraint: the hero card's one-line subtitle gives it a ~330px hypothetical size, and in a wrapping flex row line-breaking happens at hypothetical sizes before flex-shrink ever runs, so the card forced its own eviction even with the title fully capped (this is what my review's width table was showing).

The push therefore makes both caps plain lengths: `.lbsub{max-width:38ch}` and `.lbscore{max-width:260px}`, plus `min-width:0` on the title wrapper (the actual flex item) and removal of the inert flex declarations. Verified by headless renders at 1024/1080/1280/1440 in both metric modes: one row everywhere, sentence and card subtitle wrapping internally. `docs/style.css` is reverted to main per the site-bot convention.

@MathysRennela if the shape looks right to you, this should close #666 for the widths the issue reported, not just for wide desktops.

FarLab:
It renders correctly for me:

Scherm­afbeelding 2026-08-24 om 16 01 27

FarLab (approved):

vprusso (approved):

PR #700, opened 2026-08-24
submit: make the drafted PR body pass the prose gate

MathysRennela:
## Problem

Every `qldpc submit --open-pr` submission failed CI's `prose` check on arrival. The cause is structural, not agent sloppiness: `pr_body()` in `cli/qldpc.py` deterministically generated a body that violates four rules of `verify/check_prose.py`:

1. a footer instructing the contributor to edit the draft (matches the checker's unedited-draft-footer pattern);
2. HTML-comment placeholders around the frontier section and research-note pointer;
3. a hardcoded unticked checklist box about equivalence to existing entries;
4. another unticked box whenever no construction argument was given.

So a PR opened by the automated path could not pass until its body was hand-rewritten via `gh pr edit`.

## Changes (`cli/qldpc.py` only)

1. **Clean body by default.** `pr_body()` now emits only checker-accepted content:
   - placeholder comments become plain parenthetical prompts (e.g. "(Name the track and the existing entry this beats or extends...)");
   - checklist boxes are emitted ticked — the construction box only appears when a construction was actually given;
   - the draft footer is gone; the "review before requesting review" reminder lives in the CLI output instead.

2. **Pre-flight prose gate in `open_pr()`.** Before any git command touches the working tree, the tool runs `verify/check_prose.py` (from the repo's own tree) against the drafted body and the staged files. A failing body now stops with exit 1 and the checker's own diagnostics — no branch, no push, no PR that fails CI.

## Validation

- Generated-body test: output of `pr_body()` passes `verify/check_prose.py --body-file` with exit 0; old-style footer text is still detected as scaffolding.
- `uv run pytest verify/test_frontier_summary.py verify/test_check_prose.py` — 2 passed.
- `uv run ruff check cli/qldpc.py`: 21 findings, identical to the pre-change baseline (all pre-existing; none introduced).

No changes to `verify/`, so no validator re-pin is needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,7 +75,10 @@ Two more, same reason:
   not put a distance in the filename that the JSON does not support.
 - Delete the drafting scaffolding before asking for review: the `qldpc submit` footer, HTML
   comments, unticked checklist boxes, session URLs. If a checklist box is not true, make it
-  true or say why.
+  true or say why. The drafted body also carries parenthetical prompts (e.g. "(Name the
+  track and the existing entry this beats…)"): replace each with real content. Then run
+  `uv run python verify/check_prose.py --body-file <body.md> --files <changed .md files>`
+  locally; it must exit 0 before you request review.
 
 A note named `<n>-<k>-<d>.md` must state its own `[[n,k,d]]` first. Follow
 `notes/TEMPLATE.md`; its sections are what a later searcher reads.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -164,11 +164,18 @@ codes. Goal: find a CSS qLDPC code that advances a frontier, and submit it.
 
 5. Submit: ./qldpc submit yourcode.npz --authors @yourhandle --family <family>
    --model "<exact model version, e.g. Claude Opus 4.8>". It finds the witness,
-   runs the verifier (which computes the locality and weight classes), and opens
-   the PR. CI re-verifies. The PR body is drafted from the submission,
-   including a computed "what frontier does this advance?" section (the track
-   cells the code lands in and the existing entries it dominates). Review it
-   before the PR is ready for review.
+   runs the verifier (which computes the locality and weight classes), runs
+   verify/check_prose.py on the drafted PR body before opening anything, and
+   only then opens the PR. CI re-verifies. The PR body is drafted from the
+   submission, including a computed "what frontier does this advance?"
+   section. Before requesting review you MUST finish it by hand:
+     - replace every parenthetical prompt in the body with real content;
+     - tick each checklist box only once it is actually true — the
+       "equivalent to an existing entry" box needs a provenance.notes entry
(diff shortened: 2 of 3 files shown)


FarLab:
Thanks — the diagnosis is right, and the pre-flight gate is the part I like most: running `check_prose.py` itself before any git command means the tool can never drift out of sync with CI again, and a failing body dies locally with the checker's own diagnostics instead of a red PR.

One thing I don't like: the prose gate existed partly to force hand-editing of drafts, and this makes an unedited draft pass it — the parenthetical prompts ("review and edit", "Name the track…") now ship as visible text in a green PR, and the equivalence-note checklist item is dropped rather than fixed.

*Drafted with Claude (Fable 5).*

MathysRennela:
Do you have a suggestion to resolve that? It's difficult to pass the prose gate if it is too tight

FarLab:
In my experience with submitting codes, I haven't had any prose gate fails. So could it be that the instructions to the LLM agent is not precise enough?

MathysRennela:
Yes. On my side, the first prose check consistently fails, regardless of the model I use.

FarLab:
Suppose you added to your research prompt to make sure the prose gate succeeds, would that help?

MathysRennela:
Fair on both counts — pushed `9a49955d` to address this.

**You're right that tick-by-default was wrong.** The equivalence box is restored, unticked by design: judging equivalence to an existing entry genuinely needs human eyes, so an unedited draft is *deliberately* not ready for review, and the prose gate now enforces exactly that. I verified the new behavior end-to-end:

- unedited draft → `check_prose.py` exits 1 ("unticked checklist box")
- draft with the box made true (and prompts replaced) → exits 0

So the gate keeps its teeth; what changed is only that the failure surfaces **before** the PR is opened (the pre-flight gate in `open_pr()` runs the checker locally and aborts), rather than as a red CI run after.

**On the prompt:** agreed that's the other half. Updated both instruction surfaces:
- `CONTRIBUTING.md` step 5 (the "Contribute with an LLM" prompt) now spells out the required hand-edit concretely — replace every parenthetical prompt with real content, make each checklist box actually true (equivalence needs a `provenance.notes` entry or a deliberate check), add the note if missing — and requires re-running `verify/check_prose.py --body-file ... --files` until it exits 0 before requesting review.
- `AGENTS.md` "Writing the submission" got the same guidance, since agents read that first.

To your experience question: hand-edited submissions never fail because a human naturally does all of the above. The failures I hit were all from `--open-pr` shipping the tool's draft verbatim — the deterministic footer/comments/unticked-box content the generator itself emits. The combination (generator no longer emits rejectable scaffolding *except* the intentional equivalence box + prompt tells the agent exactly what to finish + pre-flight gate catches anything left) closes that path while keeping the review-forcing behavior you wanted.

FarLab:
LGTM

vprusso (approved):

PR #701, opened 2026-08-24
Fail verification on dishonest layouts instead of silent demotion

### LINKED ISSUE #661: Fail verification when locality is present but earns no class
Follow-up from @FarLab's review of #655 (pre-existing behavior, not introduced there): the verifier only hard-fails a layout on coordinate count, schema, and `interaction_radius_within_claim` (when claimed). A layout that violates spacing >= 1, qubits-per-site <= layers, or both class caps is silently demoted to `unrestricted` — it merges green and earns nothing, with no signal to the submitter.

For the mass layout-filling effort this is a footgun: a botched batch entry looks accepted while contributing no locality class and a meaningless g. Proposal: when `locality` is present but the honest-layout conditions fail (spacing violation, site over-occupancy) or no class cap is met, fail verification (or at minimum emit a named failing check) instead of demoting silently. Radius above both caps with an honest layout is arguably still legitimate (unrestricted g is defined) — the clear bug is the dishonest-layout case.

FarLab:
Closes #661.

A layout that violates site spacing >= 1 or qubits-per-site <= layers previously demoted silently to `unrestricted` and merged green. It now fails verification with named checks (`site_spacing_at_least_one`, `site_occupancy_within_layers`). An honest layout above every class cap stays valid — unrestricted g is defined — but the demotion is now surfaced via a `locality_class_computed` check detail.

Tests: crammed layout upgraded to a hard-fail assertion, new over-occupancy case, over-radius case asserts the named demotion. Scanned all 356 merged codes: none change status. `validator_manifest.json` re-pinned for the `qldpc_verify.py` change.

### DIFF
--- a/verify/qldpc_verify.py
+++ b/verify/qldpc_verify.py
@@ -419,8 +419,14 @@ def _verify_semantic(doc, report, record, refute=False, seed=None):
     #          radius cannot be faked by collapsing qubits onto a point;
     #      (c) short range: the measured interaction radius (largest check
     #          diameter) is within the class cap.
-    #    Codes that fail (b)/(c), or carry no layout, are simply `unrestricted`
-    #    (no 2D-local class), not rejected. "Short range" means a bounded,
+    #    A layout that fails (b) FAILS VERIFICATION: attaching coordinates is a
+    #    locality claim, and a dishonest layout that merges green while earning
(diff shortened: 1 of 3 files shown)


MathysRennela (approved):
LGTM!

PR #705, opened 2026-08-24
Site: render the circuit tier

vprusso:
The circuit tier has been merged and verified since #646, and three entries (`25-1-5`, `49-1-7`, `81-1-9`) carry full circuit blocks, but `site/build.py` never learned the tier exists, so nothing shows on the page. This is the deferred display follow-up from #646/#680.

## What renders

Detail pages gain a Circuit tier section after the distance section, in its shape: link to the committed `circuits/<slug>/` artifacts, recipe/rounds/pinned-stim line, min d_circ with the penalty-only note, per-basis values with collapsible witness fault sets (mechanism indices in the committed `.dem`, exactly what the verifier checked). When an entry carries a measured rate, per-basis `ler/round` with 95% CI, raw counts, and the pinned decoder render below; until then the section says the floor exists and the prefactor is unmeasured.

Board rows whose entry ships circuits get a small gear chip next to the code name showing min d_circ (tooltip notes a measured rate when present). No new column, no ranking change; the chip is a pointer, not a score.

## Verification

Headless renders inspected, not just built: the `25-1-5` detail section and its board row from the real build, and the measured-rate branch against a synthetic block with the exact shape the verified end-to-end run produces (127/10^5 shots, CI, decoder id), since no merged entry can carry one until #680 and its seed measurement land. Entries without a circuit block are byte-unchanged.

Not in scope: the RFC's full presentation design (green/amber chip states, unclaimed-schedule bounty counts, a circuit score toggle). Those want the tier populated by more than seeds first.

### DIFF
--- a/site/build.py
+++ b/site/build.py
@@ -657,6 +657,12 @@ font-variant-numeric:tabular-nums}}
 max-width:220px}}
 .etal{{color:var(--mut)}}
 .hexwrap{{display:inline-flex;align-items:center;margin-left:8px}}
+/* circuit-tier chip (issue #505): gear + min d_circ on rows whose
+   entry ships verified memory circuits */
+.circchip{{display:inline-flex;align-items:center;margin-left:8px;
+font-size:11px;color:var(--mut);border:1px solid var(--ln);
+border-radius:999px;padding:0 7px;line-height:16px;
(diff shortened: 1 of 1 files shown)


FarLab:
LGTM!

FarLab (approved):

PR #703, opened 2026-08-24
Add [[136,34,12]] GALA quasi-cyclic code (reconstructed from arXiv:2608.07431)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[136,34,12]]
- Tracks: unrestricted / weight-9plus (computed by the verifier from H and the layout)
- Distance confidence: X: upper_bound, Z: upper_bound

Family tag: other (a self-declared filter, never used for ranking).

### What frontier does this advance?

- **unrestricted / any weight**: on the Pareto frontier. Score kd²/n = 36.0,
  max check weight 12.
- Dominates no existing entry outright but is non-dominated with all current
  board entries in the cell (verified by the gate).

### Provenance

Reconstruction of a certified instance from arXiv:2608.07431 (Yang,
Duckering, Dua, "Designer Codes from GALA", Table S5). The paper publishes
explicit group-ring generators and certifies d = 12 exactly (exhaustive
exclusion of lighter logicals + verified weight-12 witness). Our
reconstruction reproduces the paper's n and k exactly; our RIS witnesses at
20k trials/side agree with the certified distance. The trusted gate passes
(`passed: true`, not refuted) and confirms this code advances the board.

Research note: `notes/136-34-12.md` (direction, evidence trail, dead ends,
reproduction recipe — the construction is reproduced inline there).

### Checklist

- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/136-34-12.json` passes locally
- [x] Construction and references filled in under `provenance`
- [x] Research note included (`notes/136-34-12.md`)
- [x] Novelty: known parameters from the literature (arXiv:2608.07431);
      `provenance.novelty` set accordingly — this PR's claim is that the
      *board* advances, not that the code is new to the literature

PR #706, opened 2026-08-24
Fieldnote: GALA reconstruction campaign (2 board advances, negative results on non-abelian lifts)

MathysRennela:
## Summary

Adds two fieldnotes:

1. **2026-08-24 GALA reconstruction** — reconstructed all 8 abelian-bottom GALA rows of arXiv:2608.07431 from the paper's explicit generator tables. Two rows ([[136,34,12]], [[192,40,12]]) advance the weight-9plus × unrestricted board and are submitted separately (#703, #704). Two more ([[132,30,12]], [[672,336,12]]) were found already seeded on the board from the same paper. Documents the block-circulant convention that makes/breaks CSS, and the negative result on non-abelian S3-top lifts: without the paper's per-row active sets Γ, the [[480,240,10]] reconstruction yields girth 4 instead of ≥6 (all 576 monomial orderings tried).

2. **2026-08-23 UB sweep** (from a prior campaign, not yet PR'd) — univariate-bicycle Table I reconstruction; 2 of 9 rows advanced at the time.

No code submissions in this PR — codes ride in #703/#704 per one-code-per-PR.

### DIFF
--- a/fieldnotes/2026-08-23-univariate-bicycle-sweep.md
+++ b/fieldnotes/2026-08-23-univariate-bicycle-sweep.md
@@ -22,16 +22,20 @@ and **two advance the `weight-8 × unrestricted` board**:
 - **[[178,24,13]]** — a(x)=1+x⁹+x¹⁰+x¹² over R₈₉, ℓ=5; kd²/n ≈ 23.9
 
 Both are witness-backed upper bounds (gate refutation at ~8k RIS trials found
-nothing lighter); no exact or WL-equivalent duplicates. Staged under
-`research/candidates/ub-*` with verdicts; reconstruction script committed at
-`research/ub_sweep.py`. Not promoted to `codes/`, no PR — human review decides.
+nothing lighter); no exact or WL-equivalent duplicates. The reconstruction
+script is committed on the board at `research/ub_sweep.py` (merged via the
+[[178,24,13]] submission, github.com/unitaryfoundation/qldpc-challenge @
+1beea345). Both advancing rows were later promoted to `codes/` with full notes:
+see `notes/124-14-11.md` and `notes/178-24-13.md`.
 
 ## What worked
 
-- The resumable arXiv metadata harvester (`research/literature/`) caught this
-  paper on its first incremental run after a 7-day gap; the abstract's
-  "single-polynomial search" framing plus an explicit Table I with exact
-  supports made it a cheap, high-confidence reconstruction target.
+- The resumable arXiv metadata harvester (committed alongside the pilot's
+  other working output in github.com/MathysRennela/qldpc-challenge-1 @
+  1c226252, branch quarantine/research-2026-08-21, `research/literature/`)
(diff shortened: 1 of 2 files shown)


vprusso (changes requested):
The GALA note is exactly what fieldnotes are for, and the two best things in it are the ones most often left out: the sign-convention gotcha (`roll(base, +i)`, with the wrong sign failing loudly at CSS) and the honest non-abelian negative with concrete reopen conditions. The dominance claims for the five non-advancing rows all check out against the board, and #703/#704 are merged, so the campaign's two advances are already real.

The prose gate is failing on three citations, and one of them is more than hygiene:

- `research/candidates/gala_recon.py` and `research/candidates/gala/` are gitignored working output, which the note contract says can never be cited as evidence. For the script this matters substantively, not just formally: the note's strongest claim, the negative result that all 576 monomial orderings of the published S3-top generators yield girth 4, is uncheckable and unreproducible without `gala_recon.py`. That script is the evidence artifact for both the reconstructions and the negative. Commit it (`research/gala_recon.py`, the `research/circuit_seed.py` precedent) and cite the committed path; then the staging mention can just say "staged locally" without a path.
- The UB note's `research/literature/` has the same problem in lighter form: point it at something in the tree or name and pin the external source.

One internal inconsistency to fix while editing: the GALA summary still says "Not promoted to codes/, no PR — human review decides", but #703/#704 are merged; the note's own header line already says "submitted separately", so the stale sentence just needs deleting.

Also worth one sentence somewhere in the note, since it is the kind of fact the next reconstruction campaign needs: the [[480,240,10]] failure means the paper's Tables S3/S4 are NOT machine-recoverable for non-abelian tops without Γ, in contrast to the abelian rows. You say this in pieces; saying it once, plainly, is the reusable lesson.

Requesting changes only for the prose-gate items; the content is right.

vprusso:
Once the failing check passes, LGTM @MathysRennela !

MathysRennela:
Prose check fixed and green ✅ (run: https://github.com/unitaryfoundation/qldpc-challenge/actions/runs/32879575230)

The failure was `verify/check_prose.py` rejecting four citations that a reviewer could not open:

- `research/candidates/gala/` and `research/candidates/gala_recon.py` — gitignored staging output, which can never serve as evidence. Replaced by pointers to the committed reproduction recipes in `notes/136-34-12.md` and `notes/192-40-12.md`.
- `research/ub_sweep.py` — exists on the board but not in this PR's tree; now cited with its upstream pin (github.com/unitaryfoundation/qldpc-challenge @ 1beea345) plus the promoted notes.
- `research/literature/` — was never committed to any main branch; now pinned to the pushed quarantine branch (github.com/MathysRennela/qldpc-challenge-1 @ 1c226252, branch `quarantine/research-2026-08-21`).

No content claims changed — only the evidence trail now resolves.

vprusso (approved):

PR #709, opened 2026-08-25
Add [[12,2,3]] SAT-solver code discovery

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[12,2,3]]
- Tracks: unrestricted / weight-4 (computed by the verifier from H and the layout)

Family tag: other (a self-declared filter, never used for ranking).

### Checklist
- [x] `python verify/qldpc_verify.py codes/12-2-3.json` passes locally
- [x] Checked against existing entries: not equivalent to any current board entry (gate dedup clean; nearest weight-4 neighbors are [[7,1,3]] and [[16,2,4]], both distinct codes)

Weight-4 × unrestricted cell. The only weight-4 entries at comparable size are [[7,1,3]] and [[16,2,4]]. This code sits on the d=3 frontier with k=2: it strictly extends [[7,1,3]] on k (2 > 1) at equal distance, and is not dominated by any entry ([[16,2,4]] has higher n; nothing weight-4 offers k ≥ 2 at n ≤ 12). Score kd²/n = 1.5.

Construction: SAT-solver code discovery (arXiv:2608.23460 Sec. VI adapted to CSS): Minisat22 CNF over X/Z row-incidence variables; even-overlap commutation chains, weight≤2 detection clauses, Sinz row-weight bound w≤4; solution enumeration via blocking clauses. Generator committed in this PR as `research/sat_search.py`.

Research note: `notes/12-2-3.md`

### DIFF
--- a/codes/12-2-3.json
+++ b/codes/12-2-3.json
@@ -1,0 +1,103 @@
+{
+ "schema_version": "0.1",
+ "name": "[[12,2,3]]",
+ "code_type": "CSS",
+ "n": 12,
+ "k": 2,
+ "checks": {
+  "X": [
+   [
+    0,
+    2,
+    4,
+    9
+   ],
+   [
+    0,
+    6,
+    7,
+    8
+   ],
+   [
+    1,
+    2,
+    3,
+    10
+   ],
+   [
+    1,
+    7,
+    8,
+    11
+   ],
+   [
+    3,
+    5,
+    7,
+    9
+   ]
+  ],
+  "Z": [
+   [
+    4,
+    7,
+    8,
+    9
+   ],
+   [
+    2,
+    5,
+    9,
+    10
+   ],
+   [
+    0,
+    5,
+    6,
+    9
+   ],
+   [
+    1,
+    3,
+    6,
+    7
+   ],
+   [
+    1,
+    2,
+    4,
+    11
+   ]
+  ]
+ },
+ "distance": {
+  "d": 3,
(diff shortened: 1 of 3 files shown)


MathysRennela:
@vprusso I think it makes sense to add the SAT-solver kit to this PR. Let me know what you think.

vprusso:
Withdrawing what I pushed here: the certifier belonged in its own PR, and it is now #711. The submission-scope gate failed this run correctly, because a PR carrying untrusted code data must not also change the code that verifies it. This branch is reset to your commit 91295b1, so it is exactly what you had.

The benchmark numbers I quoted above were wrong twice, and the final versions are in #711. Briefly, for anyone reading this thread rather than that one: the claim that symmetry breaking contributes nothing came from k=6 and k=8 instances where both variants finish in under a second and cannot be separated; the claim that it is what closes the k=28 instance came from a run against a 900 s cap on a host carrying load 15, where the timeout measured the budget and the contention. On a quiesced host both encodings close 126-28-8, symmetry breaking runs 203 s and 605 s against 913 s and 1001 s without, and the arms do not overlap.

What stands from the original comment is the reason to want any of it: the board holds 29 certificates whose solver field reads CryptoMiniSat while the repo ships no SAT certifier, so those artifacts are not reproducible from the repository. #711 closes that.

Your [[12,2,3]] and its generator are unaffected by all of this.

vprusso (approved):

PR #717, opened 2026-08-25
Add qldpc targets, showing which track cells are open

### LINKED ISSUE #639: Add a qldpc subcommand that shows which track cells are open
`qldpc` has two subcommands: `submit` and `recent`. A newcomer can see what landed lately and can submit a code, but there is nothing that answers the question they actually have first, which is what to aim at.

`TRACKS.md` describes the reference bars in prose, and the grid is a real structure, but working out which cells are thin means reading 332 JSON files by hand. Current occupancy, counted directly:

| locality | weight-4 | weight-6 | weight-8 | any |
|---|---|---|---|---|
| with layout | 25 | 40 | 13 | - |
| unrestricted | 11 | 73 | 61 | 109 |

Some cells are far thinner than others, and that is exactly where a first submission has the best chance of landing on a frontier.

The logic to do this already exists. `qldpc submit` reuses the site's own cell and Pareto code to fill in the "what frontier does this advance?" section of the PR body **after** a code is built. The same call, run over the board instead of over one candidate, answers the question up front.

## What to do

Add `./qldpc targets` (name negotiable) that prints, per track cell:

- how many codes are in it;
- the current Pareto frontier entries;
- for a chosen `n`, the `k` and `d` that would land on that frontier.

Optionally `--cell weight-6/unrestricted` to focus, and `--n 200` to answer "if I build something at n=200, what do I need?".

Reuse the existing helpers in `site/build.py` rather than reimplementing membership or the frontier test. If that means lifting a function into a shared module, that refactor is welcome and is most of the work.

## Acceptance

- `./qldpc targets` runs with no arguments and prints all cells.
- The numbers agree with the published board for at least one cell, checked by hand.
- README's "Start here" table gains a row pointing at it.

Some QEC vocabulary appears (n, k, d, check weight) but no theory is required: the frontier computation is already written and tested.

vprusso:
Closes #639.

## What this adds

`./qldpc targets`, which prints per track cell: how many codes it holds, how
many are nondominated, the best `kd²/n`, and the leading frontier entries.

```
unrestricted / weight ≤ 6
  176 codes, 115 nondominated, best kd2/n 19.20
    [[360,12,24]] w=6 kd2/n=19.20
    [[510,16,24]] w=6 kd2/n=18.07
    ...
```

`--cell weight-6/unrestricted` focuses one cell. `--n 200` switches the
listing to the entries a code of that size would have to get past, which is
the form the question usually arrives in.

## Where the numbers come from

It calls the site's `cells()` and `pareto()`, which `qldpc submit` already
imports to fill in a submission's frontier section, so no membership or
frontier logic is reimplemented and no refactor was needed.

That matters more than convenience. A command like this is worse than absent
if it drifts from the board, because it would send people at cells that are
not actually open, so a test asserts the counts cell by cell against
`load_entries()` rather than sampling one.

Checked by hand against the published board, as the issue asks. All 12
populated cells agree, including the two I verified against the rendered page:
52 codes in 2D-local bilayer / weight ≤ 4, and 176 with best 19.20 in
unrestricted / weight ≤ 6.

## Note on "nondominated"

The counts are high (115 of 176 above) because domination is over four axes at
once, n, k, d and check weight, so most entries win on something. That is the
same rule that awards record stars, so the number is honest, but it is worth
knowing that a large nondominated count does not mean a cell is easy. The
output says so in a closing line rather than leaving it to be inferred.

## Tests and cost

Four tests: runs with no arguments and covers every populated cell; counts
match the site exactly, cell by cell; `--cell` narrows and rejects nonsense;
`--n` reports what must be beaten. The module caches the board once, since
`load_entries()` re-verifies every code at about 15 s a call and four calls
made the file cost 93 s against a CI job of under four minutes. Cached, it
runs in 16 s.

README's "Start here" table gains the row the issue asks for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ See [`AGENTS.md`](AGENTS.md) for the precedence rule.
 | Submit a code you already have | [`CONTRIBUTING.md`](CONTRIBUTING.md) — one command: `./qldpc submit` |
 | Have an LLM find and submit a code on your behalf | [Contribute with an LLM](CONTRIBUTING.md#contribute-with-an-llm) (a ready-to-paste prompt) |
 | Have an agent search without publishing | [`research/AUTORESEARCH.md`](research/AUTORESEARCH.md) (stage-only research workflow) |
+| See which track cells are open right now | one command: `./qldpc targets` (add `--n 200` for what a code that size needs) |
 | Understand the boards and the targets to beat | [`TRACKS.md`](TRACKS.md) — especially the "Reference bars" section |
 | Point a coding agent at this repo | [`AGENTS.md`](AGENTS.md) |
 
--- a/cli/qldpc.py
+++ b/cli/qldpc.py
@@ -34,6 +34,7 @@ import argparse
 import datetime
 import json
 import os
+import re
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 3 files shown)


FarLab (approved):
LGTM

PR #720, opened 2026-08-25
Add [[674,170,d<=76]] generalized bicycle code (kd²/n = 1456.85)

vprusso:
## Code submission

- Parameters: `[[674, 170, d<=76]]`, check weight 28
- Track: unrestricted / weight-9plus (computed by the verifier)
- Distance confidence: X and Z both `upper_bound`
- Score: kd²/n = **1456.85**, against the board's current best of 1300.90
  (`666-150-76-b`)

### Checklist
- [x] Distance witness included for each side
- [x] `verify/qldpc_verify.py` passes locally
- [x] Checked against existing entries: nondominated on (n, k, d, w) against
      every current entry

## Construction

Prime-lift generalized bicycle on Z_337. Both `a` and `b` are weight-14
codewords of a designed degree-85 divisor ideal of `x^337 - 1`, so
`k = 2 deg gcd(a, b, x^337 - 1) = 170` by construction rather than by search.

Prime `m` is what makes the rate land on a clean grid: for prime `m`,
`x^m - 1` factors into `(x - 1)` and irreducibles all of degree `ord_m(2)`,
which is 21 here, so a divisor's degree can only be `21j` or `21j + 1`. Degree
85 gives k = 170 at n = 674.

The supports were drawn by a Prange search restricted to a heavy weight band
(per-side weight 14 to 16, row weight 28) rather than the lighter band used
earlier in this family. That choice came from a measured correlation between
row weight and confirmed distance within this family, and this code is the
first from the heavy band to survive confirmation.

## Distance evidence

The claim is a witness-backed upper bound, not an exact-distance claim.

Screening ladder: 4k, then 200k, then 2M, then two 20M-trial passes, the last
confirming `d <= 77`. Witness extraction then produced a weight-76 X logical.
The matching Z logical is that witness's image under the family's reversal
symmetry: swapping the two circulant blocks and reversing each maps `ker H_Z`
to `ker H_X` and preserves weight, because conjugating a circulant by the
reversal permutation transposes it. So `d_X = d_Z = 76` here rather than being
two independent searches.

Both witnesses were revalidated with `verify/gf2.py`, independently of the
compiled search that proposed them: each lies in the correct kernel and raises
the rank of the opposite check matrix, so neither is a stabilizer. A further
10M-trial self-attack found nothing lighter.

## Where it stands

Nondominated on (n, k, d, w) against every current entry, and it takes the top
`kd²/n` on the board. Worth stating plainly what that does and does not mean:
the distance is an upper bound from randomized search, so a deeper attack could
lower it. It has had roughly three deep passes; the previous record in this
family needed eight before it settled. If someone refutes it, that is the
mechanism working rather than a surprise.


MathysRennela (commented):
I have myself tried to obtain a similar code. A research note would be helpful to highlight what made it work.

vprusso:
Note added: `notes/674-170-76.md`.

Since you have searched this family yourself, the short answer to what made it work is one number. Across 24 confirmed candidates here, row weight and confirmed distance correlate at r = +0.615: weight 24 averaged d 73.6, weight 27 averaged 75.0, weight 28 reached 77. The board's weight cap is 32 and it was going unused, because the Prange collector was being asked for light codewords out of habit rather than for a reason. Raising the collection band from per-side weight 8-16 to 14-16, so row weight 28-32 instead of 24, is the entire difference between this code and the earlier ones from the same construction. Everything else, prime m = 337, the degree-85 divisor, the ladder, was already in place and had not produced anything above the bar.

The note also carries the failures, which are most of the search and probably more useful than the hit:

- even m collapses outright, since x^m - 1 = (x^m' - 1)^(2^v) forces divisors to be unions of whole Frobenius blocks. Predicted, then confirmed at m = 336 degree 96: 13,780 divisors, zero sparse codewords. Not few, none.
- the higher-rate direction is barren for the opposite reason: degree-105 and 126 ideals have high enough minimum weight that the band finds almost nothing, 3 fertile divisors in 95,515 attempts, and both codes built from them died at the 200k rung.
- reciprocity-closed divisors show no light-word enrichment over open ones.
- a BCH-style designed coset run, chosen to force a designed distance, lands at the 2nd percentile of 1820 random coset choices.

One measurement in there is worth having regardless of this code: 2M to 20M deflation in this family has mean ratio 0.908 over 26 matched candidates, mean drop of 7 in d, with two outright collapses (79 to 27, 78 to 28). A 2M reading is a lottery ticket rather than a measurement, which is why the claim here rests on the extracted witness rather than on any rung.

Exact supports for a and b are in the note's reproduction section, so the code rebuilds from the two lines without running the search.

PR #725, opened 2026-08-26
Add [[50,6,3]] SAT-found punctured rotated surface code

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[50,6,3]]
- Tracks: local-2d-single / weight-4 (computed by the verifier from H and the layout)
- Distance confidence: X: upper_bound, Z: upper_bound

Family tag: other (a self-declared filter, never used for ranking).

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/50-6-3.json` passes locally
- [x] Construction and references filled in under `provenance`
- [x] Research note at `notes/50-6-3.md`

### What frontier does this advance?

Track: **weight-4 × local-2d-single** (geometric efficiency g = 4kd²/(nρ²r⁴)).
This extends the small-n end of the g > 1 band: at n = 50 with a single-layer
r = √2 layout it reaches kd²/n = 1.08 (g = 1.08), above the surface code's
g = 1.0 normalization point in the same size regime. The existing g > 1
entries ([[101,5,5]], [[197,5,7]], holey rotated variants) are all hand-designed
defect layouts at n ≥ 101; this is the first search-produced entry in the band,
found by encoding the punctured-RSC grammar directly as CNF and letting
CaDiCaL choose the hole pattern.

Distance is exact on both sides (scipy MILP: no logical of weight < 3 exists);
the submission carries witness-backed upper bounds per board convention.

Construction: SAT search over punctured rotated surface code grammar: qubits on integer grid, weight<=4 plaquette checks (r=sqrt(2)), per-check boundary truncation variables; detection of all weight<=2 errors encoded in CNF, solved with CaDiCaL. Distance confirmed exact (d=3 both sides) via scipy MILP.

PR #726, opened 2026-08-26
Add pre-push prose gate (worktree-based CI simulation)

MathysRennela:
## What

A local run of `verify/check_prose.py` validates cited paths against the working tree — where uncommitted files exist. CI validates against the PR tree only. Citations to uncommitted files therefore pass locally and fail remotely (observed on PR #725: the note cited a fieldnote that existed locally but was not committed).

## Fix

`verify/prepush_prose_check.sh <pr-body.md>` re-runs the prose check inside a clean git worktree of HEAD (committed content only), so it fails exactly when CI would. AGENTS.md documents it as a mandatory step before every push of a submission branch.

## Validation

- Reintroduced the exact bad citation that broke #725 in a scratch commit: gate fails with the same message CI produced.
- Reverted: gate passes.
- Note: this PR touches verify/, so per the submission-scope rule it must be reviewed/merged independently of any code-submission PR.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,6 +77,19 @@ Two more, same reason:
   comments, unticked checklist boxes, session URLs. If a checklist box is not true, make it
   true or say why.
 
+**Before pushing: run the worktree gate.** A passing local run of `check_prose.py` is not
+sufficient — it validates paths against your working tree, where uncommitted files exist.
+CI validates against the PR tree only, so citations to uncommitted files pass locally and
+fail remotely. Before every `git push` of a submission branch, run:
+
+```
+verify/prepush_prose_check.sh <pr-body.md>
+```
+
+It re-runs the check inside a clean worktree of HEAD (committed content only), which fails
+exactly when CI would. It must exit 0 before the push. Treat this as part of pushing, not
+as an optional review step — the working-tree check proves nothing about what CI sees.
+
 A note named `<n>-<k>-<d>.md` must state its own `[[n,k,d]]` first. Follow
 `notes/TEMPLATE.md`; its sections are what a later searcher reads.
 
--- a/verify/prepush_prose_check.sh
+++ b/verify/prepush_prose_check.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+# Pre-push prose gate: reproduce CI's view of a submission branch before pushing.
+#
+# CI checks the PR body and changed notes against a tree containing ONLY
+# committed files. A local working-tree run of check_prose.py cannot catch
+# citations to uncommitted files, because those files exist locally. This
+# script runs the check inside a clean worktree of HEAD (staged + committed
+# content only), so it fails exactly when CI would.
+#
+# Usage: verify/prepush_prose_check.sh <pr-body-file>
+#   The PR body must be given as a file; pass /dev/null if you have none yet.
+#
+# Exit 0 = safe to push. Anything else = fix the reported paths first.
+
+set -e
+
+BODY=${1:?usage: verify/prepush_prose_check.sh <pr-body-file>}
+ROOT=$(git rev-parse --show-toplevel)
+WT=$(mktemp -d "${TMPDIR:-/tmp}/prose-check.XXXXXX")
+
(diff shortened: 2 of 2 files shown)


FarLab (dismissed):
Right diagnosis, and the worktree mechanism is correct: you push HEAD, CI sees HEAD, so checking a clean worktree of HEAD reproduces CI exactly (I confirmed the `notes/ fieldnotes/` filter matches `changed_prose`). Two issues:

1. **Worktree leak on a common path.** The no-changed-notes branch uses `exec python3 ...`, which replaces the shell process, so the `trap ... EXIT` cleanup never runs — every body-only or code-only invocation leaves a stale directory in /tmp and a stale registration in `.git/worktrees`. Dropping the `exec` (call python3, let the trap fire) fixes it. Also minor: the header says "staged + committed content only", but a worktree of HEAD is committed-only — the behavior is right, the comment isn't.

2. **Non-blocking, but worth discussing: this reintroduces the manual-mandate pattern.** The recent `open_pr()` preflight exists because "remember to run X before pushing" doesn't survive automation — and that preflight still checks the working tree, so it has exactly the blind spot this PR describes. Folding the worktree trick into the preflight (or a git pre-push hook) would protect every path, not just agents that read AGENTS.md. Happy to see that as a follow-up rather than blocking here.

*Drafted with Claude (Fable 5).*

vprusso (commented):
Ran the script rather than read it. Confirming Farrokh's two points with measurements, and adding two divergences from CI that I think matter more than the leak.

## Confirming the leak, and bounding it

Reproduced exactly, and the attribution is cleaner than it looks. On the failing path (changed notes, non-`exec` branch) the trap fires and cleanup is complete: 0 worktree registrations, 0 `/tmp` directories. On the `exec` path it leaks 1 of each per invocation. So it is only the no-changed-notes branch, which is the common one for a code-only or body-only push, exactly as described.

## The gate does work

Worth recording since it is the whole premise. A note citing an uncommitted file:

- `check_prose.py` against the working tree: `prose check ok (1 file(s) checked)`
- `prepush_prose_check.sh`: fails with `path does not exist in this tree`, exit 1

So it catches the class of bug it was written for.

## Two ways it still diverges from CI

Both follow from what `prose.yml` actually does, which is to check out the base branch as `trusted`, check out the PR as `submitted`, and run **the base branch's checker** against the PR tree with `--base origin/${{ github.base_ref }}`.

**1. The script runs the PR's own checker.** CI deliberately does not: `working-directory: trusted` is the base checkout. So a PR that modifies `check_prose.py` is validated locally by its own modified rules. Demonstrated by loosening one branch of `resolves()` on a probe commit: the note that failed the gate a moment earlier passes it, exit 0, while CI would still fail it. That is the same shape of blind spot the PR is closing, one level up, and it bites precisely on PRs to the checker itself.

**2. `origin/main` is hardcoded**, both in the `CHANGED` computation and in `--base`. CI uses the PR's real base ref. Every open PR happens to target main today, so this is latent, but a stacked PR would compute the wrong changed-file set and compare against the wrong base, and it would do so silently.

Neither is hard to fix: run the checker from a second worktree of the base ref rather than from the branch, and take the base from `git rev-parse --abbrev-ref --symbolic-full-name @{u}` or an argument defaulting to main.

## On the manual-mandate point

Agreed with Farrokh, and I would go further on sequencing. The AGENTS.md wording ("treat this as part of pushing, not as an optional review step") is doing the work that a hook or the `open_pr()` preflight should do, and it only reaches agents that read AGENTS.md and then comply. I would rather see the worktree trick land inside the preflight in this PR and the AGENTS.md paragraph shrink to a pointer, than land the instruction now and the enforcement later, because the instruction is the part that does not survive contact with automation.

That said, the mechanism here is right and I would take it over nothing. If you would rather keep this PR small, dropping the `exec` and parameterising the base gets it to a state where the follow-up is purely a call-site move.

## Smaller

The header comment says "staged + committed content only"; a worktree of HEAD is committed-only, and staged-but-uncommitted files are invisible to it. The behaviour is what you want, the sentence undersells it, and someone will eventually rely on the sentence.

vprusso (approved):

PR #734, opened 2026-08-28
Add batched CUDA RIS campaigns and CI compile coverage

### LINKED ISSUE #620: Add batched campaign support for the CUDA RIS accelerator
## Goal
Make the existing CUDA RIS tool efficient for finalist campaigns.

## Scope
- Process several codes in one GPU run.
- Search both sides without repeated setup.
- Keep the CUDA context alive across candidates.
- Stop early when a target weight is found.
- Write reproducible search receipts.
- Integrate results with campaign manifests and witness staging.

## Implementation
- Add a batch input format or a long-running GPU request mode.
- Allocate reusable device buffers for multiple codes and both sides.
- Return one result per code and side, with seed and trial metadata.
- Add target-weight early stopping.
- CPU-check every returned support before saving it.
- Keep CUDA optional and out of CI validation.

This extends the existing `verify/ris_gpu.py` and `verify/ris_gpu.cu` tools. It does not make CUDA required for CI or validation. Every witness must still pass CPU verification.

### LINKED ISSUE #624: CI does not compile verify/ris_gpu.cu
`verify/ris_gpu.cu` is 825 lines of CUDA that nothing in CI compiles. `verify.yml` runs `make fast` (the C++ `gf2_fast` extension) in both trees, but `make ris` appears in no workflow, so a syntax error, a changed kernel signature, or a launch-config mismatch lands on main and is only discovered by whoever next has a GPU in front of them. #622 adds 258 lines to it, which is what surfaced this.

A GPU is not needed to catch the common failures. `nvcc` compiles for a target architecture without one present, so a compile-only job is enough:

```yaml
  ris-compile:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: Jimver/cuda-toolkit@v0.2.19
        with: { method: network, sub-packages: '["nvcc"]' }
      - run: nvcc -O3 -arch=sm_75 -c verify/ris_gpu.cu -o /dev/null
```

Pinning an explicit `-arch` matters: the Makefile defaults to `-arch=native`, which has no meaning on a GPU-less runner. `RIS_ARCH` already exists as the override.

What this catches: parse and type errors, wrong kernel launch signatures, shared-memory sizing expressions that do not compile, `__launch_bounds__` violations against the target arch. What it does not catch: races, incorrect `__syncthreads()` placement, and anything about actual numerical results. Those still need a GPU box, so this is a floor and not a substitute for the manual runs contributors already do.

Worth keeping non-blocking at first (`continue-on-error: true`) so a flaky toolkit install does not wedge unrelated PRs, then promoting it once it has been stable for a while.

MathysRennela:
## Summary

Implements the CUDA RIS campaign tooling requested in #620 and adds compile-only CI coverage requested in #624.

- Add JSON campaign manifests with campaign defaults and per-code/per-side overrides.
- Add the RISBATCH1 internal request stream and one-process batch execution.
- Reuse and grow CUDA device buffers across requests.
- Persist campaign receipts and CPU-verified witness files in explicit staging output.
- Stop each code/side independently when its target is beaten and retain aggregate failure status.
- Add an advisory nvcc compile job targeting sm_75.

## Verification

- `UV_CACHE_DIR=/tmp/qldpc-uv-cache uv run --frozen pytest -q verify/test_ris_gpu.py` — 6 passed, 5 skipped.
- Python compilation and `git diff --check` pass.
- Local nvcc verification was unavailable because nvcc is not installed.

Closes #620
Fixes #624

*Implemented using Codex*


MathysRennela:
@vprusso @FarLab my C++ skills are rusty so please double-check the .cu file

vprusso (approved):

PR #742, opened 2026-08-29
Add [[682,182,75]] "i want to ride my bicycle" (generalized bicycle)

natestemen:
## Code submission

**"i want to ride my bicycle"** — a generalized bicycle on the pseudoprime
block Z_341 (the code's display name, from the `name` field of the JSON).

- Parameters: [[n, k, d]] = [[682,182,d<=75]]
- Tracks: unrestricted / weight-9plus (computed by the verifier from H and the layout)

A note on the distance claim: `qldpc submit`'s own witness search (20k RIS +
2M accelerator trials) stopped at d <= 80; the submitted file instead carries
the deeper campaign witnesses (weight 75 on both sides, ~42M RIS trials across
two independent ladder passes, the 75 floor reached by three separate seeds),
so the claimed distance is the tighter, honest bound. Both witnesses were
revalidated with the `verify/gf2.py` criteria independently of the compiled
search that proposed them, and `verify/qldpc_verify.py` (including its live
refutation pass) passes locally on the file as submitted.

### Checklist
- [x] `python verify/qldpc_verify.py codes/682-182-75.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` — no existing board entry shares these parameters (the only n=682 entry is [[682,20,22]]); a parameter-identical sibling from the same search pocket (same a, different b) was found but not submitted, as noted in `notes/682-182-75.md`

- **unrestricted / any weight**: on the Pareto frontier. Score
  kd^2/n = 1501.10, above the board's current best 1456.85
  ([[674,170,76]], the prime-lift GB on Z_337): higher k (182 vs 170),
  larger n (682 vs 674), d 75 vs 76, equal max check weight 32 — a new
  frontier point, not a domination of the incumbent.
- **unrestricted / weight-9plus**: same statement (the classes nest).

Max check weight 32, locality class unrestricted (no layout claimed).

The construction is a generalized bicycle on Z_341 = Z_11 x Z_31 — the
smallest base-2 Fermat pseudoprime, whose ord_341(2) = 10 gives x^341-1 a
finer divisor-degree grid than the mined prime blocks (m = 337, 333), which
is what unlocks k = 182 at this block length. The search method (masked RIS
mining with subgroup-quotient poison filters and ideal-exclusion, then a
star search around each hit's own gcd) is documented in the research note;
per review feedback the one-off search script itself is not committed.

Research note: `notes/682-182-75.md`


natestemen:
strange times we live in. I don't understand anything here, just fyi.

FarLab:
it is strange @natestemen! but this is a nice way to quickly learn by just doing things. 

and on the PR: no need to commit the python script that you used in the research folder :)

natestemen:
i learned nothing by telling claude to make it to the top of the leaderboard lmao

FarLab:
that makes sense, I don't expect anyone to understand everything after a few claude prompts, but you might get curious about what these bicycle things are (what is the cyclic group Z_341 doing there, what are these polynomial, what is this RIS search, etc...) and slowly get some understanding depending on your curiosity.

vprusso:
> i learned nothing by telling claude to make it to the top of the leaderboard lmao

Yes, but your PR has a Queen reference, so it all balances out!

vprusso (approved):

MathysRennela:
@natestemen  I've been starting to use quiz-generating skills and it is a great way to understand what has been done. 
It's also good to remember that a big part of the efficiency of the autoresearch loop of this challenge relies on the documentation of each code and of many code discovery campaigns. As we progress this is creating a massive knowledge base.

natestemen:
> quiz-generating skills

can you share, i'd like to check those out.

PR #754, opened 2026-08-30
Add [[682,182,77]] generalized bicycle code

EthanFeld:
## Code submission

- Parameters: [[682,182,77]]
- Track: weight-9plus / unrestricted (computed from H; no layout supplied)

### Checklist

- [x] `python verify/qldpc_verify.py codes/682-182-77.json` passes locally
- [x] Possible equivalence is recorded in `provenance.notes`; current-board WL deduplication found no match

The local official-style gate reports that `[[682,182,77]]` advances the current weight-9plus x unrestricted frontier, with no exact duplicate, no WL-equivalent board entry, and no dominated entry reported. The provenance novelty label is `unknown`; current-board deduplication found no match.

Research note: `notes/682-182-77.md`

Construction: generalized bicycle over Z_341; `H_X=[A|B]`, `H_Z=[B^T|A^T]`; exact exponent lists and the 2M confirmation ladder are in the note.

Schema reference: https://github.com/unitaryfoundation/qldpc-challenge/blob/main/schema/code.schema.json


MathysRennela:
@vprusso @FarLab if #754 and #755 turn out to be the same code I propose to attribute it to both authors.

FarLab:
I am not sure if that is the right approach: shouldn't the one that opened a passing PR get the credit?
It is a challenge after all!

vprusso:
Thank you for the contribution, @EthanFeld ! Just a note as well that it looks like the verify pipeline is failing.

MathysRennela:
> I am not sure if that is the right approach: shouldn't the one that opened a passing PR get the credit? It is a challenge after all!

Good point. (I like the idea of promoting cooperation but the rule should be consistent so first come first served)

PR #756, opened 2026-08-30
Add [[684,12,73]] affine 2BGA weight-8 code

edoaltamura:
## Code submission

- Parameters: [[n, k, d]] = [[684,12,73]]
- Tracks: unrestricted / weight-8, computed by the verifier from the checks and layout
- Distance confidence: upper_bound on X and Z, with global d <= 73; not exact
- Family tag: `2bga-coset`
- LLM model used: `GPT-5.6 Luna`

### Checklist

- [x] Distance witnesses included for both reported sides
- [x] `python verify/qldpc_verify.py codes/684-12-73.json` passes locally
- [x] Equivalence review recorded in `provenance.notes` as checked and not equivalent

This advances the unrestricted / weight-8 frontier by extending the existing
`[[684,8,85]]` point on the logical-qubit axis. At the same blocklength and
maximum check weight, this code has `k=12` instead of `k=8`. Its conservative
deep-RIS distance upper bound is `d<=73`, so this is a new Pareto point with
larger `k`, not a claim that it has a larger distance than `[[684,8,85]]`.

The corresponding operational score is `k*d^2/n = 93.491` using the reported
upper bound. The trusted validator reports that the candidate is not dominated
in the unrestricted / weight-8 cell.

The construction is a two-block group-algebra CSS code over
`Aff(F_19)=C_19 semidirect C_18` with action `r=2`:
`H_X=[L(a)|R(b)]` and `H_Z=[R(b)^T|L(a)^T]`, where
`a=[182,323,76,217]` and `b=[294,208,142,82]`. The group has order 342 and
each check has weight 8. The construction is the two-block group-algebra
family described in [arXiv:2306.16400](https://arxiv.org/abs/2306.16400).

The deep distance audit used 100,000 RIS trials first, followed by three
independent 1,000,000-trial passes with pair depths 24, 32, and 48. Those
passes returned 83, 73, and 74 respectively. The final claim uses the
conservative minimum `d<=73`, and no deep pass found a witness lighter than
73. The submitted X witness has weight 73 and the submitted Z witness has
weight 94; both pass the kernel and non-stabilizer checks.

The reproducible search and audit are documented in `notes/684-12-73.md`.


codes/684-12-73.json
MathysRennela, line 1: model credited as "GPT-5 codex". Could you credit the specific model you used?

edoaltamura: done!

PR #765, opened 2026-08-31
Add [[443,23,5]] multi-band dense-packed code

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[443,23,5]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/443-23-5.json` passes locally

`local-2d-single x weight-4` and scores g = k*d^2/n = 1.2980 at
d = 5; the trusted gate returns `board_advancing: true`.

Research note: `notes/443-23-5.md`


FarLab:
Is it correct that there isn't a model specified?

MathysRennela:
> Is it correct that there isn't a model specified?

I patched all the PRs with missing models.

FarLab:
does it make sense to make the model field mandatory?

MathysRennela:
> does it make sense to make the model field mandatory?

I personally think so. I regularly have to check the model field by hand on submissions I review and it's not sustainable.

Issue #748, opened 2026-08-29
Direct Sums of Codes


EthanFeld:
If you take the direct sums of codes you can get a ton of Pareto-optimal codes with no contribution of any new construction. For instance, with just a basic sweep I found 480~ codes on the frontier with this. I didn't see anything in the rules against this but, it seems like it goes against the spirit of the challenge and would be a massive pain for maintainers. I think the best preventative measure is mandating a connected Tanner graph or these codes are fast enough to generate that the website could just auto gen them to fill in the frontier better. I'm happy to send any code or help in any other way.

MathysRennela:
I made the same realisation a few weeks ago.
I think that direct sums of codes are genuinely interesting in the context of QEC to study fault isolation, independent logical components and so on. But they are not a genuinely useful way to compose codes.
The fact that dist(A+B) = min(dist(A),dist(B)) makes direct sums struturally uninteresting. There are genuine ways to compose codes that actually have some added value and lead to genuine code discovery.
So I do not think either that direct sums have their place in this challenge.

MathysRennela:
Gave it a bit more thought: I think that the easiest fix is to require Tanner graph connectivity.

EthanFeld:
I was looking at it more and you can conjugate the two blocks with a cnot block and connect the Tanner graphs without changing any of the relevant metrics for the code other than increasing check weight by 1. I could be wrong but, I think if someone decides to setup direct sum codes where they try to disguise the origin, its impossible to distinguish it from a arbitrary stabilizer code.

MathysRennela:
That's clearly a risk but I would push back a bit:
- the challenge already limits the possibility of doing this type of hack by asking submitters to provide provenance.
- bumping the check weight by 1 means that there are a lot of cases where gaming the challenge with a Tanner graph will not lead to a better code in the same cell.
- eliminating this case probably requires to check another property (like Clifford equivalence to a direct sum).

EthanFeld:
Wait yeah mb i think a spectral decomposition algo can find the crosses

MathysRennela:
There is a connected issue with the issue you raise: board-advancing does not necessarily mean worth submitting.
I opened PRs with codes that sit on the Pareto frontier (e.g. PR #757-770); they're a selection of codes that in my view are singular enough to justify submitting them. I have 300+ codes that are all board-advancing but not worth submitting because they do not add much more information to the existing codes I submitted. They just fill in a spot that is not well populated. 

[...]

PR #824, opened 2026-09-01
Add [[100,4,5]] multi-band dense-packed code (dense frontier region)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[100,4,5]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/100-4-5.json` passes locally

d = 5; the trusted gate returns `board_advancing: true`.

Research note: `notes/100-4-5.md`


FarLab:
Checked the Tanner graph here: `[[100,4,5]]` has **4 connected components, each `[[25,1,5]]`**. It's a direct sum of four copies of `25-1-5`, which is on the board as a surface-code baseline (`origin=baseline`, `family=topological`, named "[[25,1,5]] surface code").

The cause looks like the `rows=1` corner of the multi-band family: with a single row the bands have nothing to fuse against, so `rows=1, m=N` gives N disjoint `[[d^2,1,d]]` patches. Every `rows>=2` PR in this batch is properly connected, so this isn't about the construction in general.

Worth flagging that the checklist's exact/WL-equivalence check can't catch this — a 4-fold direct sum isn't equivalent to any single board entry. Context in #748.

MathysRennela:
Withdrawing — the pushback is correct. Independent verification confirms the Tanner graph splits into 4 connected components of 25 qubits each (k=1 per component), each matching the rotated surface code [[25,1,5]] signature: this is a direct sum of four copies of the board's `25-1-5` baseline, not a new code. Root cause is the `rows=1` corner of the multi-band sweep, where bands never fuse; the checklist's exact/WL-equivalence check structurally cannot catch a direct sum, as noted. Closing in favor of the connected `rows>=2` entries in this batch.

PR #825, opened 2026-09-01
Add [[196,4,7]] multi-band dense-packed code (dense frontier region)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[196,4,7]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/196-4-7.json` passes locally

d = 7; the trusted gate returns `board_advancing: true`.

Research note: `notes/196-4-7.md`


FarLab:
Same as #824: `[[196,4,7]]` has **4 connected components, each `[[49,1,7]]`** — a direct sum of four copies of `49-1-7`, which is on the board as a surface-code baseline (`origin=baseline`, `family=topological`).

Cause is the `rows=1` corner of the multi-band family: with a single row the bands never fuse, so `rows=1, m=N` gives N disjoint `[[d^2,1,d]]` patches. All `rows>=2` PRs in this batch are properly connected.

The checklist's exact/WL-equivalence check can't catch this — a 4-fold direct sum isn't equivalent to any single board entry. Context in #748.

MathysRennela:
Withdrawing — same issue as #824, and the pushback is correct. Independent verification confirms the Tanner graph splits into 4 connected components of 49 qubits each (k=1 per component), each matching the rotated surface code [[49,1,7]] signature: a direct sum of four copies of the board's `49-1-7` baseline. Root cause is the `rows=1` corner of the multi-band sweep; the exact/WL-equivalence checklist check cannot catch a direct sum. Closing in favor of the connected `rows>=2` entries in this batch.

PR #830, opened 2026-09-01
Add [[169,1,13]] multi-band dense-packed code (dense frontier region)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[169,1,13]]

### Checklist
- [x] `python verify/qldpc_verify.py codes/169-1-13.json` passes locally

d = 13; the trusted gate returns `board_advancing: true`.

Research note: `notes/169-1-13.md`


FarLab:
This one is connected, but I think it's the plain rotated surface code. `[[169,1,13]]` is `rows=1 m=1`, and its combinatorial signature matches `[[d^2,1,d]]` at d=13 exactly:

| | baseline `81-1-9` | `121-1-11` (merged) | this PR |
|---|---|---|---|
| checks per type | 40 | 60 | 84 |
| weight-2 / weight-4 | 8 / 32 | 10 / 50 | 12 / 72 |
| qubit degrees 2/3/4 | 4 / 28 / 49 | 4 / 36 / 81 | 4 / 44 / 121 |

Every entry is the rotated-surface-code formula: `(d^2-1)/2` checks per type, `d-1` weight-2 boundary checks, 4 corners, `4(d-2)` edge qubits, `(d-2)^2` bulk.

The board already carries this family as baselines — `25-1-5`, `49-1-7`, `81-1-9`, all named "surface code" with `origin=baseline`. The WL-equivalence check passed only because the baselines stop at d=9. The same applies to `121-1-11`, which is already merged as a submission and is the d=11 instance.

Context in #748.

MathysRennela:
Good catch I am closing it. Basically the 3 PRs you flagged are edge cases that should not have been part of the sweep.

PR #871, opened 2026-09-01
README Start here leads with the routing table

### LINKED ISSUE #640: README Start here leads with agent policy instead of the human path
The first thing under README's "Start here" heading is agent policy:

> Choose the workflow before you begin. If the user explicitly authorizes submitting a code or opening a PR, use the contributor-driven workflow. Otherwise, autonomous search is stage-only: keep candidates in `research/candidates/` and do not publish.

That rule matters, but it is addressed to a coding agent, and it is written in the second person about "the user", so a human arriving at "Start here" reads a paragraph about someone else's authorization before reaching the table that tells them what to do. The routing table immediately below it is good and does the actual work.

Two smaller things in the same area:

- README and CONTRIBUTING both say the `.npz` holds keys `hx` and `hz`. The tool is more forgiving than that and accepts `hx`/`HX`/`H_X` and `hz`/`HZ`/`H_Z`; the error message lists all six. The docs are stricter than the code, which sends people to rename arrays they did not need to rename.
- Nothing on the human path states that the `@` in `--authors @handle` is load-bearing (see the attribution issue filed alongside this one).

## What to do

- Move the agent precedence rule out of the opening position: give it its own short subsection, or fold it into the existing `AGENTS.md` pointer row of the table.
- Lead "Start here" with the routing table.
- State the accepted npz key spellings where the file format is first described, in both README and CONTRIBUTING.

## Acceptance

- A first-time human contributor reaching "Start here" sees the routing table first.
- `AGENTS.md` remains the authority on the precedence rule; the README does not restate it in a way that can drift.

Documentation only.

i-safonoff:
Fixes #640.

- Moved the agent precedence rule out of the lead position in "Start here"; it's now referenced from the AGENTS.md row of the routing table instead, so the table itself is what a first-time contributor sees first.
- Documented the accepted npz key spellings (`hx`/`HX`/`H_X`, `hz`/`HZ`/`H_Z`) in both README and CONTRIBUTING, next to where the file format is first described in each.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,8 @@ GitHub handle, pass `--anonymous`; the CLI warns that the resulting entry will
 not be bound to an account.
 
 `mycode.npz` holds your parity checks under keys `hx` and `hz` (dense 0/1
-arrays or scipy sparse). The tool:
+arrays or scipy sparse). The tool also accepts `HX`/`H_X` and `HZ`/`H_Z` as
+alternate spellings for the same two arrays. The tool:
 
 - computes n, k (= n - rank H_X - rank H_Z) and the max check weight;
 - searches for the lightest logical on each side and records it as a
--- a/README.md
+++ b/README.md
@@ -33,11 +33,6 @@ one winner. Construction family is a separate filter tag. See
 
 ## Start here
 
-Choose the workflow before you begin. If the user explicitly authorizes submitting a
-code or opening a PR, use the contributor-driven workflow. Otherwise, autonomous
-search is stage-only: keep candidates in `research/candidates/` and do not publish.
-See [`AGENTS.md`](AGENTS.md) for the precedence rule.
-
 | You want to... | Read |
 |---|---|
 | Submit a code you already have | [`CONTRIBUTING.md`](CONTRIBUTING.md) — one command: `./qldpc submit` |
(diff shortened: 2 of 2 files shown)


vprusso (commented):
The restructuring is right and meets both acceptance criteria in #640: the routing table is what a first-time contributor now sees first, and the precedence rule is pointed at rather than restated, so it cannot drift from AGENTS.md.

One thing to fix before merge, because it is the same defect the issue was filed about.

The loader accepts four spellings per side, not three:

```python
HX = _pick(z, ("hx", "HX", "H_X", "Hx"))
HZ = _pick(z, ("hz", "HZ", "H_Z", "Hz"))
```

I checked each one against `load_checks` with a small npz and all four load:

| keys | result |
|---|---|
| `hx`/`hz` | loads |
| `HX`/`HZ` | loads |
| `H_X`/`H_Z` | loads |
| `Hx`/`Hz` | loads |

The PR documents the first three, so after merging the docs are still stricter than the code, which is the complaint in #640 ("sends people to rename arrays they did not need to rename"), just narrower than before.

This is not your miscount. The tool's own error message says six:

```
need H_X and H_Z arrays (keys hx/HX/H_X and hz/HZ/H_Z); found ...
```

and the issue copied that, and the PR copied the issue. Fixing only the docs leaves the source of the miscount in place for whoever reads the error next.

Two suggestions, either spelling is fine:

1. In README and CONTRIBUTING, list all four per side, or describe the rule rather than enumerate it, something like "the key names are case-insensitive and the underscore is optional, so `hx`, `HX`, `Hx` and `H_X` all work".
2. Update the error message in `cli/qldpc.py` to match, so the docs and the code agree at the point where a user actually hits the problem.

The second is arguably outside a documentation-only PR. If you would rather keep this one scoped to docs, item 1 alone is enough to merge and item 2 can be a follow-up issue.

PR #872, opened 2026-09-01
Add help text to five bare flags, reject a .json --out path

### LINKED ISSUE #638: Five qldpc flags have no --help text; --out silently means a directory
Five `qldpc` flags are declared with no `help=`, so they render bare in `--help`:

```
--notes
--name
--seed
--out
--days     (on `recent`)
```

`--out` is the confusing one. It looks like a filename but is treated as a directory:

```
$ ./qldpc submit steane.npz --authors @me --out /tmp/probe.json
IsADirectoryError: [Errno 21] Is a directory: '/tmp/probe.json'
```

A contributor reading `--out OUT` in the help has no way to know that.

## What to do

- Add a one-line `help=` to each of the five.
- State in `--out`'s help that it is an output **directory**, and that the filename is always `<n>-<k>-<d>.json`. If a path with a `.json` suffix is passed, either accept it as a filename or fail with a message that says which is expected, rather than raising `IsADirectoryError`.
- `--notes` and `--name` should say how they relate to `--note-file` and the auto-generated `[[n,k,d]]` name respectively.

## Acceptance

- `./qldpc submit --help` and `./qldpc recent --help` have no flags with empty help.
- Passing a `.json` path to `--out` produces a clear message, not a traceback.

No QEC background needed.

i-safonoff:
Fixes #638.

`--notes`, `--name`, `--seed`, `--out`, and `--days` had no `help=`, so they rendered bare in `--help`. Added a one-line `help=` to each.

`--out` was the confusing one: it's always treated as a directory (the filename is auto-generated as `<n>-<k>-<d>.json` after verification), so passing something that looks like a filename, e.g. `--out /tmp/probe.json`, surfaced as a raw traceback from deep in `makedirs`/`open` instead of a clear message. I reproduced it locally — the actual exception is `FileExistsError` when the path already exists as a file (not `IsADirectoryError` as in the issue body, but same root cause: the flag never checks whether the value looks like a filename).

Added an early check in `cmd_submit` that rejects a `.json` `--out` path before any verification work happens, and a regression test (`verify/test_cli_out_flag.py`) following the existing `test_cli_authors.py` pattern — asserts the check fires before the code is even loaded, and that a normal directory path still reaches loading.

Full suite (`run_tests.py --skip-slow`) passes: 137 passed, 8 skipped.

### DIFF
--- a/cli/qldpc.py
+++ b/cli/qldpc.py
@@ -443,6 +443,13 @@ def validate_authors(authors, anonymous=False):
 
 
 def cmd_submit(args):
+    if args.out.endswith(".json"):
+        raise SystemExit(
+            f"--out expects a directory, not a filename: {args.out!r}\n"
+            "The output filename is always <n>-<k>-<d>.json, computed after "
+            "verification, so it is not yours to choose. Pass the directory "
+            f"it should go in instead, e.g. --out {os.path.dirname(args.out) or '.'}"
+        )
     args.authors = validate_authors(args.authors, args.anonymous)
     HX, HZ, coords, _draft = load_checks(args.code)
     if args.coords:                      # explicit coords file overrides
@@ -700,7 +707,9 @@ def main(argv=None):
                         "version, e.g. 'Claude Opus 4.8' (not a bare 'Claude'), or "
                         "'human' (claimed, not verified; the verifier requires a "
                         "version if a model is named)")
-    s.add_argument("--notes", default="")
+    s.add_argument("--notes", default="",
+                   help="short free-text notes recorded on the submission "
+                        "(not the research note; see --note-file)")
     s.add_argument("--note-file", default="",
                    help="markdown research note staged as notes/<slug>.md and "
                         "rendered publicly beside the code: the search story "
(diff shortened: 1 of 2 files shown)


vprusso (commented):
Checked the three claims that carry this change and they hold. `--out` is only ever used as a directory (`os.path.join(args.out, f"{slug}.json")` at 469, `os.makedirs(args.out, ...)` at 477), so treating a `.json` value as user error is right. The five help strings match actual usage: `--name` against `args.name or f"[[{n},{k},{d}]]"`, `--seed` against the `ris_min_logical` and `gf2_fast` calls, `--days` against `--since={args.days} days ago`, `--notes` against `prov["notes"]`. The `--notes` versus `--note-file` disambiguation is the one a reader most needs, since those two names are almost indistinguishable at a glance.

One thing worth adding to the body: the fix is better than you describe it. You document the `FileExistsError` case, where the `.json` path already exists as a file. There is a second case you do not mention, and it is the worse one. When the path does not exist, `makedirs(..., exist_ok=True)` happily creates a directory named `probe.json` and the submission lands at `probe.json/<n>-<k>-<d>.json`, with no error at all. I reproduced both:

    path exists as a file:  FileExistsError
    path does not exist:    no error, creates directory, writes probe.json/672-12-75.json

A silent surprise is a stronger argument for an early check than a confusing traceback, so it is worth having in the rationale rather than left out.

Two behaviour notes, neither blocking.

The check changes `--dry-run`. Dry-run returns at 466, before the `makedirs` at 477, so `--out foo.json --dry-run` used to work and print a would-be path; now it exits. I think that is the correct call, since dry-run exists to preview what a real run does and a real run misbehaves here, but it is an unflagged change in behaviour and belongs in the body.

The check also sits ahead of `validate_authors`, so a bad `--out` now masks a bad `--authors`. Both are cheap and neither ordering is wrong, but it does reorder what a user sees first when both are wrong.

Nit: `endswith(".json")` is case sensitive, so `--out foo.JSON` still reaches the old behaviour.

The tests are the right pair. Asserting the check fires before `load_checks` is the part that matters, and including the positive case that a directory path still reaches loading is what stops the check from silently over-rejecting later. That second test is easy to leave out and doing so would have made the first one much weaker.

vprusso:
@i-safonoff are you still intending to fix and complete this PR?

PR #873, opened 2026-09-01
Correct [[700,206,28]] generalized-bicycle code

EthanFeld:
## Code submission

- Parameters: [[n, k, d]] = [[700,206,28]]
- Tracks: unrestricted / any weight (computed by verifier from H and layout)
- Distance confidence: upper_bound (X: 28, Z: 28)
- Family tag: generalized-bicycle (filter only; not a ranking claim)

### Checklist

- [x] `python verify/qldpc_verify.py codes/700-206-28.json` passes locally
- [x] Possible equivalence noted in `provenance.notes`

Correction: after the original PR was opened, a cyclic-shift audit found a validated weight-28 Z logical and its block-swap/index-reversal/cyclic-shift-2 X counterpart. The earlier d<=74 claim is superseded. The corrected code has n=700, k=206, d<=28, maximum check weight 28, and score upper bound `kd^2/n = 230.72`. It still advances the Pareto frontier for most logical qubits at d <= 28; the PR is updated to keep the board data and witness claims honest.

Research note: `notes/700-206-28.md`

Construction: generalized bicycle over lift 350 with independently optimized row-truncated checks; exact supports and witnesses are in `codes/700-206-28.json`.


EthanFeld:
Sorry about wasting compute, it turns out this code has a very straightforward weight 28 witness, just updated the PR

MathysRennela:
@EthanFeld prose check failed otherwise it is all good

EthanFeld:
@MathysRennela Sorry about that, should be working now

vprusso:
> @MathysRennela Sorry about that, should be working now

Still appears to have a broken prose pipeline build, @EthanFeld

EthanFeld:
> > @MathysRennela Sorry about that, should be working now
> 
> 
> 
> Still appears to have a broken prose pipeline build, @EthanFeld 

The problem is the actions captures the initial pr body which has a different code specification and therefore different file names that no longer exist. I'm not sure how to fix it.

vprusso:
Your diagnosis is right, and it is a repo defect rather than anything you did wrong. Here is the mechanism and the workaround.

`.github/workflows/prose.yml` is triggered by:

```yaml
on:
  pull_request:
```

With no `types:` that defaults to `opened, synchronize, reopened`. It does not include `edited`, so editing the PR body never reruns the check. The body itself reaches the checker from `github.event.pull_request.body`, which is the body as of the triggering event, so the run keeps replaying whatever the body was at your last push.

That is visible in the failing run: it captured

```
Research note: `notes/700-206-74.md`
```

while the tree now has `notes/700-206-28.md`.

The workaround is to make a push event, since that fires `synchronize` and carries the current body:

```
git commit --amend --no-edit
git push --force-with-lease
```

An empty commit works too if you would rather not amend. Nothing about the content needs to change; the check just needs a fresh event.

I hit this on my own submissions earlier and it cost me a while to work out, so the underlying fix is worth doing separately: adding `types: [opened, synchronize, reopened, edited]` to that workflow would make a body edit rerun the check, which is what everyone expects it to do. I will open an issue for that so it does not land on you.

On the substance: no need to apologise for the compute. Finding your own weight-28 witness and correcting the claim before merge is exactly the behaviour the board wants, and it is a good deal better than a d <= 74 entry going in and being refuted later.

EthanFeld:
@vprusso thank you, just did that

vprusso:
> @vprusso thank you, just did that

Thanks, @EthanFeld ! Assuming that passes, LGTM!

vprusso (approved):

Issue #594, opened 2026-08-17
Run an adversarial integrity audit


MathysRennela:
## Summary

Run a focused adversarial security and integrity audit of the qLDPC Challenge before the submission volume and use of automated agents grow. This should test the challenge’s trust boundaries.

## Threat model

Assume an attacker can:

- Open a PR from a fork.
- Submit arbitrary JSON under `codes/` and Markdown/prose accompanying it.
- Know the public verifier, CI configuration, and test cases.
- Try to maximize CI cost, evade distance refutation, manipulate leaderboard metadata, or inject content into the generated site.

The assets to protect are:

- `main` branch and maintainer credentials/tokens.
- CI runner availability and cost.
- Correctness and credibility of the leaderboard.
- Safety of generated public pages.
- Integrity of verification, provenance, and certification claims.

### 1. GitHub Actions and trust boundaries

Review and, where practical, test:

- Whether fork PRs can access secrets or write-capable tokens.
- Whether any submitted-tree command can execute attacker-controlled code.
- The submitted-tree `make fast`, test, dependency-installation, artifact, and path-handling steps.
- Whether workflow changes from a PR can run with more privilege than intended.
- Repeated PR updates, concurrency, cancellation, and CI amplification scenarios.
- Whether third-party actions should be pinned to immutable commit SHAs instead of tags.
- The `contents: write` site-build workflow, including race and post-merge behavior.

### 2. Parser and resource-exhaustion testing

Fuzz malformed and worst-case submissions against:

- `verify/qldpc_verify.py`
- `verify/validate_candidate.py`
- `verify/gate_changed.py`
- `site/build.py`

Include maximum-size valid inputs, duplicate or out-of-range indices, empty/repetitive checks, pathological locality coordinates, large numeric values, deeply nested JSON, and inputs designed to maximize runtime while staying within declared limits. Record peak memory, wall time, and whether failures are clean and bounded.

### 3. Generated-site output safety

Trace every submission-controlled field into generated HTML, inline JavaScript, SVG, URLs, and Markdown-rendered content. Test payloads in names, construction descriptions, authors, references, notes, model names, and other metadata, including attribute, script, URL, and Unicode/newline contexts.

Confirm that generated pages cannot produce stored XSS, unsafe links, malformed markup, or script-breaking inline JSON.

### 4. Distance-gate adversarial testing

[...]

PR #880, opened 2026-09-02
Add explicit timeout-minutes to every CI job

MathysRennela:
## Add explicit `timeout-minutes` to every CI job

GitHub's default is 360 minutes per job. Every resource path in the
pipeline is currently bounded by design — submission size caps, dense
matrix caps, and the refutation gate's wall-clock budget — but that
bound is a design invariant, not an enforcement mechanism: one hostile
regression in the verifier would convert a PR into a six-hour runner
occupation.

This adds an explicit cap to every job, sized at roughly 2–4x the
largest observed run over recent weeks (from workflow run history):

| Workflow | Job | Cap | Observed |
|---|---|---|---|
| lint | ruff | 10 min | seconds |
| prose | prose | 10 min | seconds |
| verify | ris-compile, verify | 20 min | 1.5–6 min |
| site | build | 15 min | < 1 min |
| refute-weekly | refute-board | 120 min | 50–58 min |

The pinned workflow entries in `verify/validator_manifest.json`
(`verify.yml`, `prose.yml`, `refute-weekly.yml`) are re-pinned
accordingly.

Found as part of the #594 integrity audit work.


vprusso (approved):

FarLab:
for some of these larger codes, the refutation takes around an hour. so with this PR, that would not be possible anymore?

MathysRennela:
> for some of these larger codes, the refutation takes around an hour. so with this PR, that would not be possible anymore?

@FarLab which cap do you suggest for verify?

FarLab:
I think a max of 90 mins per job for now is good, we don't have that many PRs of these bigger codes.

MathysRennela:
Took your review into account: the verify job cap is now 90 minutes.

MathysRennela:
@FarLab are you ok with the changes?

FarLab:
sounds good!

FarLab (approved):

PR #881, opened 2026-09-02
Make the authorship check fail closed on git failures

MathysRennela:
## Make the authorship check fail closed on git failures

`check_authorship.py` exited 0 when the diff against the base ref could
not be computed — for example with an unresolvable base ref, the tool
printed git's error and returned success, so CI would show a green
authorship result even though the check never ran.

This makes it exit nonzero with an explicit "failing closed" message in
that situation, mirroring what `gate_changed.py` already does for the
identical failure ("could not compute diff vs base …; failing closed").
A broken diff can no longer silently pass authorship.

Includes a regression test (an unresolvable base ref must exit nonzero)
and the one-line re-pin of `check_authorship.py` in
`verify/validator_manifest.json`.

Found as part of the #594 integrity audit work.


vprusso (approved):

vprusso:
Looks like the verify is failing, but other than that LGTM!

vprusso (approved):

PR #900, opened 2026-09-06
Add [[360,10,28]] 2BGA on Z_30:Z_6

shiy1022:
## Code submission

- Parameters: [[n, k, d]] = [[360,10,28]]
- Tracks: unrestricted / weight-8 (computed by the verifier from H and the layout)

### Checklist
- [x] `python verify/qldpc_verify.py codes/360-10-28.json` passes locally

- **unrestricted / weight ≤ 8**: on the Pareto frontier
- **unrestricted / any weight**: on the Pareto frontier

Score kd^2/n = 21.778, max check weight 8, locality class unrestricted.

To be clear about *how* it advances: it does not lead the cell on score. The
weight-8 unrestricted cell is led on kd^2/n by high-rate codes an order of
magnitude above this one. This code is non-dominated because the cell's
**low-k, high-distance corner at moderate n** is thinly populated — the frontier
runs over (n, k, d, w) jointly, so it joins by filling an empty corner rather
than by beating a leader. `qldpc submit` reports no board entry dominating it on
all four axes.

### Nearest board neighbours (why this is non-dominated)
Codes at comparable n *do* reach higher distance, but not inside this cell.
Weight is a ranking axis on this board, so a heavier code cannot dominate a
lighter one, and a larger-n code cannot dominate a smaller-n one:

| board entry | w | why it does not dominate |
|---|---|---|
| [[400,12,50]] | 12 | weight 12, so `weight-9plus` |
| [[396,10,39]] | 12 | weight 12, so `weight-9plus` |
| [[390,82,38]] | 32 | weight 32, so `weight-9plus` |
| [[390,82,32]] | 16 | weight 16, so `weight-9plus` |

This code is therefore the **weight-8 representative at moderate n** in a region
otherwise held by weight-9-to-12 codes at similar n, or by weight-6/7 codes at
much larger n ([[528,4,34]] and [[700,6,32]] at w = 6, n = 528 and 700).

### Distance evidence
Reported as a witness-backed **upper bound**, not exact — no integer program was
run, and certification is not expected to terminate at this block length.

The confirmation depth was calibrated rather than assumed, because a 1M-trial
floor turned out to be insufficient in this family and block-length band: of five
candidates from the same campaign, one was over-claiming at that depth and two
more had not yet reached their final value. This code was therefore confirmed at
**5M RIS trials/side** on a rising ladder (1M → 2M → 5M → 10M) under a seed
independent of the one that produced it, and held flat at d = 28 at every rung.
The per-side asymmetry (d_X ≤ 28, d_Z ≤ 32) is an artifact of the witness search
favouring its better side, not evidence about the Z side; the claim is the
minimum, which is a valid upper bound either way.

### Equivalence and literature
`validate_candidate.py`: no exact-fingerprint duplicate, no WL-equivalent board
entry. Independently checked against the complete published non-abelian 2BGA
enumeration of arXiv:2306.16400 (codes at github.com/QEC-pages/2BGA-codes,
`nonabelian.zip`, 71051 rows parsed): no published row dominates it. At n = 360
it lies beyond that enumeration's n ≤ 200 range, so its completeness result does
not settle the question either way. **Novelty against the wider literature
remains unverified.**

Construction: Two-block group algebra (2BGA) code over F2[G] with G = Z_30 : Z_6
metacyclic of order 180, the semidirect product with action j: i -> 19^j * i
(mod 30); 19^6 = 1 (mod 30) holds. Elements (i, j) with i in [0,30), j in [0,6)
are indexed i*6+j, identity at 0, as built by
`research/kit/group_algebra.py:metacyclic(30, 6, 19)`. Supports a = [61, 67, 81,
91] and b = [144, 160, 166, 171] as index sets in G. H_X = [L_a | R_b], H_Z =
[R_b^T | L_a^T], with L and R the left and right regular representations of
F2[G]; CSS commutation holds for any supports because left and right
multiplication commute. Both supports have size 4, so every check has weight 8.
Found by random support sampling over this family, screened with the repo's RIS
surrogate (`research/kit/surrogate.py`) and confirmed as described above.

Research note: `notes/360-10-28.md`


FarLab:
@shiy1022 thanks for your code contribution! Starring the repo would be much appreciated!

PR #918, opened 2026-09-07
Add weight-5 BB code [[90,4,9]]

cbjuan:
## Code submission

- Parameters: [[n, k, d]] = [[90, 4, 9]]
- Tracks: `unrestricted` x `weight-6` (and the looser `weight-8` / `weight-9plus` cells it also competes in — max check weight is 5)
- Distance confidence: exact

### Checklist
- [x] `python verify/qldpc_verify.py codes/90-4-9.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` (not applicable — not a duplicate/WL-equivalent of any board entry; see `validate_candidate` output below)

A weight-5 bivariate bicycle (BB) code, `A = 1 + x^2y^8 + x^3y^7`,
`B = 1 + x^2y^6` over `F_2[x,y]/(x^5-1, y^9-1)`. It advances the
`unrestricted` x `weight-6` board cell (and the looser weight cells above
it) as a new, non-dominated Pareto point at this (n, k, d).

Distance is **exact**: proved via this repository's own
`verify/certify.py` (scipy/HiGHS MILP) — no nontrivial logical lighter than
weight 9 exists on either side — not merely inherited from the source
campaign's internal claim. See `notes/90-4-9.md` for the full evidence
trail (surrogate search at 5k/50k trials, an independent BP+OSD decoder
corroboration, then the MILP certification).

This code was not found by a fresh search in this repository; it was mined
from an already-completed weight-5 BB evolutionary campaign in a companion
repository (https://github.com/qiskit-community/qcode-discovery, weight5-campaigns
branch — not yet part of a public branch there, so nothing here relies on
that source being externally checkable) and independently re-verified here
end to end using only this repository's own trusted tools (`research/kit/`,
`verify/`). This code is a connected component (ell=5, m=9) of a
disconnected parent presentation (ell=10, m=9, k=8, n=180); the parent is
not separately submitted here to avoid a low-novelty duplicate (see the
already-submitted `[[96,4,10]]`/`[[140,6,10]]` for the same pattern).

Per that campaign's own per-candidate model-attribution log, this code was
**independently discovered by both GPT-5.6-sol and Claude Opus 5** within
the same LLM-ensemble evolutionary run (GPT-5.6-sol first, chronologically;
Claude Opus 5 rediscovered it independently later in the same run) — this
is disclosed honestly as joint/dual attribution rather than collapsed to a
single model. This submission's re-verification and packaging in this
repository was done separately by Claude Sonnet 5.

`validate_candidate.py` verdict on this submission: `passed: true`,
`board_advancing: true` in cell `["weight-6", "unrestricted"]`, no
duplicate/WL-equivalent board entry, not refuted under random-seed
re-search.


FarLab:
Thanks for the careful attribution. One small change to `provenance.model`: the schema supports a list for exactly this case, so please use the array form rather than a free-text string:

```json
"model": ["GPT 5.6 Sol", "Claude Opus 5"]
```

Reasons:
- The site splits ensembles on `+` or `,`, not on "and", so the current string renders as one oddly long model name on the board instead of "GPT 5.6 Sol +1" with the full list on hover.
- The "(joint/independent discovery)" commentary is already in `notes`, which is where it belongs; the model field should hold names only.
- "GPT 5.6 Sol" matches the spelling the 13 existing board entries use for that model.

Everything else looks good, CI is green.

cbjuan:
Thanks for the comment, @FarLab. Attribution fixed in https://github.com/unitaryfoundation/qldpc-challenge/pull/918/commits/0d4768de4eb22ad0900b54da2461b51e2064935c

PR #923, opened 2026-09-07
Circuit tier: [[56,2,8]] GB memory circuits (d_circ = 7/7, interleaved depth-7)

FarLab:
Circuit-tier entry for the [[56,2,8]] generalized bicycle code (Wang, Pryadko, arXiv:2203.17216; the paper gives no circuit-level treatment) at d_circ = 7/7 (effective 7 against d = 8). **Corrections after opening:** an exact SAT search (kissat, weight cap 7) found a weight-7 undetected logical fault set in the X memory, and a locality-restricted RIS (edge-ball search, described below) found one in the Z memory; both heuristic searches originally cited had missed them, and the committed witnesses are now those sets.

- `circuits/56-2-8/memory_{x,z}.stim` + committed `.dem` (stim 1.16.0, canonical noise recipe, rounds = 8, 17.2k mechanisms per basis, under the 25k tier cap)
- `codes/56-2-8.json`: `circuit` block with witnessed d_circ.X = 7 (exact SAT find), d_circ.Z = 7 (edge-ball RIS find), schema 0.1 → 0.2
- Schedule: interleaved two-block extraction in the shape of the [[72,12,6]] entry (#724): joint reset/measure layers, 7 CX layers per round, X-check ancillas coupling their terms in slots 1–6 and Z-check ancillas in slots 0–5; committed term slots X [6, 3, 4, 2, 5, 1], Z [0, 3, 4, 2, 1, 5] (recorded in the JSON note).
- Method (code files only in this PR; the generator is not committed): the schedule generator recovers the A/B term structure from H alone (permutation matrices inside supp(A) commuting with B and vice versa, so no construction string is read), enumerates the 576 orders satisfying the overlap-parity criterion (even number of shared qubits touched X-before-Z for every overlapping check pair), confirms determinism with stim, and screens sampled orders. The committed order is the first of 2 screened orders that survived the seeded screen at full distance.
- Witness: the code-tier logical lifted into the DEM (data errors in the first idle layer), so d_circ ≤ d is witnessed by construction and the searches only try to refute it.
[...]

PR #924, opened 2026-09-07
Circuit tier: [[60,8,6]] twisted-torus BB memory circuits (d_circ = 6/6, interleaved depth-7)

FarLab:
Circuit-tier entry for the [[60,8,6]] twisted-torus bivariate-bicycle code (Liang, Liu, Song, Chen, arXiv:2503.03827; the paper gives no circuit-level treatment) at full circuit distance, d_circ = 6/6.

- `circuits/60-8-6/memory_{x,z}.stim` + committed `.dem` (stim 1.16.0, canonical noise recipe, rounds = 6, 13.5k mechanisms per basis, under the 25k tier cap)
- `codes/60-8-6.json`: `circuit` block with witnessed d_circ.X = 6, d_circ.Z = 6, schema 0.1 → 0.2
- Schedule: interleaved two-block extraction in the shape of the [[72,12,6]] entry (#724): joint reset/measure layers, 7 CX layers per round, X-check ancillas coupling their terms in slots 1–6 and Z-check ancillas in slots 0–5; committed term slots X [4, 3, 2, 1, 5, 6], Z [4, 3, 2, 5, 0, 1] (recorded in the JSON note).
- Method (code files only in this PR; the generator is not committed): the schedule generator recovers the A/B term structure from H alone (permutation matrices inside supp(A) commuting with B and vice versa, so no construction string is read), enumerates the 936 orders satisfying the overlap-parity criterion (even number of shared qubits touched X-before-Z for every overlapping check pair), confirms determinism with stim, and screens sampled orders. The committed order is the first of 1 screened orders that survived the seeded screen at full distance.
- Witness: the code-tier logical lifted into the DEM (data errors in the first idle layer), so d_circ ≤ d is witnessed by construction and the searches only try to refute it.
- Refutation evidence, two independent searches per basis: (i) a seeded search — for every mechanism j touching check 0 in three rounds (1315 of 13470 mechanisms; the group acts transitively on checks, so every fault set has a translate touching check 0) and every observable o, BP+OSD (ldpc, OSD-0) decodes j's syndrome under [H_dem; L_o] with the observable bit flipped, i.e. it returns the lightest set containing j that is undetected and flips o — lightest finds weight 6 (X) and 6 (Z); (ii) 2 seeds × 1500 RIS trials per basis (gf2_fast DEM kernel), lightest finds weight 8 (X) and 7 (Z). The committed witnesses are the lightest sets either search produced. `circuit_verify.py`, `verify_all.py` and the schema pass locally.
- Locality-aware search (added after opening; the strongest evidence here): an edge-ball RIS. Two detectors are adjacent when a mechanism touches both; for each check-0 detector c (any round, both check types; check 0 suffices because the check group acts transitively) and each neighbour c′, take the mechanisms whose detectors all lie in the radius-1 balls of c and c′ (about 2k mechanisms) and run RIS on the kernel of that sub-matrix (60 permutations). Any fault set that two of its own detectors cover this way is found near-exhaustively, which is the shape of hook-plus-late-error sets. Calibrated on known-bad circuits it finds the [[56,2,8]] weight-7 set of #923 in 9 min and two refuted [[48,4,8]] designs in 1 s and 7 min, where 4500 plain RIS trials per basis and the seeded search had found nothing. On these circuits it finds nothing below 6 in either basis over all 922 distinct ball pairs per basis (27 min each).
- Exact check (not obtained): kissat on the full DEM SAT encoding (weight cap 5) ran 9 h per basis without a verdict; every satisfiable instance met in this work resolved within an hour, so the silence is UNSAT-shaped, but it is not a proof.
- Why the seeded search is there: it is much stronger than RIS on these DEMs. It refuted RIS-clean orders for 56-2-8 and 48-4-8 (RIS: 3 × 1500 trials per basis, nothing below 8; seeded search: weight 7 in 17 s, and a weight-6 set on 48-4-8 that RIS only met on its third seed). Existing RIS-only d_circ claims on the tier deserve a pass with it; a pass over the merged entries is the natural follow-up.
- Sweep context: the same generator screened every two-block n<100 board code (RIS screen only, so "holds" below means RIS-clean, not seeded-clean): 70-6-8, 72-8-7, 32-8-4 and 42-8-3 held full distance; 30-4-6, 62-10-6, 88-8-7, 18-4-4, 24-6-4 and 96-12-4-mvb lose exactly one unit in every sampled order; 48-4-8 loses at least one unit in one basis in every one of 27 seeded-screened depth-7 orders (best 6/7); every weight-8 code with d ≥ 8 and the d ≥ 10 weight-6 codes exceed the 25k-mechanism cap at rounds = d.


MathysRennela:
## emlint v0.2.2 static DEM review

Ran [emlint](https://pypi.org/project/emlint/) v0.2.2 (`pip install emlint`, static analysis for stim Detector Error Models) against both committed DEM files in this PR. All files were taken from this PR's head branch (`submit-circuit-60-8-6`).

| File | Detectors | Observables | Mechanisms | Result |
|---|---|---|---|---|
| `circuits/60-8-6/memory_x.dem` | 360 | 8 | 13,470 | all 6 checks pass |
| `circuits/60-8-6/memory_z.dem` | 360 | 8 | 13,470 | all 6 checks pass |

Checks run (full default battery, exit code 0, no errors or warnings, no skipped checks, no unknown instructions, no disabled checks on any file):

- **detectability** — all error mechanisms that flip observables also trigger detectors
- **sensitivity** — all detectors are triggered by at least one error mechanism
- **observable_coverage** — all declared observables are flipped by at least one mechanism
- **probability_bounds** — all mechanism probabilities in (0, 0.5]
- **duplicates** — no duplicate mechanism signatures
- **correctability** — every syndrome maps to at most one distinct set of logical observables

Consistency notes:
- Mechanism counts match the PR body (13.5k per basis, under the 25k tier cap); detector count (360 = 6 rounds × 60) and observable count (k = 8) are as expected for rounds = 6.

Caveat: emlint is static DEM hygiene analysis. A pass here does not certify the claimed `d_circ = 6/6`; that rests on the refutation searches described in the PR body and remains `upper_bound` until certified by the trusted `verify/` stack.

PR #927, opened 2026-09-08
Circuit tier: [[48,4,8]] twisted-torus BB memory circuits (d_circ = 8/8, flagged interleaved depth-7)

FarLab:
Circuit-tier entry for the [[48,4,8]] twisted-torus bivariate-bicycle code (Liang, Liu, Song, Chen, arXiv:2503.03827; the paper gives no circuit-level treatment) at full circuit distance, d_circ = 8/8.

- `circuits/48-4-8/memory_{x,z}.stim` + committed `.dem` (stim 1.16.0, canonical noise recipe, rounds = 8, 19.6k mechanisms per basis, under the 25k tier cap)
- `codes/48-4-8.json`: `circuit` block with witnessed d_circ.X = 8, d_circ.Z = 8, schema 0.1 → 0.2
- Schedule: interleaved two-block extraction in the shape of the [[72,12,6]] entry (#724): joint reset/measure layers, 7 CX layers per round, X-check ancillas coupling their terms in slots 1–6 and Z-check ancillas in slots 0–5; committed term slots X [3, 2, 4, 5, 1, 6], Z [2, 4, 3, 0, 5, 1] (recorded in the JSON note).
- Method (code files only in this PR; the generator is not committed): the schedule generator recovers the A/B term structure from H alone (permutation matrices inside supp(A) commuting with B and vice versa, so no construction string is read), enumerates the 1 orders satisfying the overlap-parity criterion (even number of shared qubits touched X-before-Z for every overlapping check pair), confirms determinism with stim, and screens sampled orders. Without flags every one of 27 seeded-screened depth-7 orders lost at least one unit in one basis (best 6/7); the committed order is one of those, now with flags, and it survived the seeded screen at full distance.
- Flags: two flag qubits per check ancilla (Chao-Reichardt style). Flag A takes CX with its ancilla right after the ancilla's 1st and 5th data CX, flag B after its 2nd and 4th (X-check flags |0>, CX ancilla→flag, measured in Z; Z-check flags |+>, CX flag→ancilla, measured in X; each flag measurement is a detector). An ancilla fault spreading to 2 data qubits flips one flag and one spreading to 3 flips both, so a hook costs as many faults as the data errors it produces. Without flags every one of 27 screened depth-7 orders loses at least one unit in one basis (the refuting sets are one 3-qubit hook plus five late data errors); with them the first screened order held full distance in both bases. Flag layers add 6 CX layers per round and 96 qubits (192 total).
- Witness: the code-tier logical lifted into the DEM (data errors in the first idle layer), so d_circ ≤ d is witnessed by construction and the searches only try to refute it.
- Refutation evidence, two independent searches per basis: (i) a seeded search — for every mechanism j touching check 0 in three rounds (2565 of 19560 mechanisms; the group acts transitively on checks, so every fault set has a translate touching check 0) and every observable o, BP+OSD (ldpc, OSD-0) decodes j's syndrome under [H_dem; L_o] with the observable bit flipped, i.e. it returns the lightest set containing j that is undetected and flips o — lightest finds weight 8 (X) and 8 (Z); (ii) 2 seeds × 1500 RIS trials per basis (gf2_fast DEM kernel), lightest finds weight 9 (X) and 8 (Z). The committed witnesses are the lightest sets either search produced. `circuit_verify.py`, `verify_all.py` and the schema pass locally.
- Locality-aware search (the strongest evidence here): an edge-ball RIS. Two detectors are adjacent when a mechanism touches both; for a check-0 detector c (any round, both check types, its flags; check 0 suffices because the check group acts transitively) and each neighbour c′, take the mechanisms whose detectors all lie in the radius-1 balls of c and c′ (about 1.3k mechanisms), and run RIS on the kernel of that sub-matrix (60 permutations, about a second). Any fault set that two of its own detectors cover this way is found near-exhaustively, and hook-plus-late-error sets are that local. Calibration on the circuits this PR replaced: it finds the unflagged design's weight-6 set in the third ball (1 s), the first flag design's weight-7 set in 7 min, and the 56-2-8 weight-7 set of #923 in 9 min, where 4500 plain RIS trials per basis and the seeded search had found nothing. On the committed circuits it finds nothing below 8 in either basis over all ~1390 distinct ball pairs per basis (35 min each).
- Exact check (not obtained): kissat on the full DEM SAT encoding ran 6+ h per (basis, observable) instance without a verdict; a windowed encoding (mechanisms with detectors inside a 3-step time window, XOR-native CryptoMiniSat, symmetry-broken) was indeterminate on all 14 windows at a 1.6 h cap. Every satisfiable instance met in this work resolved within an hour, so the silence is UNSAT-shaped, but it is not a proof. The exact check did refute two earlier flag designs at weight 7: windows (1,5),(2,4) admit a *pair* of hooks on one ancilla bracketing data CXs 2–4 (both CXs of one flag inside, neither of the other), and (1,5),(0,3) admit a single hook at gap 3 whose one flagged detector is cancelled by a flag measurement error. The committed windows come from a brute force over all one- and two-fault ancilla patterns (faults riding on flag CXs included) requiring faults + flags flipped ≥ effective spread; (1,4),(2,5) is one of four safe two-flag designs for weight-6 checks.
- Why the seeded search is there: it is much stronger than RIS on these DEMs. It refuted RIS-clean orders for 56-2-8 and this code (RIS: 3 × 1500 trials per basis, nothing below 8; seeded search: weight 7 in 17 s, and a weight-6 set on 48-4-8 that RIS only met on its third seed). Existing RIS-only d_circ claims on the tier deserve a pass with it; a pass over the merged entries is the natural follow-up.
- Sweep context: the same generator screened every two-block n<100 board code (RIS screen only, so "holds" below means RIS-clean, not seeded-clean): 70-6-8, 72-8-7, 32-8-4 and 42-8-3 held full distance; 30-4-6, 62-10-6, 88-8-7, 18-4-4, 24-6-4 and 96-12-4-mvb lose exactly one unit in every sampled order; 48-4-8 without flags loses at least one unit in one basis in every one of 27 seeded-screened depth-7 orders (best 6/7); every weight-8 code with d ≥ 8 and the d ≥ 10 weight-6 codes exceed the 25k-mechanism cap at rounds = d.


MathysRennela:
## emlint v0.2.2 static DEM review

Ran [emlint](https://pypi.org/project/emlint/) v0.2.2 (`pip install emlint`, static analysis for stim Detector Error Models) against both committed DEM files in this PR. All files were taken from this PR's head branch (`submit-circuit-48-4-8`).

| File | Detectors | Observables | Mechanisms | Result |
|---|---|---|---|---|
| `circuits/48-4-8/memory_x.dem` | 1,152 | 4 | 19,560 | all 6 checks pass |
| `circuits/48-4-8/memory_z.dem` | 1,152 | 4 | 19,560 | all 6 checks pass |

Checks run (full default battery, exit code 0, no errors or warnings, no skipped checks, no unknown instructions, no disabled checks on any file):

- **detectability** — all error mechanisms that flip observables also trigger detectors
- **sensitivity** — all detectors are triggered by at least one error mechanism
- **observable_coverage** — all declared observables are flipped by at least one mechanism
- **probability_bounds** — all mechanism probabilities in (0, 0.5]
- **duplicates** — no duplicate mechanism signatures
- **correctability** — every syndrome maps to at most one distinct set of logical observables

Consistency notes:
- Mechanism counts match the PR body (19.6k per basis, under the 25k tier cap, and consistent with the "2565 of 19560" seeded-search scope).
- Detector count (1,152 = 8 rounds × 144, i.e. 48 check ancillas + 96 flag measurements per round) and observable count (k = 4) are as expected for the flagged schedule described.

Caveat: emlint is static DEM hygiene analysis. A pass here does not certify the claimed `d_circ = 8/8`; that rests on the refutation searches and the still-pending exact SAT check described in the PR body, and remains `upper_bound` until certified by the trusted `verify/` stack.

MathysRennela (approved):

PR #931, opened 2026-09-08
Circuit tier: [[12,4,2]] BB memory circuits at full d_circ = 2/2

FarLab:
Circuit-tier entry for the [[12,4,2]] twisted-torus BB code (arXiv:2503.03827) at **d_circ = 2/2** — the syndrome-extraction circuit realizes the full code distance in both bases.

- `circuits/12-4-2/memory_{x,z}.stim` + committed `.dem` (stim 1.16.0, canonical noise recipe, rounds = 2, 681 mechanisms per basis)
- `codes/12-4-2.json`: `circuit` block with witnessed d_circ.X = 2, d_circ.Z = 2, schema 0.1 → 0.2
- **Schedule**: interleaved two-block extraction (joint reset/measure layers, 7 CX layers per round), term slots X `[3, 4, 2, 5, 6, 1]`, Z `[4, 3, 2, 0, 1, 5]` over terms A_0..A_2, B_0..B_2, with the term structure recovered from `H` alone by a commuting permutation decomposition (no construction string is read).
- **Order choice**: RIS screen over the 1968 slot assignments satisfying the overlap-parity criterion. The first order tried lost the Z side (d_circ.Z = 1); the second holds both at 2.
- **Refutation evidence per basis**: 3 seeds × 1500 RIS trials on the committed DEM, plus an observable-constrained seeded BP+OSD search over symmetric seeds. The lightest undetected logical fault set either search reached is weight 2 in both bases, which is the committed witness.
- `circuit_verify.py`, `qldpc_verify.py` and `check_submission_scope.py` pass locally.

**Honest framing of the bar.** With d = 2 the schedule only has to avoid a weight-1 undetected logical fault set, so full d_circ here is a much weaker statement than for, say, [[42,8,3]] (#930) — the first screened order already failed it, but not by much. The value is coverage: this is the smallest board code with a certified circuit, produced by the same generator and held to the same evidence discipline as the larger entries.


MathysRennela:
Good to have a couple of d_circ=2 codes but overall it is not a practically useful code. Maybe the circuit tier should start at d >=3?

PR #937, opened 2026-09-08
Add [[682,140,86]] generalized-bicycle code

DennisWayo:
Submit `[[682,140,d<=86]]` cyclic-ideal generalized-bicycle code.

This updates the PR after the earlier `d <= 88` draft was refuted at weight
86. The submitted code is a witness-backed upper bound, not an exact-distance
claim.

Evidence:

- `codes/682-140-86.json` verifies locally with computed `k = 140`, maximum
  check weight 30, X witness weight 86, and Z witness weight 88.
- `notes/682-140-86.md` records the construction, search path, and follow-up
  evidence for this candidate.
- The X-side weight-86 witness is the valid RIS-fast witness found by the
  challenge refutation gate against the earlier `d <= 88` claim.
- The downgraded `d <= 86` claim also survived a local current-main CI-style
  refutation mirror: Python RIS found no lighter logical, and RIS-fast ran
  8,000,000 trials with no witness below 86.
- `verify/qldpc_verify.py codes/682-140-86.json` passed locally.

Score: `140 * 86^2 / 682 = 1518.240`.

This PR was supported by OpenAI Codex (GPT-5).

PR #945, opened 2026-09-09
Rerun the prose check when a PR body is edited

### LINKED ISSUE #897: Prose checker receives a stale copy of the PR Body
When running the prose check on a PR that changed the body, the changes aren't received so if the specification of the code changes it does not work.

This can be seen in the actions report for PR #873

FarLab:
Fixes #897.

## The bug

`.github/workflows/prose.yml` triggered on a bare `pull_request`, which defaults to `opened, synchronize, reopened`. Editing a PR body is an `edited` event, so it fired nothing. The body reaches the checker from `github.event.pull_request.body`, frozen at the triggering event, so after an edit the job kept replaying the body from the contributor's last push.

Re-running does not clear it, which is what makes this worse than a stale pass: a re-run replays the original event payload, stale body included. The failing run on #873 was attempt 3.

This is not historical. PR #937 is stuck on it right now, red on its current head over `notes/690-168-90.md`, a path that appears nowhere in its present body. Checking that PR's tree out and running the checker against the body it has today gives `prose check ok (5 file(s) checked)`, so the only thing between it and a green check is a rerun with the fresh body.

## The fix

`types: [opened, synchronize, reopened, edited]` on the trigger. `edited` also fires on title and base changes; the job is stdlib-only and takes about ten seconds, so that is a fair price.

A `concurrency` group comes with it, keyed on the PR number with `cancel-in-progress: true`. A burst of body edits should settle on the last one rather than race, so the reported status always reflects the current body.

Widening the trigger exposes nothing. The workflow is `pull_request`, not `pull_request_target`, holds only `contents: read`, and passes the body to the checker as a file through the environment rather than into a shell.

## Failure message

The issue also suggested the message does not say where an unresolvable path came from. It already does, via the `PR body` label, which is how both currently-failing PRs were diagnosed. What was missing is the remedy: the closing advice talked only about pointing at files that exist and never said to edit the body. `verify/check_prose.py` now adds a line when any problem carries that label, and names the label once as a constant so the reporter and the hint cannot drift apart.

The message #873 would have received:

```
  PR body: path does not exist in this tree
    notes/700-206-74.md

Anything labelled 'PR body' above is in the PR description, not in a
committed file: fix it by editing the PR body. Nothing in the tree has to
change for it.
```

## Checks

Five cases added to `verify/test_check_prose.py`: a body-sourced problem fails, is labelled, and points at the description; a file-only problem fails without blaming the body. Full suite is green at 163 passed, 6 skipped. `ruff check` reports no new findings on either changed file.

`verify/validator_manifest.json` is re-pinned, since both `.github/workflows/prose.yml` and `verify/check_prose.py` are in the trusted closure. Exactly those two hashes move.

## Not included

PR #872's prose failure is genuine rather than stale, but arguably a false positive of a different kind: its body passes `--out` a file under the system temp directory while describing the reproduction for the very flag it fixes, and the checker reads that as an absolute local path. Writing this section is what proved the point: quoting that path verbatim here failed the gate on this very PR. Left alone here; it deserves its own issue.

## Demonstrated on this PR

Editing this body after the checks first went green triggered a fresh prose run, which is the behaviour the issue reports as missing. On main the edit would have been silent.

### DIFF
--- a/.github/workflows/prose.yml
+++ b/.github/workflows/prose.yml
@@ -8,6 +8,22 @@ permissions:
 
 on:
   pull_request:
+    # `edited` is not in the default set (opened, synchronize, reopened) and it is
+    # the one that matters here: the checker reads the body from
+    # `github.event.pull_request.body`, which is the body as of the triggering
+    # event. Without it, editing a PR body to fix a cited path reruns nothing, and
+    # re-running from the PR page replays the same stale payload -- so a body-only
+    # failure could not be cleared except by an unrelated push (issue #897).
+    # `edited` also fires on title and base changes; the job is stdlib-only and
+    # takes about ten seconds, so that is a fair price.
+    types: [opened, synchronize, reopened, edited]
+
+# A burst of body edits should settle on the last one rather than race: the newest
+# run for a PR wins and any older one still in flight is cancelled, so the check
+# status always reflects the current body.
+concurrency:
+  group: prose-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   prose:
--- a/verify/check_prose.py
+++ b/verify/check_prose.py
@@ -39,6 +39,11 @@ import sys
 
 ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+# Label for problems that came from the PR description rather than a committed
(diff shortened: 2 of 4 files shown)


MathysRennela (approved):
LGTM!

Issue #942, opened 2026-09-08
Multiple cyclic generalized bicycle codes are overstated, codes/682-172-79.json distance 79 is over-stated: a weight-76 logical exists


EthanFeld:
# `[[682,172,d<=79]]` cyclic GB (`b6d86d076d1b5806`): witness improvable to d ≤ 76, 

**Entry:** `682-172-79.json` — "frontier cyclic GB b6d86d076d1b5806", generalized bicycle on Z₃₄₁, produced by `gf2_fast.dem_rand_witness` (2,000,000 trials, 211 s, side X).

## Summary

- The code itself checks out: `H_X H_Zᵀ = 0`, `rank H_X = rank H_Z = 255`, so `k = 682 − 510 = 172`.
- Both stored weight-79 witnesses are valid. The bound is simply not tight.
- **New witnesses of weight 76** on both the X and Z sides, verified. Attached as `682-172-76.json` (full schema-0.1 entry) and `682-172-76-witness-patch.json` (witness supports only).

## How the weaker witness was found

The random witness search operated on the raw 682-bit code. Working in the ring `R = F₂[x]/(x³⁴¹ − 1)` instead makes a much smaller search space visible.

Write the checks as `H_X = [A | B]`, `H_Z = [Bᵀ | Aᵀ]` with `a(x)`, `b(x)` the circulant symbols, and `f = x³⁴¹ − 1`. Three gcds:

```
deg gcd(a, b, f) = 86     ->  k = 2*86 = 172   (consistent)
deg gcd(a, f)    = 86
deg gcd(b, f)    = 96     <-  the asymmetry
```

The last line is the weak spot. A vector `(0 | w)` lies in `ker H_X` exactly when `b*(x)·w(x) ≡ 0 mod f`, where `b*(x) = b(x⁻¹)`. Those `w` form the cyclic code

```
C_b = < f / gcd(b*, f) >,   length 341, dimension 96
```

Dimension 96, not 86. The stabilizer part of `C_b` is only a 10-dimensional minimal ideal (the extra irreducible factor of `gcd(b*, f)` that does not divide `gcd(a, b, f)`), so all but a 2¹⁰ sliver of `C_b` maps to nontrivial logical classes.

[...]

PR #946, opened 2026-09-09
Structure-aware refutation for circulant generalized-bicycle codes

FarLab:
Refs #942. Implements the structure-aware refutation that issue suggests as a follow-up, in the C++ accelerator alongside the existing search.

## The idea

A generalized bicycle code is `H_X = [circ(a) | circ(b)]` over the ring of polynomials mod `x^L - 1`. That structure leaks distance. A logical operator whose support lies entirely in one block is constrained only by that block, so hunting one means searching the kernel of a single `L x L` circulant instead of the whole n-column kernel, with no stabilizer quotient to take.

On the `[[682,172]]` entry from the issue that is 96 dimensions instead of 427. It is why a second of this finds what two million general trials missed.

## Detect the structure, never trust the tag

The check runs on codes detected as circulant from the parity checks themselves. It never reads the submission's `family` field. That field is self-declared and unverifiable, and `verify/qldpc_verify.py` says so in as many words, so a gate keyed to it would be escaped by omitting it or writing something else. Detection requires every row of each half to be the cyclic shift of its first row, which a submitter cannot fake. Anything else reports block size 0 in microseconds and falls through to the general search untouched.

This also fixes a hole in the issue's own proposal. It suggests flagging codes where the two single-block gcd degrees differ, calling that gap the structural leak. But `674-170-76` is perfectly symmetric at 85, 85, 85 and still falls from 76 to 64. The asymmetry is not what makes the attack work, the reduction is, so the mechanism runs on every circulant GB code rather than only the asymmetric ones. Of the 56 circulant GB entries on the board, an asymmetry-keyed gate would have examined 20 and skipped the entry that lost the most.

## What was added

`circulant_gb_witness` in `verify/gf2_fast.cpp` reuses the existing randomized trial core unchanged. Only the kernel handed to that core is restricted, so the nontriviality test and the witness contract are the ones already in production. The new code is the detector, the single-block kernel construction, and the wrapper.

`verify/gate_changed.py` gains `_structural_refute`, wired in as one more entry in the mechanism dictionary the gate already collects, where a hit from any mechanism refutes.

## Soundness and cost

Sound on the same terms as the existing fast pass. The accelerator only proposes a weight, side and support. The find counts only after the pinned python stack confirms the support is a genuine nontrivial logical of that weight, lighter than the claim. An invalid or non-improving find is reported as a miss.

Strictly additive. Without the built extension the gate's output is byte-for-byte what it is today. I verified that by hiding the built module and re-running the gate, which printed the old label with no structural entry.

Cost is paid only where it helps. A non-circulant code returns immediately. The standard budget is 50,000 trials, the deep budget 400,000, which is seconds against the four minutes the gate already spends on a record claim.

## Staging

The gate now runs in two stages. Stage 1 is the structural search. On a hit it stops there, because a validated refutation cannot be undone by more searching and the general battery is where the wall-clock goes. On a miss, or on any code that is not circulant, stage 2 is the existing battery, unchanged.

The converse is deliberately not implemented. A stage 1 miss short-circuits nothing. That search only sees single-block logicals, and 52 of the 56 circulant entries on the board have a mixed-support witness it structurally cannot reach, so a code that survives stage 1 still owes the full battery.

| path | stage 1 | stage 2 |
|---|---|---|
| circulant, over-claimed | refutes | skipped |
| circulant, honest | clears it | runs in full |
| not circulant | zero trials, microseconds | runs in full |

Measured on the over-claimed `674-170-76` entry, the whole gate now finishes in 24 seconds. The battery it skipped would have cost about 44 minutes, from 332 microseconds per trial across 8,000,000 trials with setup measured separately.

Receipts record what actually ran rather than what was budgeted. Seeds, trials and budget go to zero on a short circuit, and a new field names the mechanism that ended the run, so a short method list reads as deliberate rather than truncated.

## Measured

All witnesses independently validated for kernel membership and nontriviality, by a checker written separately from the one the gate uses.

| code | claimed | found | time |
|---|---|---|---|
| 682-172-79 | 79 | 72 | 71 s |
| 674-170-76 | 76 | 64 | 17 s |
| 666-150-76 | 76 | 66 | 20 s |
| 390-68-28 | 28 | 26 | 4 s |

The 682 result is weight 72, four lighter than the 76 in the issue. At the standard gate budget all four still fall, the largest of them in under a second.

## Checks

Four new cases in `verify/test_gf2_fast.py` cover structural detection, the skip path, witness validity and determinism. Two more in `verify/test_refute_gate.py` cover the mechanism as the gate calls it, and skip cleanly when the extension is not built. Three more in `verify/test_refute_gate.py` pin the three staging paths, including the one that matters most, that an honest circulant code still pays the full battery. Full suite passes at 168 passed, 5 skipped.

`verify/validator_manifest.json` is re-pinned. The accelerator source and the gate are both in the trusted closure and exactly those two hashes move.

## Not included

The board is not corrected here. Four entries are now known to be over-stated and the witnesses exist, but changing them is a separate refutation submission and the issue's reporter may want to file their own. The weekly board sweep in `verify/refute_board.py` is also untouched, so the existing 56 circulant GB entries are not yet re-attacked in bulk. Both are worth doing next.


FarLab:
FYI @EthanFeld

MathysRennela (approved):
LGTM!

vprusso (approved):

PR #948, opened 2026-09-09
Bound syndrome decoder cross-check in refutation gate

DennisWayo:
## Summary

This bounds the optional BP+OSD syndrome-decoder cross-check in `verify/gate_changed.py` by running it in a child process with a hard wall-clock budget. If one decoder call stalls past the intended `max_seconds` window, the gate logs a note and continues with the existing RIS gates instead of letting the whole verify job run until GitHub cancels it.

The syndrome decoder remains fail-closed for real errors: missing output, child failures, and decoder exceptions still raise. The timeout case is treated like the optional decoder being unavailable after its budget is exhausted.

## Context

This is the verify-side companion to the merged prose rerun fix in #945. That PR fixes stale PR-body payloads for prose checks, but PR #937 also exposed a separate verify failure mode: the gate can enter `decode/distance.py` and then sit inside a single `BpOsdDecoder.decode(...)` call, so the decoder's internal `max_seconds` check is not reached.

Local probes against the #937 GB candidate reproduced that behavior: claims 86, 84, 82, 80, 78, and 76 each timed out after 90 seconds inside the syndrome-decoder path. Lowering the claimed distance alone therefore does not prevent the hosted verify job from being cancelled.

## Checks

- `UV_PYTHON=/opt/homebrew/bin/python3 /opt/homebrew/bin/uv run --offline --frozen python run_tests.py`
  - `145 passed, 8 skipped`
- `UV_PYTHON=/opt/homebrew/bin/python3 /opt/homebrew/bin/uv run --offline --frozen python -m pytest verify/test_refute_gate.py verify/test_gate_diff_classes.py`
  - `26 passed`
- `UV_PYTHON=/opt/homebrew/bin/python3 /opt/homebrew/bin/uv run --offline --frozen python verify/check_validator_integrity.py`
  - `ok: trusted validation stack matches the pin (27 files)`
- `UV_PYTHON=/opt/homebrew/bin/python3 /opt/homebrew/bin/uv run --offline --frozen ruff check --select I,F verify/gate_changed.py verify/test_refute_gate.py`
  - `All checks passed`
- `git diff --check`
  - clean

### DIFF
--- a/verify/gate_changed.py
+++ b/verify/gate_changed.py
@@ -55,7 +55,9 @@ Usage:
 """
 import glob
 import json
+import multiprocessing as mp
 import os
+import queue
 import secrets
 import subprocess
 import sys
@@ -66,13 +68,13 @@ sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 import circuit_tools as CT
 import gf2
 import heuristic_distance as H
+from build_receipt import make_receipt, write_receipt
 from circuit_verify import MAX_DEM_MECHANISMS, SIDE_FILES
 from qldpc_verify import file_size_error
 from validate_candidate import validate_candidate
-from build_receipt import make_receipt, write_receipt
 
 try:
-    import gf2_fast as GF          # optional C++ accelerator (make fast); the
+    import gf2_fast as GF  # optional C++ accelerator (make fast); the
 except ImportError:                # gate degrades to the python passes without it
     GF = None
 
@@ -81,6 +83,8 @@ ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # Fixed thread count for the fast pass so its verdict is reproducible from the
 # printed seed alone (the per-thread RNG streams depend on the split).
 FAST_THREADS = 4
+SYNDROME_SECONDS = 10.0
+SYNDROME_GRACE_SECONDS = 5.0
 
 
 def _load_syndrome():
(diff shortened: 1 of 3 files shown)


vprusso (commented):
The failure mode is real and the child-process approach is the right shape for it: an in-process budget cannot bound a call that never returns to check the budget, so moving it somewhere killable is the only thing that works. The fail-closed handling of child crashes and decoder exceptions is right, and `SYNDROME_SECONDS = 10.0` matches the existing default in `decode/distance.py:refute_check`, so the budget is being enforced rather than tightened. Worth saying explicitly in the description, since a reader might otherwise assume the cap is new.

One thing I would want changed before merge, and it is about the receipt rather than the timeout logic.

## The timeout is invisible in the artifact

On timeout the function returns `(False, None, None, 0)` and the caller stores it:

```python
results["syndrome-decoder"] = _syndrome_refute_bounded(doc, seed=seed + 1)
```

which then feeds

```python
gate = {..., "methods": list(results), ...}
```

So the receipt records `syndrome-decoder` among the methods that ran, with a clean negative, in the case where the check was killed and never completed. A later reader of that receipt cannot distinguish "the decoder ran and found nothing" from "the decoder was terminated after 15 seconds". The printed note goes to the job log, which is not the artifact anyone audits later.

This also does not match what the description says the change does:

> The timeout case is treated like the optional decoder being unavailable after its budget is exhausted.

Unavailability and timeout are handled differently in the code. When `SD` is None the key is never set (line 698 guards on `if SD is not None`), so `methods` correctly omits it. The timeout path sets the key with a negative result, so `methods` includes it. Those are opposite outcomes in the artifact.

Two ways to close it, either is fine:

1. Signal the timeout distinctly and let the caller omit the key, matching the unavailable path.
2. Record it, mirroring what the circuit tier already does. That tier carries `circ_notes` and a `failed` field into `gate["circuit"]` for exactly this purpose, so there is a house pattern to follow: a `gate["syndrome_decoder"] = {"ran": False, "reason": "timeout", "budget_seconds": ...}` would be consistent with it.

I would lean to 2, because "the optional check was skipped here" is a fact about the run that a receipt should carry, not one it should erase.

## Smaller points

`out.get_nowait()` immediately after `proc.join(...)` is tighter than it needs to be. A child that has put an item is not guaranteed to have had it drained into the parent's end at the moment `join` returns, and the failure is a fail-closed `RuntimeError("...produced no result")`, so the cost is a flaky gate failure rather than a wrong verdict. `out.get(timeout=grace_seconds)` would remove that.

`_syndrome_worker` itself is never exercised. Both tests inject fakes, which is the right way to test the bounding logic, but it means the code path that actually runs in CI has no coverage. A third test calling `_syndrome_refute_bounded` with the real worker against a small fixture would cover the wiring.

The import reordering in `gate_changed.py` and `test_refute_gate.py` is unrelated to the fix. Not a problem, but it makes the diff harder to read than the 96 lines suggest, and it puts style churn in the same commit as a change to the trusted validation stack, which is the one place where a reviewer most wants the diff to be only what it claims.

## On the manifest

`validator_manifest.json` is updated for the new `gate_changed.py` hash, which is right and easy to forget. Worth a line in the description confirming `check_validator_integrity.py` was rerun after the final edit rather than mid-review, since the hash pins the file exactly.

FarLab:
We will probably drop BP+OSD from the refutation gate, so closing this one. Thanks for working on this!

PR #955, opened 2026-09-09
Correct distance of codes/682-172-79.json to d=76

### LINKED ISSUE #942: Multiple cyclic generalized bicycle codes are overstated, codes/682-172-79.json distance 79 is over-stated: a weight-76 logical exists

EthanFeld:
Closes #942

`codes/682-172-79.json` claimed `d <= 79`. Issue #942 reports verified X- and Z-type logicals of weight 76.

Changes:

- migrate the entry to schema version 0.2;
- change the witness-backed bound from `d <= 79` to `d <= 76`;
- replace both distance witnesses with the reported weight-76 supports;
- record the correction in `provenance.notes`.

Verification:

- `uv run python verify/qldpc_verify.py codes/682-172-79.json` passes;
- both supports have weight 76 and pass the verifier's kernel and nontriviality checks.


FarLab:
@EthanFeld I think you should remove your handle from the authors list (distance refutation doesn't make you a coauthor of the code, this is an explicit check).

EthanFeld:
Sorry about that, got confused by the new schema

FarLab:
No worries, appreciate the PRs!

PR #969, opened 2026-09-09
Submit [[576,296,<=10]]: check-deletion of the board's [[576,294,12]]

MathysRennela:
## Code submission

- Parameters: [[576, 296, 10]] (upper bound)
- Tracks: weight-9plus x unrestricted (computed by the verifier)

### Checklist
- [x] `python verify/qldpc_verify.py codes/576-296-10.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` — checked, not equivalent: derived from `codes/576-294-12.json` by check deletion (different check sets, k differs).

### Correction after the CI refutation

The first revision of this PR claimed `[[576, 296, 12]]`. The trusted
refutation gate refuted it: its accelerated RIS pass returned a weight-10
X-type logical, support `[152, 215, 231, 297, 305, 353, 354, 421, 425, 514]`,
and the pinned GF(2) stack confirms that support commutes with `H_Z` and lies
outside `rowspace(H_X)`. The over-claim came from reading the check-deletion
bound's new-logical term as each deleted check's own weight (16): that weight
only upper-bounds the new logical coset, whose lightest representative is the
weight-10 logical. This revision claims the corrected witness-backed bound,
X `d <= 10` and Z `d <= 12`, so `d <= 10`; the weight-10 witness and its
`witness_provenance` are embedded in `codes/576-296-10.json`.

The weight-9plus x unrestricted cell. The code is a check-deletion derivative
of the board's `codes/576-294-12.json`: k rises from 294 to 296 exactly, while
the new logical coset exposed by the deletion caps the X distance at 10 rather
than the source's 12. It therefore trades k for d against its source rather than
dominating it, and the verifier reports the entry as advancing its cell
(`board_advancing: true`, no dominators). The full method, the correction, and
the dead ends are in `notes/576-296-10.md`.


FarLab:
REFUTED  codes/576-296-12.json [RIS-fast]: weight-10 logical < claimed distance 12
         witness = [152, 215, 231, 297, 305, 353, 354, 421, 425, 514]

PR #974, opened 2026-09-09
Submit [[34,4,3]]: r=1 graft of the board's [[36,4,3]]

MathysRennela:
## Code submission

- Parameters: [[34, 4, 3]] (upper bound)
- Tracks: weight-4 x local-2d-single (computed by the verifier)

### Checklist
- [x] `python verify/qldpc_verify.py codes/34-4-3.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` — checked, not equivalent: derived from `codes/36-4-3.json` by r=1 graft (2 stabilizer rows and 2 qubit columns removed); stated in `provenance.notes`.

The weight-4 x local-2d-single cell. The code strictly dominates its source entry
`codes/36-4-3.json` on n (36 -> 34) at equal k (4), equal
distance tier (d <= 3), and equal weight class. It is an r=1 graft
derivative: the move removes a qubit together with its unique stabilizer,
preserving commutation and k by rank arithmetic, with the d floor defended
by per-step multi-seed confirms. Passed the trusted validation gate. Full
method in `notes/34-4-3.md`.


MathysRennela:
@FarLab it seems that verify gets stuck at the pytest stage (verify/test_receipt.py). The same issue was encountered with other codes.

FarLab:
can you push an empty commit? otherwise the verify workflow will keep using the old code.

PR #978, opened 2026-09-09
Submit [[240,12,12]] graft as codes/240-12-12-graft.json (r=1 of [[242,12,12]])

MathysRennela:
## Code submission

- Parameters: [[240, 12, 12]] (upper bound)
- Track: weight-8 x local-2d-bilayer (computed by the verifier)
- Slug: `codes/240-12-12-graft.json`. The board already carries a `[[240,12,12]]`
  entry at `codes/240-12-12.json` — @vprusso's weight-6 2bga-coset code, no
  layout, i.e. a different cell (weight-6/unrestricted) and a different
  construction. The graft therefore ships under its own slug instead of editing
  an entry its author did not submit.

### Checklist
- [x] `python verify/qldpc_verify.py codes/240-12-12-graft.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` — checked, not equivalent: derived from `codes/242-12-12.json` by r=1 graft (2 stabilizer rows and 2 qubit columns removed); stated in `provenance.notes`. Distinct from the existing `[[240,12,12]]` entry, which is a weight-6 2bga-coset code in the unrestricted cell.

The weight-8 x local-2d-bilayer cell. The code strictly dominates its source entry
`codes/242-12-12.json` on n (242 -> 240) at equal k (12), equal
distance tier (d <= 12), and equal weight class. It is an r=1 graft
derivative: the move removes a qubit together with its unique stabilizer,
preserving commutation and k by rank arithmetic, with the d floor defended
by per-step multi-seed confirms. Passed the trusted validation gate. Full
method in `notes/240-12-12-graft.md`.


FarLab:
There seems to be already a 240-12-12 code (@vprusso submitted if I am not wrong) that this PR is overwriting. Is that correct?

If they are not the same, maybe rename it so it has a different filename (like 240-12-12b.json or something similar).

vprusso (commented):
@FarLab is right that this overwrites, and there is an established convention for the fix. But the substance of the submission is fine, so it is worth separating the two.

## The collision is real

`codes/240-12-12.json` exists on `origin/main` and this PR modifies that exact path. The two are different codes:

| | on main | this PR |
|---|---|---|
| author | @vprusso, 2026-07-31 | @mathysrennela |
| construction | 2BGA on dicyclic Dic_30, order 120 | r=1 graft of codes/242-12-12.json |
| family | 2bga-coset | bivariate-bicycle |
| check weight | 6 | 8 |
| layout | none, so unrestricted only | coordinates, interaction_radius, layers |

That also explains the 1,516 deletions, which is not a shape a genuinely new entry should have.

## The convention already exists

Three entries on the board already carry a suffix for exactly this situation: `144-8-16-b.json`, `666-150-76-b.json` and `96-12-4-mvb.json`. The first states the pattern in its own notes:

> Same [[n,k,d]] parameter set as the board's existing [[144,8,16]] (w=10 checks) but a different construction reaching it at weight 8.

which is the same relationship as here with the weights swapped. So `codes/240-12-12-b.json` plus `notes/240-12-12-b.md` follows precedent rather than inventing anything.

One thing to check when renaming: `check_prose.py` matches note slugs with `SLUG = re.compile(r"^(\d+)-(\d+)-(\d+)$")`, which does not match `240-12-12-b`. The three existing suffixed entries therefore skip the note-versus-filename `[[n,k,d]]` consistency check rather than failing it, so the rename will not break prose, but it does mean that particular check silently stops applying. Worth knowing rather than discovering later.

## The submission itself looks correctly scoped

Checking whether the existing entry sinks this one, since same n, same k, same d, and weight 6 against weight 8 would ordinarily dominate:

- In the unrestricted cells it does dominate. The weight-6 code is strictly better there.
- In `local-2d-bilayer` it does not compete at all, because it ships no layout and so cannot claim a locality cell.

This PR claims "the weight-8 x local-2d-bilayer cell", which is precisely the cell where the existing code is absent, and it dominates its own source `242-12-12` there on n at equal k, d and weight. So the frontier claim is right and correctly narrowed. I would suggest saying so explicitly in `provenance.notes`, something to the effect that the board's other [[240,12,12]] is weight 6 and dominates this one in the unrestricted cells, and that the contribution is confined to the bilayer cell it cannot enter. That is a stronger and more honest framing than "checked, not equivalent", and it pre-empts the obvious reviewer question.

## Summary

Rename to `240-12-12-b` per the existing convention, add the one-line comparison against the weight-6 entry in `provenance.notes`, and I think this is good to go. The graft method and the per-step multi-seed distance confirms are the parts I would not want to see relaxed, and they are the parts the note documents well.

vprusso (approved):

PR #985, opened 2026-09-10
Add [[300,8,13]] reduced single-layer planar bivariate-bicycle code

dorakingx:
## Code submission

- Parameters: [[n, k, d]] = [[300, 8, d ≤ 13]]
- Tracks: 2D-local single-layer / weight-6 (nests into single-layer weight-8 and any-weight, and into
  the bilayer and unrestricted cells at weight-6, weight-8 and any weight)

This is the last of the reduced planar bivariate-bicycle line — after #925, #935, #943 and #944 — and
the note says why the line stops: within the radius-4 single-layer layout, k <= 8 is forced at every
check weight and the polynomial pair is unique up to symmetry.

### Checklist
- [x] `python verify/qldpc_verify.py codes/300-8-13.json` passes locally
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`: checked, not equivalent.
      The gate reports no exact duplicate and no WL-equivalent entry. The board entries at n = 300 are
      [[300,4,27]] and [[300,60,14]], neither with a 2D layout and both differing in k and d, and no
      board entry has the unreduced size n = 338 at k = 8.

2D-local single-layer / weight-6: it joins the Pareto frontier at the low-n end. It is the smallest code
in that cell with k = 8 and d >= 13; the previous smallest was my [[373,8,15]]. Nothing in the cell
dominates it. It is **not** the cell leader by kd^2/n — at 4.507 it is fifth, behind [[454,8,17]]
(5.093), [[457,8,17]] (5.059), [[410,8,16]] (4.995) and [[450,8,16]] (4.551) — and it **dominates no
existing board entry**, so it adds a frontier point rather than displacing one. The claim is a
single-layer one: twelve merged entries dominate it in the unrestricted cell.

### How it was found
Open-boundary planar bivariate-bicycle code (arXiv:2504.08887), f = x + x^2 + y^2, g = 1 + x^2 y + x^2 y^2
on a 13 x 13 grid ([[338,8,<=13]]), reduced to 300 qubits by 38 removals of three kinds: 22 by
restricted r=1 lattice grafting (Sec. III E), 13 by weight-1 stabilizer cleanup (Sec. III D step 4,
which runs no distance search because it preserves k and d exactly by the CSS argument), and 3 by the
capped merge-graft introduced with [[454,8,17]] (#935). The single-layer layout is the one from
[[450,8,16]] (#912): qubit (site (i,j), block c) at (i + j, j - i + c), measured interaction radius
exactly 4 and spacing exactly 1.

Distance: a fresh-seed fast-RIS ladder, 20k -> 200k -> 1M -> 5M -> **20M** trials, flat at 13 throughout.
The ladder goes to 20M because 5M is demonstrably not enough in my work: a k = 10 candidate from a
sibling search held its value under three fresh seeds to 5M and then fell at 20M, which is the
correction in #981. All six of my merged entries have since been re-measured to 20M and all held.

The note also records the two results that close this construction. The layout map scales distances by
sqrt(2), so radius 4 confines every support to a diameter-2.83 disc, and over every monomial pair inside
that constraint — 3+3, 4+4, 3+5, 5+3 and the weight-7 splits, 220,000-odd pairs — the Newton mixed
volume never exceeds 8 and no split reaches k >= 10 at all. Of those, exactly 22 give a weight-6
radius-4 code, and the best realised distance slope among them is attained only by this family's
six-element symmetry orbit. Pairs with a higher *bulk* slope exist, but their boundary generators come
out at weight 14..26 and they realise far less.


FarLab:
Thanks for the code submissions! Can you merge your branch with the current main? We had a bug in the CI that is fixed now and so you should merge it with main to rerun with the new CI.

dorakingx:
The red `ris-compile` here is an infrastructure failure, not a problem with the submission.

The job dies in the CUDA toolkit install step, before it reaches any of this PR's files:

```
Err:9 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64  Packages
  File has unexpected size (1829834 != 1832398). Mirror sync in progress?
E: Failed to fetch .../Packages  File has unexpected size ...
##[error]Error: The process '/usr/bin/sudo' failed with exit code 100
```

`verify` and `prose` both pass on the same head. I cannot re-run the job from a fork, so if it needs
to be green please re-run it, or say the word and I will push an empty commit to trigger a fresh run.

PR #991, opened 2026-09-10
Add [[350,70,18]] lifted-product code

EthanFeld:
## Code submission

- Parameters: [[n, k, d]] = [[350,70,18]]
- Tracks: unrestricted / weight-9plus (computed by the verifier from H and the absence of a layout)
- Distance confidence: upper_bound (X: 18, Z: 18)
- Family tag: lifted-product (a self-declared filter, never used for ranking)

### Checklist

- [x] `uv run python verify/qldpc_verify.py codes/350-70-18.json` passes locally
- [x] Possible equivalence is noted in `provenance.notes`

The trusted preflight reports this as a new, board-advancing point in the unrestricted / weight-9plus cell. It is not dominated there on `(n, k, d, max check weight)` by the current board snapshot. No exact duplicate or WL-equivalent board entry was reported. Score upper bound: `k*d^2/n = 64.800`. The claim is `d <= 18`; exact distance is not asserted.

Research note: `notes/350-70-18.md`

Construction: `generic 1x2 group-algebra chain product over product(cyclic(5),dihedral(14)); A=[[6, 14, 24, 31, 33], [0, 29, 32, 38, 63]]; B=[[20, 36, 42, 53, 63], [0, 31, 55, 56, 69]]`


FarLab:
Did you use any LLM for this submission? If yes, can you add that to the model field? Thanks!

MathysRennela:
The references field is supposed to be for scientific papers, not to point at files of the repository. 
Can you correct that in this submission and the other recents submissions you've made?

EthanFeld:
Sorry about that, fixed

PR #998, opened 2026-09-11
Add [[270,5,13]], a reduction of the board's [[299,5,13]] with certified exact distance

dorakingx:
Adds `codes/270-5-13.json` with `notes/270-5-13.md`.

A reduction of @FarLab and @vprusso's `codes/299-5-13.json` by 29 qubits, at unchanged k,
unchanged weight class and unchanged locality cell. **The distance is certified exact, not
bounded.**

*(This PR previously carried [[276,5,13]]. The same pipeline, generalised, goes six qubits
further and the result is MILP-certifiable, so I replaced the file rather than leave a
dominated entry for review. Everything below describes what is in the branch now.)*

## One move

Let `S` be any element of the **row space** of `H_X` — an element of the stabilizer group,
not necessarily a published generator — with support `T`, and let `a` be in `T`. Replace one
generator of the subset summing to `S` by `S` (the span is unchanged, that generator being `S`
plus the rest of the subset); add `S` into every other X row meeting `a`; then apply the CNOT
fan-out from `a` to `T \ {a}`, which acts on the check matrices as `H_X[:, b] ^= H_X[:, a]`
and `H_Z[:, a] ^= H_Z[:, b]`.

Only `S` still meets `a`, so it becomes the weight-1 stabilizer `X_a` and every other X row is
untouched; and column `a` of `H_Z` becomes the XOR of its entries over `T`, which commutation
forces to be 0. Qubit `a` is disentangled and is deleted: `n → n-1`, `k` unchanged, and the Z
rows only **lose** an index, so their weights and diameters cannot rise.

The weight of `S` is what separates the regimes, entirely through the clearing step `R ^= S`:

- `|S| = 1` — nothing is added anywhere, so weight, radius **and distance** are exactly
  preserved and no distance search is needed. This is the row-space form of the weight-1
  cleanup of arXiv:2504.08887 Sec. III D step 4; `boundary_engine._cleanup` fires only on a
  *literal* weight-1 generator ROW, which is strictly stronger than row-space membership.
  Thirteen board entries carry such qubits. `codes/299-5-13.json` carries ten — a stride-13
  tail at originals 25, 38, 51, 64, 77, 90, 103, 116, 129, 142, each in zero X-checks.
- `|S| = 2` — the row loses `a` and toggles one other index, so its weight changes by 0 or −2.
- `|S| ≥ 3` — the row can grow, so each touched row is checked against the weight class and the
  locality radius. This generalises the capped merge-graft from #935, which only ever built `S`
  from the generators a single qubit happens to lie in.

Accepted here: `|S|=1`: 5, `|S|=2`: 8, `|S|=3`: 5, `|S|=5`: 5, `|S|=6`: 6.

## Verification

**Exact distance.** `scipy.optimize.milp` (HiGHS), one subproblem per logical basis element per
side — a vector of `ker(H_opp)` is a non-trivial logical exactly when it anticommutes with at
least one basis element, so k subproblems per side cover all `2^k − 1` classes. **All 10 proved
optimal at weight 13**, none hitting its time limit. A timeout would have been an upper bound
and is labelled as one; this is not a timeout.

**Corroboration.** Fresh-seed RIS ladder at 20k → 200k → 1M → 5M → 20M trials per rung,
independent seeds, both sides jointly, nothing lighter than 13.

**The exact stage, independently.** For each of the ten fixed qubits,
`rank([H_Z; e_q]) == rank(H_Z)` and the `H_X` column is empty; deleting the ten columns from the
original matrices gives k = 5, commuting checks, and the same two row spaces as the tool
produces. The intermediate [[289,5,13]] is separately MILP-certified exact at 13.

## Numbers

Max check weight 6 throughout. Interaction radius 3.1623 → 4.0000, inside the bilayer cap of 7.0
and in fact inside the single-layer cap of 4.0; the code still needs two layers, so it stays
`local-2d-bilayer`, layers unchanged at 2. Qubit positions are the source's — the reduction only
deletes. kd²/n 2.826 → 3.148. `verify/validate_candidate.py` passes with `board_advancing: true`,
`dominated_by: []`, no duplicate and no WL-equivalent entry.

A single-layer layout was attempted and failed: seeded from the code's own positions, ten
restarts of 600k anneal steps on a spacing-1 triangular lattice, best radius 4.3589 against a
cap of 4.0.

The k = 5 open-boundary family is @FarLab and @vprusso's; this is a reduction of their code.


dorakingx:
Updated the branch: this PR now carries **[[270,5,13]]** instead of [[276,5,13]].

Generalising the reduction move — graft a qubit away using *any* low-weight element of the stabilizer row space, rather than only the generators one qubit happens to lie in — takes the same source six qubits further, and at n = 270 the distance is small enough to **certify exactly**: 10/10 MILP subproblems proved optimal at weight 13, none hitting its time limit. The [[276,5,13]] version left one subproblem unproved at a 900 s limit, so it would have gone in as an upper bound.

Replacing the file rather than opening a second PR, since [[270,5,13]] dominates [[276,5,13]] on all four axes and I would rather not put a dominated entry in front of a reviewer. The title and description are updated to match what is in the branch.

FarLab:
No worries, the PR is already merged, so its fine to open a new PR with the new code.

dorakingx:
Superseded by #999, and closed by accident rather than by choice — GitHub closed this PR when I rewrote the fork branch, and it will not let me reopen it. Apologies for the noise.

#999 carries **[[270,5,13]]**, the same reduction of `codes/299-5-13.json` taken six qubits further by generalising the move, with the distance **certified exact** (10/10 MILP subproblems proved optimal at weight 13, none hitting its time limit) rather than bounded by a ladder. It dominates the [[276,5,13]] that was here on all four axes, so nothing is lost by reviewing #999 instead.

dorakingx:
Correction to my previous comment, which was wrong: this PR **merged**, it was not closed. I misread the API state field. Sorry for the confusion, and thank you for merging.

One thing needs saying plainly, because it is my mistake. The files that merged here are `codes/276-5-13.json` and `notes/276-5-13.md` — [[276,5,13]]. The title and description were edited to describe [[270,5,13]] shortly before the merge, while a force-push to my fork had not yet propagated to the PR head. **The merged code and its note are consistent with each other and correct**; it is the PR title and description that describe the wrong code. Nothing in `codes/` or `notes/` is inaccurate.

#999 adds [[270,5,13]], which is the same reduction taken six qubits further, with the distance **certified exact** (10/10 MILP subproblems proved optimal at weight 13). It dominates [[276,5,13]] on all four axes, so merging #999 retires this entry from the frontier. If you would prefer I open a PR that removes `codes/276-5-13.json` once #999 lands, say the word and I will.

PR #1001, opened 2026-09-11
Add [[196,15,11]], a further reduction of [[201,15,11]]

dorakingx:
Adds `codes/196-15-11.json` with `notes/196-15-11.md`.

A further reduction of `codes/201-15-11.json` — my own entry from #996, itself a reduction of
@vprusso's `codes/216-15-11.json`. Five more qubits, at unchanged k, unchanged weight class and
unchanged locality cell.

**Why anything was left.** #996 used a move set framed around a single qubit: remove a qubit lying
in exactly one stabilizer of a type, or merge the two or three it lies in until it lies in one. The
framing was the limit, not the code. The general form is:

Let `S` be **any** element of the row space of `H_X` — an element of the stabilizer group, not
necessarily a published generator — with support `T`, and let `a` be in `T`. Replace one generator
of the subset summing to `S` by `S` (the span is unchanged); add `S` into every other X row meeting
`a`; then apply the CNOT fan-out from `a` to `T \ {a}`, i.e. `H_X[:, b] ^= H_X[:, a]` and
`H_Z[:, a] ^= H_Z[:, b]`. Only `S` still meets `a`, so it becomes the weight-1 stabilizer `X_a`;
and column `a` of `H_Z` becomes the XOR of its entries over `T`, which commutation forces to be 0.
Qubit `a` is disentangled and is deleted — `n → n-1`, `k` unchanged, and the Z rows only **lose** an
index, so their weights and diameters cannot rise.

`|S|` picks the regime through the clearing step `R ^= S`: `|S| = 1` adds nothing anywhere, so
weight, radius **and distance** are exactly preserved with no search (the row-space form of the
weight-1 cleanup of arXiv:2504.08887 Sec. III D step 4, which `boundary_engine._cleanup` implements
only for a literal weight-1 generator ROW); `|S| = 2` changes a row's weight by 0 or −2; `|S| ≥ 3`
can grow a row, so each touched row is checked against the weight class and the locality radius.

Accepted here: `|S| = 2` once, `|S| = 3` once, `|S| = 4` twice, `|S| = 6` once — five qubits that no
single-qubit framing selects.

**Verification.** Fresh-seed RIS ladder at 20k → 200k → 1M → 5M → 20M trials per rung, independent
seeds, both sides searched jointly, nothing lighter than 11. In-loop screens are a filter, not
evidence. Distance is an upper bound; on this project a [[682,10,38]] candidate survived three fresh
seeds to five million trials and then fell at twenty million.

**Numbers.** Max check weight 8 throughout. Interaction radius 5.3852 → 6.0828, inside the bilayer
cap of 7.0; layers unchanged at 2 — an `|S| ≥ 3` graft can widen a check and is capped by the class
rather than by the source's radius, so the code is less local while staying in the same cell. Qubit
positions are the source's; the reduction only deletes. kd²/n 9.030 → 9.260. It dominates
`201-15-11`, `216-15-11` and `214-15-11` on (n, k, d, w). `verify/validate_candidate.py` passes with
`board_advancing: true` and `dominated_by: []`.

The construction is @vprusso's; @mathysrennela established the reduction-as-contribution precedent
with `codes/240-12-12-graft.json`.


MathysRennela:
@dorakingx Good that you caught one that I didn't. For future reference: the grafting technique is unpublished work. I am planning to publish it in the coming weeks.

FarLab:
@MathysRennela is your grafting technique related to this https://github.com/unitaryfoundation/qldpc-challenge/blob/main/research/local2d/boundary_engine.py ?

MathysRennela:
> @MathysRennela is your grafting technique related to this https://github.com/unitaryfoundation/qldpc-challenge/blob/main/research/local2d/boundary_engine.py ? 

Yes, but the result is about safe grafting, not a new grafting method in itself

PR #1004, opened 2026-09-11
Enforce connected Tanner graphs in verifier (#921)

### LINKED ISSUE #921: Enforce the connected Tanner graph rule in the verifier
Follow-up to #748, which agreed the rule and is closed. This tracks enforcing it.

## The rule

The Tanner graph of a submitted code (qubits and checks as vertices, edges from the combined check matrix) must not split into disconnected components. A literal direct sum `A + B` fails by construction.

## Why it needs its own issue

The rule is agreed but nothing enforces it. `verify/qldpc_verify.py` currently records `schema_valid`, `css_commutation`, `k_matches_claim`, the weight class, per-side witness validation and the refutation gate. There is no connectivity check, so a direct sum passes today.

## Suggested implementation

Union-find or BFS over the bipartite graph on `n` qubit vertices and `len(H_X) + len(H_Z)` check vertices, with an edge for every nonzero entry. Milliseconds at `n <= 700`. Record it alongside the existing checks so it appears in the report and fails the gate:

```
record("tanner_connected", ncomponents == 1,
       f"Tanner graph has {ncomponents} connected component(s)")
```

Two details worth deciding while implementing:

- Whether the check runs on the combined graph or per side. A code can have `H_X` splitting while the combined graph is connected, and the combined graph is the right object for the direct-sum question.
- What to do about existing board entries that would now fail. Grandfathering them silently would leave the board inconsistent with its own rule; flagging them in a sweep and correcting or annotating them is the honest option. Worth counting how many there are before choosing.

## Known evasion, from the #748 discussion

Conjugating the two blocks with a CNOT layer connects the graph at the cost of +1 check weight. The weight bump usually costs the cell, since weight classes nest downward. Detecting disguised sums by Clifford-equivalence or spectral component structure is the expensive backstop and should stay off the hot path, applied on suspicion rather than by default.

Raised as a pre-UnitaryCON concern in #748, so it is worth landing before submissions pick up.

MathysRennela:
## Summary

- enforce the combined X/Z Tanner graph connectivity rule in the trusted verifier;
- report the component count through the `tanner_connected` check, including isolated vertices;
- add regression coverage for connected, disconnected, isolated, and cross-side connectivity cases;
- add an explicit 34-entry migration warning set for legacy board entries in `verify_all.py`;
- document the connectivity requirement in `CONTRIBUTING.md`.

The hard verifier remains strict for every submission. The migration set affects only the all-board sweep and only when connectivity is the sole failure.

## Temporary migration window

The separate cleanup PR [#1005](https://github.com/unitaryfoundation/qldpc-challenge/pull/1005) removes the 34 legacy disconnected entries and the temporary exception. Because the site builder uses the hard verifier directly, merging this PR before #1005 will temporarily reduce the public board from 581 to 547 entries and remove those 34 detail pages. That transition is expected only until #1005 lands immediately afterward; #1005 is intentionally kept stacked on this PR until then.

## Validation

- `uv run pytest verify/test_verifier.py verify/test_verify_all.py verify/test_validator_integrity.py` — 37 passed
- `uv run python verify/check_validator_integrity.py` — passed
- `uv run python verify/verify_all.py` — 582/582 passed, with 34 explicit legacy warnings

Closes #921

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,6 +266,8 @@ accepted and recorded but will not sit on the frontier. See `TRACKS.md`.
 - Store `interaction_radius` as the exact measured max check diameter, not a
   rounded value.
 - Do not repeat a qubit index within a single check.
+- Submit one connected code, not a direct sum: the combined X and Z Tanner
+  graph must be a single connected component.
 - Use `provenance.origin` only for provenance (`baseline` vs `submission`), not
   novelty. If the same `[[n,k,d]]` parameter set exists in the literature, set
   `provenance.novelty` to `known_parameters` and cite it in
--- a/verify/qldpc_verify.py
+++ b/verify/qldpc_verify.py
@@ -203,6 +203,33 @@ def _vec(support, n):
     return v
 
 
+def _tanner_component_count(checks, n):
+    """Count connected components of the combined qubit/check Tanner graph."""
+    rows = checks["X"] + checks["Z"]
+    neighbors = [[] for _ in range(n + len(rows))]
+    for i, support in enumerate(rows):
+        check_vertex = n + i
+        for qubit in support:
+            neighbors[qubit].append(check_vertex)
+            neighbors[check_vertex].append(qubit)
+
+    seen = bytearray(len(neighbors))
+    components = 0
+    for start in range(len(neighbors)):
+        if seen[start]:
+            continue
+        components += 1
+        seen[start] = 1
+        stack = [start]
+        while stack:
+            vertex = stack.pop()
+            for neighbor in neighbors[vertex]:
+                if not seen[neighbor]:
+                    seen[neighbor] = 1
+                    stack.append(neighbor)
+    return components
+
+
 def verify(doc, refute=False, seed=None):
     """Verify a submission. If ``refute`` is set, run the distance refutation with
     ``seed`` -- when ``seed is None`` a fresh RANDOM seed is drawn, so the gate is
@@ -264,6 +291,13 @@ def _verify_semantic(doc, report, record, refute=False, seed=None):
                   if len(set(s)) != len(s)]
     record("no_repeated_qubits_in_a_check", not empty_rows,
            f"rows with repeats: {empty_rows[:5]}")
+
+    # The challenge rule applies to the combined X/Z Tanner graph. Count all
+    # qubit and check vertices, including isolated vertices, so a disconnected
+    # direct sum or an unused qubit cannot pass by construction.
+    ncomponents = _tanner_component_count(doc["checks"], n)
+    record("tanner_connected", ncomponents == 1,
+           f"Tanner graph has {ncomponents} connected component(s)")
 
     # 3. CSS commutation
     css = not bool(((HX @ HZ.T) % 2).any())
--- a/verify/test_verifier.py
+++ b/verify/test_verifier.py
@@ -15,7 +15,6 @@ import glob
 import importlib.util
 import json
 import os
-import subprocess
 import sys
 import tempfile
 import qldpc_verify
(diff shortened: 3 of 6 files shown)


vprusso (commented):
Checked the allowlist against the board rather than taking the count on faith: running this branch's verifier over all 581 entries in `codes/` gives exactly 34 failures, every one of them with `failed == ["tanner_connected"]`, and the set matches `LEGACY_DISCONNECTED` in both directions (nothing failing that is unlisted, nothing listed that passes).

One thing that came out of that sweep and is worth recording somewhere: not one of the 34 is disconnected because of an unused qubit. Every qubit is covered by some check row, so all 34 are genuine direct sums, with component counts running from 2 up to 12 (`192-24-4.json`). The isolated-vertex case is real for hostile input and the regression test for it is right to be there, but it does not occur on the current board.

Counting components over `checks["X"] + checks["Z"]` is the right object. A code can have a connected X graph and a connected Z graph and still be a direct sum, so the combined graph is the only version of the check that answers the question in #921.

The gap that matters for merge order: `verify_all.py` is not the only board-wide consumer of `verify()`. `site/build.py:1412` calls the same function and does `if not rep["ok"]: continue` with no message, and the orphan pruner at `site/build.py:3937` keys on the surviving slugs, so it deletes those detail pages too. Merging this PR on its own removes 34 entries and their 34 detail pages from the published board while `codes/`, `notes/` and the allowlist all still say they are present, and the sweep reports them green-with-warning. The migration exception covers one of the two consumers. Either give `site/build.py` the same exception for the interval, or state in the description that the public board goes from 581 entries to 547 at merge, so the count moving is not a surprise.

That interval is short if #1005 lands right after, since #1005 carries this exact commit plus the removals and deletes `LEGACY_DISCONNECTED` again. If the split exists so the enforcement logic gets reviewed apart from 34 file deletions, that is worth the scaffolding and I would keep it. It only bites if #1005 stalls.

The `test_verifier.py` switch from `glob.glob(codes/*.json)` to `git ls-files codes` is a coverage regression and I do not think this PR needs it. Untracked `codes/*.json` are now skipped, so a locally built disconnected code passes the self-tests silently, which is the exact case this PR exists to catch. The allowlist is keyed on the repo-relative path, so an untracked file is simply not in it and gets checked normally under the old glob. It also makes the self-tests require a git checkout rather than just a tree. Suggest reverting to the glob and keeping only the allowlist branch.

Last, and it is the part a submitter will actually feel: nothing contributor-facing states the rule. `CONTRIBUTING.md`, `README.md` and `TRACKS.md` do not mention connectivity, so the first person to submit a direct sum learns about it from a CI failure. The Tips section already carries the parallel line "Do not repeat a qubit index within a single check", which mirrors `no_repeated_qubits_in_a_check`. One line beside it would cover this:

- Submit one connected code, not a direct sum: the combined X and Z Tanner graph must be a single connected component.

MathysRennela:
I’ll restore glob-based discovery in `test_verifier.py` so untracked local submissions are still checked, add the connectivity rule to `CONTRIBUTING.md`, and update this PR description to make the temporary 581 → 547 site transition explicit while #1005 is waiting.

vprusso (approved):
All four points are addressed, and I re-checked the two that could have drifted rather than reading the diff for them.

The allowlist is still exact against current main, which is the one thing that could have broken while this sat open. Running this branch's verifier over all 593 entries on `origin/main` gives 34 failures, all `tanner_connected`, matching `LEGACY_DISCONNECTED` in both directions. The twelve entries that landed since your branch point are all connected, so nothing slipped past into the gap.

The glob restoration does restore the coverage, and I checked it by experiment rather than by eye: dropping an untracked disconnected code into `codes/` now fails the suite with `FAIL zz-probe.json verifies`, where under `git ls-files` it was silently skipped. That was the case worth protecting, since a locally built direct sum is how someone actually hits this.

One stale detail in the description. The 581 to 547 figures come from this branch's base; main is now at 593, so the transition is 593 to 559, and it will keep drifting while the PR is open. Either update the numbers or drop them and say the board loses the 34 legacy entries and their detail pages until #1005 lands, which stays true whatever main does.

Approving. The ordering is the only thing left and it is a merge-time decision rather than a code one: keep #1005 right behind this, since the interval is when the board is short 34 entries with the files still present.

PR #1005, opened 2026-09-11
Remove legacy disconnected board entries

MathysRennela:
## Summary

This is the cleanup follow-up to [#1004](https://github.com/unitaryfoundation/qldpc-challenge/pull/1004) and should be merged only after the enforcement PR.

- remove the 34 pre-existing board entries whose combined Tanner graphs are disconnected;
- remove the temporary legacy warning allowlist from `verify_all.py`;
- remove migration-only tests and restore the invariant that every remaining board entry passes the hard verifier;
- update the trusted-stack manifest.

The branch now includes the latest #1004 head. It remains draft until #1004 merges; after that merge, it can be retargeted/rebased so the final review diff contains only these cleanup changes.

## Validation

- `uv run pytest verify/test_verifier.py verify/test_validator_integrity.py` — 33 passed
- `uv run python verify/check_validator_integrity.py` — passed
- `uv run python verify/verify_all.py` — 548/548 passed

Depends on #1004.


vprusso (commented):
I verified the removal list independently rather than reading it, since a deletion of this size is the one place a wrong entry is hardest to notice afterwards. Ran union-find over the bipartite Tanner graph (n qubit vertices plus one vertex per X and Z check row, edge per nonzero) on every entry at `origin/main`, counting components that contain at least one qubit.

Both directions come out clean:

- All 34 entries in the removal list are genuinely disconnected. None is connected and slated for removal by mistake.
- Scanning all 566 entries on main finds exactly 34 disconnected, and they are the same 34. Nothing is missed, so the invariant the PR restores actually holds afterwards.

That matches your `verify_all.py` 548/548 and is the result I would want confirmed before a 71k-line deletion lands.

## One gap in the cleanup

The PR removes `codes/*.json` but leaves the research notes. 25 of the 34 have one:

```
notes/112-8-5.md  notes/128-12-8.md  notes/144-16-8.md  notes/144-24-6.md
notes/162-36-4.md notes/176-22-4.md  notes/192-24-4.md  notes/192-6-8.md
notes/214-15-11.md notes/242-2-11.md notes/261-16-12.md notes/288-32-8.md
notes/288-48-6.md notes/324-4-9.md   notes/338-2-13.md  notes/360-16-14.md
notes/360-24-10.md notes/360-32-6.md notes/484-4-11.md  notes/676-4-13.md
notes/78-6-5.md   notes/88-8-7.md    notes/90-6-5.md    notes/96-12-4-mvb.md
notes/96-12-4.md
```

Unlike the detail pages these are not self-cleaning. `site/build.py` prunes orphan `docs/codes/*.html` explicitly (the loop at the end of the page-writing block removes any HTML whose slug is not in `slugs`), so the 34 site pages go on the next rebuild without anything further. There is no equivalent pass for `notes/`, and `build.py` links notes by `notes/<slug>.md` from the entry, so an orphaned note is simply unreferenced rather than broken. It will not fail anything; it will just sit there describing a code the board no longer has.

Two defensible options and I do not have a strong preference:

- Delete them with the codes, which keeps `notes/` in one-to-one correspondence with `codes/` and matches what the detail-page pruning does.
- Keep them deliberately and say so in the PR body. There is an argument for this: the notes are the only record of how those codes were found, and "this construction produces disconnected codes" is itself a useful negative result for the next person who tries it.

What I would not do is leave it unstated, because a later reader cannot tell whether the notes were kept on purpose or missed.

## On the remedy, since I raised the question in #921

When I filed the enforcement issue I flagged the open question as correcting or annotating versus removing, and said the count should be established before choosing. The count is 34 of 566, about 6 percent, and the authorship spread is worth recording:

| author | entries |
|---|---|
| @mathysrennela | 16 |
| @FarLab, @vprusso | 7 |
| @cbjuan | 7 |
| @vprusso | 2 |
| @FarLab | 2 |

The PR author is removing more of their own entries than anyone else's, which is the right way for a cleanup like this to be proposed. @cbjuan also anticipated this in #748 ("I have few disconnected codes that would be frontier"), so none of the affected contributors is being surprised.

Given that spread I think removal is the right call over annotation. An annotated entry still occupies a frontier cell unless the ranking also learns to skip it, which is a second mechanism to build and keep correct; deletion needs no ongoing enforcement beyond #1004.

## Ordering

The body says merge only after #1004 and that this branch should be rebased so the final diff is cleanup only. Worth holding to strictly: if this lands first, the board spends the interval with the entries gone but nothing preventing new ones, which is the worst of both states during UnitaryCON submissions.

With the notes question answered either way, this looks good to me.

Issue #648, opened 2026-08-20
Show locality.contributed_by layout credit on the site


willzeng:
Follow-up from the #643 review: layout credit now lives in `locality.contributed_by` (by design — it must not touch `provenance.authors`), but `site/build.py` doesn't read the field, so a contributor who retrofits a layout onto someone else's entry gets a JSON record and nothing on the page.

Since the layout-filling effort (filling the ~261 missing geometric-efficiency scores) is the reason the binding exists, and visible credit is most of the incentive, the site should surface it — e.g. on the code page next to the layout/locality details ("layout contributed by @handle, <date>"), and plausibly in the contributor leaderboard's counting.

Nothing besides `verify/check_authorship.py` reads `contributed_by` today, which is deliberate for the security property (it confers no authority); display is read-only and doesn't change that.

PR #1012, opened 2026-09-11
Bind circuit-tier contributions to their contributor

FarLab:
## The gap

The circuit tier (RFC 0001, #505) has no path for anyone but a code's listed authors to contribute syndrome-extraction circuits to an entry.

`verify/check_authorship.py` admits a non-author through exactly two alternative bindings, and a `circuit` block fails both:

- **refutation binding** — `field 'circuit' changed (only the distance claim may change)`
- **layout binding** — `no locality block was added`

`circuit` also had `additionalProperties: false` and no credit field, so there was nowhere to record who built a schedule even if the gate allowed it.

Of the 17 entries carrying circuits today, 14 are literature baselines that land only because no-`@handle` entries are exempt from the check. A third party could not contribute a schedule to the three that are submissions (`45-5-4`, `50-6-4`, `66-5-5`).

## What this does

**Schema.** `contributed_by` on `circuit`, sharing a new `$defs/contributedBy` with `locality.contributed_by` (same `{by, date, method?}` shape). No new `schema_version` conditional: the whole `circuit` block already requires 0.2.

**Gate.** `layout_binding` generalizes to `addition_binding(field, noun)`; `circuit_binding` is its second instance. A change that ONLY adds a first `circuit` block, crediting the PR author in `circuit.contributed_by.by`, binds. Anything that also moves `checks`, `distance`, `authors`, `model`, or adds a layout does not. `provenance` must be byte-identical apart from a notes append, so a contributed tier never changes the author list.

Replacing an existing `circuit` block stays with the listed authors, which matters more here than for layouts: the tier is penalty-only (`d_circ <= d`), so a donated schedule can *lower* an entry's score where a donated layout can only sharpen its locality class. It is still a true fact about the code — the same principle that lets a non-author refute a distance — and reserving replacement leaves a mediocre donated schedule the authors' to beat.

**`circuits/<slug>/` is part of the claim surface.** The gate now folds a diff there onto `codes/<slug>.json`, the same mapping `gate_changed.py` already prices the circuit search on. Without this the new binding would be decorative: the gate diffs only `codes/`, so someone's committed `.stim`/`.dem` files could be swapped with their JSON left untouched and nothing would check authorship at all. That case now reports `the JSON is unchanged, so the diff is in this entry's committed circuits/ artifacts; replacing those is reserved to its listed authors`.

The gate still does no circuit math — witnesses, the canonical noise recipe and per-TICK parallelism stay `circuit_verify.py`'s job, and `gate_changed.py` re-runs the circuit refutation on any changed claim surface, so a donated tier cannot under-claim `d_circ` without a valid witness at that weight.

## Tests

- 15 circuit-binding cases mirroring the layout ones (credit present/absent/wrong, authors, model, checks, distance, notes, mixed circuit+layout diff, replacement by non-author vs. author) and one merged-state escalation case: a merged circuit binding must not widen `@bob`'s rights on the next PR.
- 3 cases for the `circuits/` claim surface, including a bound circuit addition that also tampers with a *different* entry's circuits.
- `site/test_security.py`'s free-form-string inventory follows `method` to the shared `$def` and now carries an XSS payload through `circuit.contributed_by.method`.

Full suite green (`verify` 119 passed / 10 skipped, `site` 9, `research` 15). All 585 board codes still validate against the schema.

`verify/validator_manifest.json` is re-pinned for the two files it tracks (`schema/code.schema.json`, `verify/check_authorship.py`) — that diff is two hash lines and wants a look.

Verifier-only change, no `codes/` edits, per the trust-boundary rule in CONTRIBUTING.


vprusso (commented):
Read this against the layout binding it generalizes, and poked at `circuit_binding` directly rather than only through the test fixtures. The mechanism is sound and I think it should land. Four things, one of which I would rather see resolved before merge.

What I checked and found clean, so nobody re-does it: both manifest hashes recompute exactly against the branch. A bound circuit addition cannot claim a `d_circ` with no committed artifacts, because `_circuit_refute` opens `circuits/<slug>/memory_*.stim` and a miss lands in `circ_failed`, which increments `failed` and prints FAIL closed. The `schema_version` escape hatch in `addition_binding` is not exploitable: the enum is 0.1/0.2 and all three `allOf` conditionals force the version up from a feature, never down, so a non-author cannot re-interpret an entry's existing claims through it. A `circuit` block carrying only `contributed_by` does bind at this gate, but `required: [d_circ, rounds, stim_version]` rejects it one layer down. The cross-entry tamper case does what the test says: the other slug arrives through `from_circuits`, its JSON is unchanged, and it takes the `doc == base_doc` violation.

The thing I would resolve first is that the credit is invisible. #643 gave layouts the binding and #648 gave them the render, and this PR is the binding half only. `site/build.py` reads `locality.contributed_by` in three places: the `layout_cb` entry field at 1449, the credit line on the detail page at 2132, and the contributor modal at 2489, where a layout earns its contributor a row and the `g` credit. Nothing anywhere reads `circuit.contributed_by`. So a circuit contributor passes the gate, their handle lands in the JSON, the row gets a gear chip with someone else's name beside it, and they get no row of their own. That is the part of the contribution path a contributor actually experiences. Either the three site hooks come along here or there is an issue tracking them, but shipping the gate alone means the first person to use it gets bound and not credited.

Second, the penalty-only rationale does not describe the current board, and it is doing real work in the design (it is the stated reason replacement is reserved). `d_circ` does not enter `pareto()` or `cells_by_key()`; `eff` is `k*d*d/n` off `distance.d`. It is a display chip, `site/build.py:3759`. A layout is the donated artifact that does move standing: it sets the locality class, which decides which Pareto cells the entry competes in, and it generates `geo`. So today the asymmetry runs the other way from the one in the description. Reserving replacement is still defensible, and it is the right call if `d_circ` ever ranks, but I would not justify it by a score effect the board does not currently have.

Third, `code_for_circuit_path` duplicates `gate_changed.map_changed` (lines 105 to 116) exactly. Both files are pinned in the manifest and both now define what an entry's circuit claim surface is. Neither comment points at the other, so if one gains a case, a nested directory say, the authorship gate and the refutation gate disagree about which entry a file belongs to and nothing catches it. A cross-reference comment in each would be enough; a shared helper would be better but `gate_changed` pulls stim, so importing it here is not free.

Fourth, small: the `doc == base_doc` branch asserts the diff is in `circuits/`, but a whitespace-only or key-reordered JSON edit reaches it too and gets told the wrong reason. The rejection stays correct either way, so this is only the message.

One note in passing, inherited rather than introduced: a binding non-author may rewrite `name` wholesale, not just append to it. I confirmed it binds. That came in with the layout binding and the site escapes the value, so it is not new, but this PR doubles the number of routes that reach it and it may be worth restricting to an append the way `provenance.notes` is.

MathysRennela (approved):
LGTM!

vprusso (approved):

PR #1014, opened 2026-09-11
Add [[372,8,15]], @mathysrennela's [[392,8,15]] reduced and de-stacked onto one layer

dorakingx:
Adds `codes/372-8-15.json` with `notes/372-8-15.md`.

@mathysrennela's `codes/392-8-15.json` reduced by 20 qubits **and moved from two layers onto one**.
Neither half works without the other, and the reduction has to be held back for the de-stack to be
possible at all.

## The move

Let `S` be **any** element of the row space of `H_X` — an element of the stabilizer group, not
necessarily a published generator — with support `T`, and let `a` be in `T`. Replace one generator
of the subset summing to `S` by `S` (the span is unchanged); add `S` into every other X row meeting
`a`; then apply the CNOT fan-out from `a` to `T \ {a}`: `H_X[:, b] ^= H_X[:, a]` and
`H_Z[:, a] ^= H_Z[:, b]`. Only `S` still meets `a`, so it becomes the weight-1 stabilizer `X_a`, and
column `a` of `H_Z` becomes the XOR of its entries over `T`, which commutation forces to 0. Qubit
`a` is disentangled and is deleted: `n → n-1`, `k` unchanged, Z rows only lose an index.

`|S| = 1` preserves weight, radius and **distance** exactly with no search; `|S| = 2` changes a row's
weight by 0 or −2; `|S| ≥ 3` can grow a row. Accepted here: 4, 5, 5, 1, 1, 4 for `|S| = 1..6`.

## The de-stack is 2-SAT

The source has two qubits per site on a spacing-1 square lattice. This code has one, via
`(site (i,j), slot c) → (i+j, j−i+c)` — injective, since a collision needs `i+j = i'+j'` and
`j−i = j'−i'+1`, whose sum gives `2j = 2j'+1`. A check pair at site displacement `(a,b)` lands at
`(a+b, b−a+dc)` with `dc = c_u − c_v`, and the dependence on `dc` is **not symmetric in its sign**:
`(±2,±2)` lands at `(±4, dc)` so both `dc = ±1` break the cap and the two qubits must share a slot,
but `(2,−2)` lands at `(0,∓4+dc)` and `(2,−1)` at `(1,−3+dc)`, each broken by only one sign.
Equalities mixed with implications is 2-SAT over one boolean per site, solved with the implication
graph and Tarjan's SCC. Satisfiable here: realised radius **exactly 4.0000**, one qubit per site,
minimum site spacing **exactly 1.0000**.

For every reduction of this source that used the full bilayer slack, the instance is
**unsatisfiable** — 187 to 189 sites forced both ways, a proof rather than a failed search. An
`|S| ≥ 3` graft buys a qubit with check diameter, and check diameter is what pays for a single layer.
So the reduction is capped at the source's own check diameter of 2.8284 rather than at the bilayer
cap of 7.0, with the 2-SAT condition enforced after every accepted graft. That converges at n = 372,
which is what this needed: my own `codes/373-8-15.json` is in the same cell.

## Verification

Fresh-seed RIS ladder at 20k → 200k → 1M → 5M → 20M trials per rung, independent seeds, both sides
jointly, nothing lighter than 15. In-loop screens are a filter, not evidence. Distance is an upper
bound; on this project a [[682,10,38]] candidate survived three fresh seeds to five million trials and
then fell at twenty million. A MILP certification is running and I will report what it says either way.

`verify/validate_candidate.py` passes: `board_advancing: true`, `dominated_by: []`, no duplicate and
no WL-equivalent entry, cell `weight-6 × local-2d-single`.

## Numbers

Max check weight 6 throughout. Interaction radius 4.0000 on one layer against 2.8284 on the source's
two. kd²/n 4.592 → 4.839. It dominates `392-8-15` and my own `373-8-15` on (n, k, d, w).

The construction is @mathysrennela's; the single-layer sublattice map is the one I introduced with
[[450,8,16]] (#912).


dorakingx:
Reporting the MILP as promised, including the part that did not finish.

**15 of the 16 subproblems proved optimal at weight 15**; one (Z side, logical 4) hit its 2400 s limit with an incumbent of 15 and no proof. A timeout is not a proof, so the distance in `codes/372-8-15.json` stands exactly as submitted — a witness-backed upper bound of 15, corroborated by 20M fresh-seed RIS trials — and I am not upgrading the confidence field.

For what it is worth as evidence rather than as a claim: every subproblem that did finish came back at exactly 15, nothing lighter was found by any method, and the one that timed out never found anything below 15 either. I have restarted that single subproblem with a 4-hour limit and will report whichever way it lands. If it ever returns something lighter than 15 I will open a correction PR immediately.

dorakingx:
Correcting my own comment above, which overstated the MILP. I wrote "15 of the 16 subproblems proved optimal" from one timeout line in the log plus an assumption, instead of reading the table. The actual state of that run:

| side | subproblem | weight | proved optimal |
|---|---|---|---|
| X | 0, 1, 2, 5, 6, 7 | 15 | yes |
| X | 3, 4 | 15 | **no — 2400 s limit** |
| Z | 0, 1, 2, 3 | 15 | yes |
| Z | 4 | 15 | **no — 2400 s limit** |
| Z | 5, 6, 7 | — | **never ran** |

So it is **10 of 16 proved, 3 timed out, 3 not reached** — not 15 of 16. Every subproblem that ran returned an incumbent of exactly 15 and nothing lighter was found by any method, but that is evidence, not a proof, and the six unresolved ones are unresolved.

Nothing about the submitted file changes: `codes/372-8-15.json` claims 15 as a witness-backed upper bound corroborated by 20M fresh-seed RIS trials, which is what it says, and I am not touching the confidence field. The run has been restarted with a 4-hour per-subproblem limit and I will report the complete table whichever way it lands — and open a correction PR immediately if anything lighter than 15 turns up.

Apologies for the noise; I would rather post the correction than leave the wrong number standing.

PR #1015, opened 2026-09-11
Add single-layer 2D-local layouts for three bilayer entries

dorakingx:
Moves three entries from `local-2d-bilayer` to `local-2d-single`. Only `locality` changes — the
codes, their distances and `provenance.authors` are untouched, and the layout credit goes in
`locality.contributed_by`, as #674/#675 did for the bilayer layout batches.

All three are Liang, Eberhardt and Chen's open-boundary codes (arXiv:2504.08887), which carry no
`@handle` authors.

| entry | layers | radius | in-cell status as single-layer |
|---|---|---|---|
| `128-8-6` | 2 → 1 | 2.8284 → 4.0 | **on the frontier**, dominated by nothing |
| `288-8-12` | 2 → 1 | 2.8284 → 4.0 | **on the frontier**, dominated by nothing |
| `200-8-9` | 2 → 1 | 2.8284 → 4.0 | dominated by `198-8-9` if that one is ever de-stacked |

## The map, and why it is a decision problem rather than a search

Each of these sits on a spacing-1 square lattice with two qubits per site. De-stack by

> qubit at (site `(i,j)`, slot `c`) → `(i+j, j−i+c)`

— a 45° rotation scaled by √2, with the slot separating co-sited qubits by one unit. It is
injective: a collision needs `i+j = i'+j'` and `j−i = j'−i'+1`, whose sum gives `2j = 2j'+1`. Every
site then carries exactly one qubit at spacing exactly 1.

Whether the result fits the radius cap is **not** simply √2 times the bilayer radius. A check pair
at site displacement `(a,b)` lands at `(a+b, b−a+dc)` with `dc = c_u − c_v ∈ {−1,0,1}`, and the
dependence on `dc` is not symmetric in its sign:

- `(±2,±2)` with `a+b = ±4` lands at `(±4, dc)` — **both** `dc = ±1` break the cap, so the two
  qubits must share a slot;
- `(2,−2)` lands at `(0, ∓4+dc)` and `(2,−1)` at `(1, −3+dc)` — each broken by only **one** sign.

Equalities mixed with implications is **2-SAT** over one boolean per site (does this site swap its
two slots), solved with the implication graph and Tarjan's SCC. The board has **73** entries that
are two-per-site on an integer lattice; **7 are satisfiable and the other 66 are unsatisfiable**,
each with dozens of sites forced both ways — a proof that this map cannot de-stack them, not a
search that gave up.

I first wrote this as a union-find with parity, treating every critical pair as an equality. It was
wrong, and it declared a layout feasible that then measured 4.1231. The tool now re-verifies every
clause against the returned assignment, so a solver bug reports itself rather than passing as a
result — which is how I caught a second error, a backwards polarity in reading Tarjan's
reverse-topological component numbers.

## The other four, and why they are not here

Four more satisfiable entries are **not** in this PR: `198-8-9` and `392-8-15` (@mathysrennela), and
`45-5-4` and `50-6-4` (@FarLab, @vprusso). `check_authorship.py` is right to refuse them — the
`contributed_by` binding covers a *first* layout, and replacing an existing one is reserved to the
entry's listed authors. My first push included them and CI caught it.

`198-8-9` is the one that matters: de-stacked it lands on the single-layer frontier, dominated by
nothing, and it would then dominate `200-8-9` in this PR. @mathysrennela — the assignment is a
plain 2-SAT solve on the published coordinates and I am happy to hand over the layout or the tool
if you want it in.

Each file here was re-run through `verify/qldpc_verify.py`: `ok: true`, `locality_class:
local-2d-single`, one qubit per site, minimum site spacing exactly 1.0000.


dorakingx:
Corrected a count in the description: the board has **73** two-per-site integer-lattice entries, of which **7 are 2-SAT satisfiable (these) and 66 are unsatisfiable**. The first version said 21 unsatisfiable, which was me reading off a truncated scan rather than the scan's own total. Nothing about the seven layouts changes.

vprusso (commented):
The de-stacking map is correct and the 2-SAT framing is the right shape for the question. I checked the geometry arithmetic and every domination claim in the table; all seven domination rows are right. Two things block merge as it stands, and one row of the table is a worse trade than the description lets on.

The map is better than injective, it is a bijection onto the integer lattice. Given a target `(u,v)`, `c = (u+v) mod 2` is forced, and then `j = (u+v-c)/2` and `i = (u-v+c)/2` are forced too. That is why every site ends up carrying exactly one qubit at spacing exactly 1, which is a stronger statement than the parity argument in the description needs to make. The sign asymmetry is right as written: `(±2,±2)` lands at `(±4, dc)` with squared length `16 + dc^2`, so only `dc = 0` survives and the pair is an equality; `(2,-2)` lands at `(0, -4+dc)` and only `dc = -1` breaks it. Worth naming where the 4.0 comes from, since the description just calls it the cap: it is the `local-2d-single` radius threshold in `verify/qldpc_verify.py:437`, not a target you chose.

The blocker is authorship. CI fails, and it fails on exactly the case the layout binding reserves: `check_authorship.py` admits a non-author only for a FIRST `locality` block, and these seven all have one. Four are rejected outright (`198-8-9` and `392-8-15` to @mathysrennela, `45-5-4` and `50-6-4` to @FarLab and me). The other three pass only because `128-8-6`, `200-8-9` and `288-8-12` are Liang/Eberhardt/Chen baselines with no `@handle`, so they take the baseline exemption rather than binding to you. That exemption is the thing #674/#675 relied on, which is why those batches landed and this one does not. This needs the listed authors to open it or to co-author, and #1012 is currently restating that same reserve for circuits, so I would not want to route around it here.

The author table needs fixing before anyone judges the PR from it. `128-8-6` and `200-8-9` are Liang/Eberhardt/Chen baselines, not mine; `198-8-9` is @mathysrennela, not mine; `45-5-4` and `50-6-4` are @FarLab and me, not me alone. Four of seven rows are wrong about whose entry is being edited, which is the one column a reader uses to decide whether to object.

On the geometry, something the description should say outright: for five of the seven the geometric efficiency is numerically unchanged. `geo = 4kd^2/(n rho^2 r^4)`, and `2*sqrt(2) -> 4` is exactly the factor that cancels `rho^2 = 4`, since `(4/(2 sqrt 2))^4 = 4`. Computed from the coordinates either side, `128-8-6`, `198-8-9`, `200-8-9`, `288-8-12` and `392-8-15` all come out at ratio 1.000. So "radius 2.8284 -> 4.0" reads like a cost in the table and is in fact free; the whole gain for those five is the locality class. `45-5-4` genuinely improves, `rho^2 r^4` going 324 to 169 for a 1.917x gain.

`50-6-4` is the row where the trade is negative and the table does not say so. `r` goes `sqrt(5) -> sqrt(13)` while `rho` goes 2 to 1, so `rho^2 r^4` goes 100 to 169 and geo drops from 0.07680 to 0.04544, to 59 percent of its value. It stays mid-pack either way (rank 171 of 248 before, so no record is lost), but it is the one entry in the batch that is strictly worse off on a number the board displays, and "dominated by 36-6-4" does not communicate that. As a listed author on it I would rather it came out of the batch, or the description said plainly what it costs.

Scoping what I did not check: I verified the seven layouts and the arithmetic above, not the 73/7/66 census. The corrected count in your comment is the sort of claim that only the tool can settle, and re-running it is not what decides this PR.

dorakingx:
Dropped four of the seven and reduced this to three entries, after `check_authorship.py` correctly refused them.

The rule I missed: `locality.contributed_by` binds a **first** layout, and replacing an existing one is reserved to the entry's listed authors. `198-8-9` and `392-8-15` (@mathysrennela) and `45-5-4` and `50-6-4` (@FarLab, @vprusso) all already have layouts, so they are not mine to replace. What remains are the three Liang–Eberhardt–Chen entries, which carry no @handle authors.

I also had the author column wrong in the first description — `128-8-6`, `200-8-9` and `288-8-12` are Liang, Eberhardt and Chen's, not @vprusso's. Fixed.

Of the four dropped, `198-8-9` is the one worth having: de-stacked it sits on the single-layer frontier dominated by nothing, and it would dominate the `200-8-9` in this PR. @mathysrennela, it is a plain 2-SAT solve on your published coordinates — happy to hand over the assignment or the tool.

MathysRennela:
> Dropped four of the seven and reduced this to three entries, after `check_authorship.py` correctly refused them.
> 
> The rule I missed: `locality.contributed_by` binds a **first** layout, and replacing an existing one is reserved to the entry's listed authors. `198-8-9` and `392-8-15` (@MathysRennela) and `45-5-4` and `50-6-4` (@FarLab, @vprusso) all already have layouts, so they are not mine to replace. What remains are the three Liang–Eberhardt–Chen entries, which carry no @handle authors.
> 
> I also had the author column wrong in the first description — `128-8-6`, `200-8-9` and `288-8-12` are Liang, Eberhardt and Chen's, not @vprusso's. Fixed.
> 
> Of the four dropped, `198-8-9` is the one worth having: de-stacked it sits on the single-layer frontier dominated by nothing, and it would dominate the `200-8-9` in this PR. @MathysRennela, it is a plain 2-SAT solve on your published coordinates — happy to hand over the assignment or the tool.

@dorakingx  Feel free to add yourself as co-author, that's ok with me.

vprusso (approved):

Issue #1024, opened 2026-09-12
Two d_circ=4 entries measure 10x apart in logical error rate: 32-8-4 against 16-2-4


vprusso:
Two entries on the circuit tier have the same distance, the same `d_circ`, and the same round count, and their measured logical error rates are a factor of ten apart. The tier exists to make exactly this visible, so it is worth recording as a finding and as a diagnosis task.

Measured with the pinned decoder `bposd-cs-10` at the canonical reference rate, 10,000 shots for `32-8-4` and 60,000 for `16-2-4`, both bases:

| entry | d | d_circ | rounds | failures / shots | p per shot | LER per round |
|---|---|---|---|---|---|---|
| `codes/16-2-4.json` | 4 | 4 | 4 | 413 / 60,000 (X), 408 / 60,000 (Z) | 6.9e-3 | 1.73e-3 |
| `codes/32-8-4.json` | 4 | 4 | 4 | 632 / 10,000 (X), 590 / 10,000 (Z) | 6.3e-2 | 1.66e-2 |

With 632 and 413 failures behind them the gap is not a sampling artefact. For scale, `14-6-2`, a distance-2 entry, measures 5.9e-2 per shot, so `32-8-4` has a higher per-shot failure rate than a code with half its distance. Per round it comes out better than the distance-2 code and 9.6 times worse than the other distance-4 one.

Nothing here says the claim is wrong. `d_circ = 4` is an upper bound on the circuit distance and is defended by a witness; a rate is a different quantity, and the point of the LER tier is that distance is a floor while the prefactor is free to be bad. So this is a schedule or decoder question, not a refutation.

Two candidate causes, both checkable on a workstation:

The schedule. `16-2-4` is a rotated surface code whose circuits came from `research/circuit_seed.py` and its standard geometric schedule. `32-8-4` is not surface-shaped, so its schedule was built differently, and #724 already showed for `72-12-6` that a sequential X-then-Z adaptation measurably loses a unit of circuit distance against an interleaved one. A schedule that preserves `d_circ` can still put far more weight-`d_circ` fault sets in the DEM than a good one, and the rate sees that where the distance does not.

The decoder. `32-8-4` has 6,952 DEM mechanisms against `16-2-4`'s 1,184, and BP degrades on graphs with many short cycles. If the gap is decoder rather than schedule, a higher OSD order or more BP iterations should close part of it, which is a cheap thing to test even though the tier pins the decoder for comparability.

[...]

Issue #702, opened 2026-08-24
Batch the measured-LER decode loop so the tier can admit d >= 7


vprusso:
Follow-up from the measured-LER review thread: at the pinned decoder's current per-shot cost the tier admits d = 5 memory circuits and nothing better. Measured on the merged seeds (arm64 numbers from the #680 review, reproduced within noise on x86): `bposd-cs-10` decodes the `25-1-5` DEM at ~1.1 ms/shot, `49-1-7` at ~16.8 ms/shot, and `81-1-9` at ~132.7 ms/shot. The failure floor needs about 100 failures per basis, so an honest claim costs minutes at d = 5, roughly an hour per basis at d = 7, and ~30 hours per basis at d = 9; the 120 s verification replica buys ~3 expected failures at d = 7 and ~0 at d = 9, so both fail closed as unverifiable within budget. The rule behaving honestly is by design, but the admissible set should grow.

The lever is the decode loop, which is embarrassingly batchable and currently is not: `measure_failures` decodes shot by shot through `BpOsdDecoder.decode`. Known routes, roughly in order of effort:

- batch the BP calls: `ldpc` exposes batched decoding paths, and message-passing over many syndromes vectorizes; a ~10x there re-admits d = 7 outright.
- reuse the decoder's internal state across shots instead of reconstructing per call (the constructor cost is already amortized, but per-call overhead in the Python loop is not).
- sample-side: `stim`'s sampler already batches; the loop structure just needs to consume it in blocks.

Acceptance: d = 7 honest claims verify within the existing 120 s/basis budget, demonstrated on the merged `49-1-7` seed, with the tier's statistical floors unchanged. d = 9 likely needs more than one 10x and can stay out of scope.

vprusso:
Checked the premise before starting on this, and the batching route is not available as stated.

`ldpc` 2.4.1's `BpOsdDecoder` exposes no batch entry point. Its public surface is `decode` (single syndrome), plus configuration and result attributes: `bp_method`, `osd_method`, `osd_order`, `max_iter`, `ms_scaling_factor`, `channel_probs`, `update_channel_probs`, `omp_thread_count`, and the various `*_decoding` outputs. So "batch the per-shot decode loop" cannot mean handing an array of syndromes to one call, because there is nothing to hand it to.

The two levers that do exist:

1. `omp_thread_count` on the decoder itself. Per-decode parallelism, likely poor for these problem sizes since each decode is small, but it costs one line to measure.
[...]

PR #1025, opened 2026-09-12
Measured LER: the first eight rates on the tier

vprusso:
Part of: #460

First measured rates on the logical-error-rate tier. No entry on the board carried one before this, although every layer was already in place: the schema block, `verify/ler_verify.py` with its replication gate, and the site render that prints `ler/round` on a detail page. The tier was complete and empty.

| entry | d | d_circ | rounds | X fail / shots | Z fail / shots | LER per round (X, Z) | detects |
|---|---|---|---|---|---|---|---|
| `12-4-2` | 2 | 2 | 2 | 493 / 10,000 | 501 / 10,000 | 2.53e-2, 2.57e-2 | 1.25x |
| `14-6-2` | 2 | 2 | 2 | 2,358 / 40,000 | 2,527 / 40,000 | 3.04e-2, 3.27e-2 | 1.18x |
| `16-2-4` | 4 | 4 | 4 | 413 / 60,000 | 408 / 60,000 | 1.73e-3, 1.71e-3 | 1.38x |
| `25-1-5` | 5 | 5 | 5 | 150 / 100,000 | 127 / 100,000 | 3.00e-4, 2.54e-4 | 1.46x |
| `32-8-4` | 4 | 4 | 4 | 632 / 10,000 | 590 / 10,000 | 1.66e-2, 1.55e-2 | 1.23x |
| `42-8-3` | 3 | 3 | 3 | 277 / 20,000 | 255 / 20,000 | 4.66e-3, 4.29e-3 | 1.40x |
| `45-5-4` | 4 | 4 | 4 | 262 / 40,000 | 268 / 40,000 | 1.65e-3, 1.69e-3 | 1.41x |
| `50-6-4` | 4 | 4 | 4 | 296 / 40,000 | 421 / 40,000 | 1.86e-3, 2.65e-3 | 1.38x |

Method: for each entry the DEM is re-derived from the committed `.stim` under the pinned stim version, sampled at the canonical reference rate `p = 0.001`, and decoded with the pinned `bposd-cs-10`. Seed 7 throughout. Shot counts differ only because they were raised until the 100-failure floor cleared. All eight return exit 0 from `verify/ler_verify.py` with the replication test included; the last column is the under-report factor the independent replica would have caught.

The tier immediately earns its keep. Four of these eight carry `d_circ = 4` at 4 rounds, and their measured per-round rates span an order of magnitude: `16-2-4` at 1.7e-3, `45-5-4` at 1.7e-3, `50-6-4` at 1.9e-3, and `32-8-4` at 1.7e-2. Identical distance, identical circuit distance, identical round count, ten times the failure rate. `32-8-4`'s per-shot rate is also higher than `14-6-2`'s, a distance-2 entry. That gap is the prefactor `d_circ` cannot see, and it is filed as #1024 with two candidate causes and a diagnostic.

A correction, twice over. The first commit said three entries were admissible inside the 120 second per basis replica budget; the second added two more; this one adds three more again, for eight of the seventeen circuit-carrying entries. Each time I had predicted admissibility from short pilots rather than measuring it, and each time the prediction was wrong in the direction that understates what the gate can do. The cause is that replica cost is 150 expected failures divided by the rate, so it depends on the rate as much as on throughput, and a pilot with single-digit failures cannot estimate a rate. `42-8-3` measured 1.4e-2 against a pilot's 8e-3, which was enough to move it across the line.

What does hold up is the throughput scaling, which has no sampling error in it and is posted on #702: decode cost tracks DEM mechanism count rather than distance, from 2,888 shots/s at 730 mechanisms down to 2 shots/s at 19,560. It just does not predict admissibility on its own. A campaign over the remaining nine entries is running and the measured boundary will go on #702.

`14-6-2` was first measured at 10,000 shots, which put the claim about two sigma below its own replica. Two sigma is inside tolerance, but low is the gaming direction, so I re-measured at 40,000 and the claim now sits inside noise of it.

Scope: no new codes, and no change to any distance, check matrix, layout or provenance. Each diff is the addition of one `circuit.ler` block. All eight entries are literature baselines with no `@handle` authors, so they take the baseline exemption in the authorship gate. Worth noting for #1012: there is no `ler` binding in `check_authorship.py`, so a non-author could not contribute a rate to a real submission the way this PR contributes to baselines.


vprusso:
Superseded by the rebuilt branch: this one was branched before the connectivity cleanup removed `45-5-4`, so it carried a modification to a deleted file and could not compute a merge commit, which is also why no CI run was ever dispatched for its last two commits.

Issue #908, opened 2026-09-06
codes/684-8-85.json claims d = 85 and admits a weight-84 logical


vprusso:
codes/684-8-85.json claims d = 85. A weight-84 Z-type logical exists, so the claim is over-stated.

The margin here is one, which is the thinnest of the refutations in this cluster, and that is worth stating first. It does not make the witness less checkable: a weight-84 logical either satisfies the three conditions below or it does not, and it does.

Witness, weight 84:

2 20 23 28 33 50 62 75 101 102 107 109 117 121 139 146 164 167 168 169 181 190 191 207 222 223 229
230 238 244 272 281 286 299 305 310 314 316 332 337 343 353 362 369 373 374 375 392 401 402 412 418
422 427 450 455 469 490 499 504 505 513 516 518 520 533 544 546 549 553 560 564 578 580 585 592 599
628 630 631 634 639 651 660

How it was found and checked:

- `uv run python verify/heuristic_distance.py codes/684-8-85.json --fast-trials 8000000 --seed 1` returns `"verdict": "refuted"` and exit code 2, after 15261 s.
- The witness was verified by rank against the code's own check matrices, independently of that tool: it lies in ker H_X, it raises the rank of H_Z from 338 to 339 so it is outside the stabiliser group, and it is nontrivial against the gf2 logical basis.
- The code itself checks out: n = 684 and k = 8 by rank, max row weight 8, and H_X H_Z^T = 0.

Denominator. Four entries in this n = 684 cluster were screened at 8,000,000 trials with codes/684-12-81.json as a positive control. The control refuted at exactly d = 66, reproducing issue #896 by a different route. Of the four, two were refuted (this one and codes/684-12-73.json) and two were not: codes/684-14-78.json returned found = 79 and codes/684-12-77.json returned found = 83, both `inconclusive`, meaning the budget never reached the claimed weight. Those nulls are not evidence that those two claims are correct.

MathysRennela:
@vprusso maybe it is preferable to open a PR for the refutation to avoid having an issue followed by a PR with the same content?

vprusso:
Good call, done. Converted both to PRs that make the correction directly: #911 for this one and #910 for codes/684-12-73.json. Each lowers the claimed distance, replaces the witness on the affected side, and records the tool invocation and the independent rank checks in provenance.notes. The issues are linked with closing keywords so they close on merge.

[...]

Issue #907, opened 2026-09-06
codes/684-12-73.json claims d = 73 and admits a weight-70 logical


vprusso:
codes/684-12-73.json claims d = 73. A weight-70 X-type logical exists, so the claim is over-stated.

Witness, weight 70:

11 16 33 42 50 55 56 73 92 93 101 125 144 145 166 171 195 201 206 211 219 225 232 235 241 251 281 282 289 294 298 311 319 324 329 332 369 394 400 410 413 427 455 463 483 498 510 514 515 524 529 539 540 546 551 560 568 604 605 606 609 610 611 642 643 656 662 669 675 679

How it was found and checked:

- `uv run python verify/heuristic_distance.py codes/684-12-73.json --fast-trials 8000000 --seed 1` returns `"verdict": "refuted"` and exit code 2, after 12277 s.
- Independently of that tool, the witness was verified by rank against the code's own check matrices: it lies in ker H_Z, it raises the rank of H_X from 336 to 337 so it is outside the stabiliser group, and it is nontrivial against the gf2 logical basis.
- The code itself checks out: n = 684 and k = 12 by rank, max row weight 8, and H_X H_Z^T = 0.

On the budget, because it matters for reproducing this. These n = 684 entries are low rate and the search needs trials in proportion. The already-known refutation of codes/684-12-81.json (issue #896) returns exactly its claimed 81 at the default 2,000,000 trials and only falls at 8,000,000. Anyone re-running this should use `--fast-trials 8000000` rather than the default.

Denominator, so the result is not read as broader than it is. Four entries in this n = 684 cluster were screened at 8,000,000 trials with codes/684-12-81.json as a positive control. The control refuted at exactly d = 66, reproducing #896 by a different route, so the screen demonstrably sees. Of the four, two were refuted (this one and codes/684-8-85.json) and two were not: codes/684-14-78.json returned found = 79 and codes/684-12-77.json returned found = 83, both `inconclusive`, meaning the budget never reached the claimed weight. Those two null results are not evidence that their claims are correct, only that 8,000,000 trials did not beat them.

vprusso:
Converted to a PR that makes the correction directly: #910. Same reasoning as in #908.

Issue #896, opened 2026-09-05
codes/684-12-81.json distance 81 is over-stated: a weight-66 logical exists


vprusso:
`codes/684-12-81.json` claims `distance.d = 81`. A nontrivial logical operator of
weight 66 exists, so the claim is over-stated by 15 and the true distance is at
most 66.

The witness is a Z-type logical, 66 qubits:

```
[7, 13, 37, 42, 52, 62, 78, 80, 86, 91, 99, 102, 115, 119, 125, 127, 130, 133,
 150, 151, 165, 182, 183, 212, 222, 255, 261, 264, 274, 287, 293, 310, 316, 330,
 333, 354, 362, 377, 393, 395, 426, 430, 433, 459, 470, 471, 488, 501, 503, 508,
 522, 538, 544, 550, 554, 557, 565, 572, 576, 594, 610, 623, 634, 639, 662, 666]
```

Checked against the entry's own parity checks, and the nontriviality confirmed a
second way by rank rather than only by the routine that produced it:

| check | result |
|---|---|
| weight | 66 |
| `H_X v = 0` | true, syndrome weight 0 |
| outside rowspace of `H_Z` | true |
| `rank(H_Z)` vs `rank(H_Z + v)` | 336 vs 337, so `v` is not in the rowspace |

### Reproduction

```
python verify/heuristic_distance.py codes/684-12-81.json --fast-trials 8000000 --seed 1
```

That exits 2 and reports `"verdict": "refuted"`, `"d_heuristic": 66`, with the
same Z-side witness listed above.

The default `--fast-trials` of 2000000 does not find it. The full ladder on this
code:

| trials | reading |
|---|---|
| 200000 | 91 |
| 2000000 | 81, exactly the claimed value |
| 8000000 | 66, refuting |

At the default budget the search returns precisely the claim and the entry
passes, which is presumably what happened at PR time.

### Why the per-PR gate missed it

Nothing here suggests a problem with the gate as designed. This is a k = 12 code,
and low-rate codes are much harder for a random information set search than
their length or check weight suggests: there are few nontrivial logical classes,
so a random information set rarely lands on a light representative. Measured on
the same machine and the same tool, a budget that reads a high-rate code exactly
under-searches a low-rate one badly.

That shows up in this repository's own CI timings too. On four recent
submissions the distance refutation gate took 8 minutes at k = 6, 41 minutes at
[...]

PR #1032, opened 2026-09-14
Admit n <= 1000 when check weight <= 8 and claimed d <= 40

### LINKED ISSUE #1016: Empirical study: can the 700-qubit cap be raised to 1000?
## Motivation

The current $n  \leq 700$ cap is explicitly a **verification-budget rule**: the CI gate runs under a fixed wall-clock budget, so the number of RIS trials per qubit decreases as `n` grows. The purpose of the cap is therefore not to claim that codes above 700 are intrinsically problematic, but to keep the distance refutation strong enough that the board does not overstate `d`.

The natural next step is to determine empirically whether $n \leq 1000$ is in fact beyond what the current verification machinery can reliably support. This issue proposes a benchmark study of the distance gate on codes spanning the current limit and the proposed new range.

## Proposed study

Construct a benchmark set of (fresh) representative codes in 4 categories:

A. `n <= 700`
B. `700 < n <= 800`
C. `800 < n <= 900`
D. `900 < n <= 1000`

and run the **same PR-time distance gate** repeatedly with independent random seeds.

For each code, compare the result of the normal gate against a much deeper RIS search (e.g. the 100k+ trial re-verification recommended in `CONTRIBUTING.md`), and record:

* the distance bound returned by the gate;
* the bound after deep re-verification;
* whether the result was subsequently refuted;
* number of RIS trials actually obtained;
* trials per qubit;
* runtime;
* variation across random seeds.

The key quantity is not simply runtime, but **how often the finite PR-time search lets an inflated distance upper bound survive**.

A useful summary would therefore be something like

$$
P(\text{PR gate accepts a bound later refuted by deep search})
$$

as a function of `n`.

## Benchmark codes

There are already useful candidates in the repository/literature around the proposed boundary. In particular, the reference bars include codes in the `775–975` range, so these provide natural stress tests for the verification pipeline.

It would also be useful to include several codes below 700, rather than only testing large codes, so that the study measures the change in behaviour relative to the current operating regime.

Ideally the benchmark set would include codes with:

* different `k/n` ratios;
* different claimed distances;
* different construction families;
* both codes whose distance is stable under deep search and codes known to exhibit distance inflation.

The latter is particularly important because the concern here is precisely the possibility of a shallow RIS search returning an optimistic bound.

## Possible outcome

If the empirical failure rate at `n <= 1000` is comparable to the current failure rate at `n <= 700`, that would provide evidence for raising the cap to 1000.

If reliability degrades substantially, the study would tell us where the degradation occurs and whether the issue is better addressed by:

* increasing the adaptive wall-clock budget;
* changing the trial allocation;
* improving the RIS implementation;
* making the cap dependent on $k/n$ rather than just $n$.
* or keeping the current cap.

This would also give us an evidence-based answer to the broader question of whether `n` itself is the right proxy for the verification budget.

FarLab:
## Summary

Two-tier blocklength cap, toward #1016:

- `n <= 700`: admitted as before, no change.
- `700 < n <= 1000`: admitted only when max check weight is `<= 8` **and** claimed `d <= 40`. Anything else is rejected at the resource layer, before dense matrices or the search budget are spent.

## Why these two bounds

**Weight `<= 8`.** Per-trial cost of the accelerated RIS search grows as roughly `n^2` and never depends on check weight, so the weight bound does not make the search cheaper per se; it keeps the extended tier at the sparse end where the BP+OSD cross-check is meaningful and the CI bill is predictable. Measured 8M-trial fast pass (laptop ms/trial, extrapolated with the 1.9x CI factor):

| code | ms/trial | 8M trials on CI |
|---|---|---|
| board `[[684,20,72]]`, w=8 | 0.152 | ~38 min |
| synthetic GB n=700, w=8 | 0.151 | ~38 min |
| synthetic GB n=1000, w=8 | 0.344 | ~87 min |

**`d <= 40`.** The verifier's search already distinguishes *refuted*, *corroborated* (it reached the claimed weight) and *inconclusive* (it never got down to the claim); the gate only fails on *refuted*, so inconclusive claims pass. Information-set decoding hits a weight-`d` logical with probability about `c^d`, so reach is logarithmic in the budget. With 50k accelerated trials on the board's `w <= 8` entries at `n >= 300`:

| claimed d | codes | reached the claim | gap (found − claimed) |
|---|---|---|---|
| 3–16 | 10 | 10 of 10 | 0 |
| 21–28 | 4 | 3 of 4 | 0 to +3 |
| 30–36 | 6 | 2 of 6 | +1 to +10 |
| 70–81 (all n=684) | 6 | 0 of 6 | +12 to +23 |

Reach improves by roughly 4–6 weight units per 4x trials, which puts the gate's 160x deeper pass near `d ≈ 40`. Every over-stated `n ≈ 684` entry (#896, #907, #908, #942) lived above that line, so above 700 the verifier refuses claims the search could not have tested. No published `w <= 8` code at `n <= 1000` claims `d` above the mid-30s, so the bound excludes nothing real.

## Changes

- `verify/qldpc_verify.py`: `BASE_MAX_N = 700`, `MAX_N = 1000`, `EXT_MAX_CHECK_WEIGHT = 8`, `EXT_MAX_D = 40`; the tier check lives in `resource_errors` next to the existing cap. `MAX_COORDINATES` follows `MAX_N`.
- `schema/code.schema.json`: `n` maximum 1000 (the conditional part is enforced by the verifier, which the description says).
- `CONTRIBUTING.md`, `TRACKS.md`, `schema/SCHEMA.md`: describe the tiers; the weight-9 over-cap literature bars (mitten, ZSZ-LP) stay bars.
- `verify/test_verifier.py`: synthetic resource-layer fixtures for both tiers, each violation alone, both together, and above `MAX_N`.
- `verify/validator_manifest.json`: re-pinned for the two changed trusted files.

## Test plan

- `uv run --frozen python run_tests.py --skip-slow`: 143 passed, 10 skipped after the manifest re-pin.
- No existing board code is affected (all are `n <= 700`).

## Not in this PR

- The gate budget in `gate_changed._budget` still scales with `n` only; indexing it on claimed `d` (#896) would let the extended tier's bound rise with the budget.
- Site copy that quotes the 700 cap in generated notes is left as is.


FarLab:
something to consider: cap the distance for all code submissions? say d<=50? very high distance codes with 8m RIS trials are weakly supported claims (as we have seen many times).

MathysRennela:
> something to consider: cap the distance for all code submissions? say d<=50? very high distance codes with 8m RIS trials are weakly supported claims (as we have seen many times).

I'd push back against that: we are getting less and less submissions with overstated and high distance upper bound. This is because it is increasingly difficult to find codes with very high distance that are board advancing. I'd say that this area has been more or less exhausted so we do not need to worry about it.

On top of that, distance upper bound estimation algorithms are bound to improve, especially with GPU access. So I would defer the decision for now.

schema/SCHEMA.md
```diff
@@ -203,11 +203,11 @@ JSON cannot force unbounded dense-matrix allocation in CI. Current automatic
 limits are:
 
 - JSON file size: 5 MB.
-- `n <= 700` (the verification-budget cap, issue #249; raise-only).
+- `n <= 700`, or `n <= 1000` when max check weight is at most 8 and the claimed `d` at most 40 (the verification-budget cap, issues #249 and #1016; raise-only).
```
MathysRennela: mentioning issues in instruction files tends to confuse agents

verify/qldpc_verify.py
```diff
@@ -62,12 +62,26 @@
...
+# check weight is at most EXT_MAX_CHECK_WEIGHT and its claimed distance at most
+# EXT_MAX_D. The weight bound keeps the per-trial search cost (which grows as
+# roughly n^2 and never depends on weight) at the sparse end and keeps the
+# BP+OSD cross-check meaningful; the distance bound is the depth the gate's
+# 8M-trial RIS pass actually reaches (information-set decoding hits a weight-d
+# logical with probability ~c^d, so reach is logarithmic in budget; measured
+# 2026-09-14: 50k trials reach every claim at d <= 28 and none at d >= 70, and
+# the gate's 160x deeper pass lands at d ~ 40). A claim above that would pass as
+# "inconclusive" rather than corroborated, which is the gap every over-stated
+# n ~ 684 entry lived in (issues #896, #907, #908, #942).
```
MathysRennela: maybe a shorter comment is more suitable here

verify/qldpc_verify.py
```diff
@@ -93,10 +107,27 @@ def resource_errors(doc):
...
     errs = []
+    max_weight = max((len(s) for s in supports), default=0)
     if n > MAX_N:
         errs.append(f"n={n} exceeds the blocklength cap {MAX_N} "
                     f"(verification-budget rule, issue #249; the pipeline "
                     f"cannot stand behind a distance claim above it)")
+    elif n > BASE_MAX_N:
+        # Extended tier (issue #1016): sparse checks and a reachable distance
+        # claim only. The claimed d is read here, before the dense pass, so an
+        # over-cap claim is rejected without spending the search budget.
```
MathysRennela: I'd suggest removing this comment

verify/qldpc_verify.py
```diff
@@ -93,10 +107,27 @@ def resource_errors(doc):
...
+            errs.append(f"n={n} is above {BASE_MAX_N}, where the cap admits only "
+                        f"check weight <= {EXT_MAX_CHECK_WEIGHT} (found {max_weight}); "
+                        f"the verification budget cannot stand behind heavier "
+                        f"codes at this size (issue #1016)")
+        if isinstance(claimed, int) and claimed > EXT_MAX_D:
+            errs.append(f"n={n} is above {BASE_MAX_N}, where the cap admits only "
+                        f"claimed distance <= {EXT_MAX_D} (claimed {claimed}); "
+                        f"the gate's search does not reach deeper claims at this "
+                        f"size, so they could pass unrefuted but uncorroborated "
+                        f"(issue #1016)")
```
MathysRennela: issue mentioned in the comment might confuse agents

verify/test_verifier.py
```diff
@@ -164,6 +164,40 @@ def main():
     r = rep(d)
     check("oversized n rejected", not r["ok"])
 
+    # 6c. two-tier blocklength cap (issue #1016): above BASE_MAX_N only sparse
+    # checks with a reachable distance claim are admitted. Synthetic shapes,
+    # checked at the resource layer (before dense matrices), so no board code
+    # is pinned and the rule is tested independently of any real submission.
```
MathysRennela: issue comment might confuse agents

CONTRIBUTING.md
```diff
@@ -110,17 +110,31 @@ by the weekly board sweep, in which case the code is removed. The seed is printe
 so any failure reproduces.
 
 Public CI also enforces resource limits before dense verifier matrices are
-allocated: 5 MB JSON files, `n <= 700`, at most 10000 checks per side, max check
-weight 32, at most 200000 total support entries, and at most 700 locality
+allocated: 5 MB JSON files, `n <= 700` (or `n <= 1000` when max check weight is
+at most 8 and the claimed `d` at most 40), at most 10000 checks per side, max
+check weight 32, at most 200000 total support entries, and at most 1000 locality
```
MathysRennela: This is a part where agents tend to get confused, so maybe it is a good time to turn this part into a logical contract:
- n <= 700 or (n <= 1000 and w <= 8)
- d <= 40
- checks-per-side <= 10000
- w <= 32
- ...

Actually, this is probably worth implementing as a boolean function that the agent should call to verify whether the code is suitable, rather than leaving this paragraph up to interpretation. If any of those requirements fall out of the agent's context window (which happens often with non-frontier models), it might mislead the agents into investigating codes that are not suitable.

CONTRIBUTING.md
```diff
@@ -110,17 +110,31 @@ by the weekly board sweep, in which case the code is removed. The seed is printe
...
+which grows as roughly `n^2` and never depends on weight, at the sparse end
+(measured: the gate's 8M-trial pass costs about 87 CI minutes for a weight-8
+code at `n = 1000` against 38 at `n = 700`) and keeps the BP+OSD cross-check
+meaningful. The distance bound is the depth that pass actually reaches:
+information-set decoding hits a weight-`d` logical with probability roughly
+`c^d`, so reach is logarithmic in the trial budget, and a claim beyond it passes
+the gate as "inconclusive" (never reached) rather than corroborated. That gap is
+where every over-stated `n ~ 684` entry lived (issues #896, #907, #908, #942),
+so above 700 the verifier rejects claims the search could not have tested.
+
```
MathysRennela, line 141: imo this should be self-contained without mentioning issues that will bloat the context or consume tokens.

MathysRennela (changes requested):
It's good overall but in practice I think that some of the changes made will confuse agents a bit. Let me know what you think of the comments.

FarLab:
makes sense, but for the hackathon we should maybe only accept n<=1000, w<=8 and d<=40. otherwise it will be not uniform.

MathysRennela:
> makes sense, but for the hackathon we should maybe only accept n<=1000, w<=8 and d<=40. otherwise it will be not uniform.

I think it's ok to make the changes to the CI gate if it becomes a boolean function that determines acceptance.

schema/SCHEMA.md
FarLab: Dropped the issue numbers here; the line now states the rule and names the predicate (`qldpc_verify.admissible(n, w, d)`).

verify/qldpc_verify.py
FarLab: Cut to six lines: what the two tiers are and that `admissible()` is the contract. The measurements live in the PR description instead.

verify/qldpc_verify.py
FarLab: Removed. The tier check is now a call to `admissible()` with one error message.

verify/qldpc_verify.py
FarLab: Issue references removed from the messages (the pre-existing `#249` mention in the over-MAX_N message too, for consistency).

verify/test_verifier.py
FarLab: Removed; the comment now just says what the block tests.

CONTRIBUTING.md
FarLab: Done both ways. CONTRIBUTING now lists the limits as a checklist (`n <= 700` or `n <= 1000` with `w <= 8` and `d <= 40`; checks per side; `w <= 32`; support entries; coordinates; file size), and the blocklength line is one boolean, `qldpc_verify.admissible(n, w, d)`, so a search can test a candidate before spending anything on it. `structure_errors(doc)` returning `[]` covers the whole list for a full document. Per the thread, `d <= 40` applies in the extended tier only; `n <= 700` keeps accepting any `d`.

CONTRIBUTING.md
FarLab: Rewritten as one short self-contained paragraph, no issue numbers.

FarLab:
Addressed the review in faf1985e:

- The cap is one predicate, `qldpc_verify.admissible(n, w, d)`: `n <= 700`, or `n <= 1000` with `w <= 8` and `d <= 40`. `resource_errors` calls it and emits a single message when it fails.
- CONTRIBUTING lists the resource limits as a checklist with the predicate next to it, plus a short rationale; SCHEMA.md and the schema description match.
- Issue numbers are gone from every agent-facing surface touched here (schema, SCHEMA.md, verifier comments and messages, test comments).
- Tests cover the predicate at each bound and the resource layer's use of it; validator manifest re-pinned. `run_tests.py --skip-slow`: 143 passed, 10 skipped.

The `d <= 40` bound stays scoped to the extended tier per the thread above; below 700 nothing changes.

MathysRennela (approved):
LGTM!

vprusso (approved):

PR #1043, opened 2026-09-15
[ADVERSARIAL TEST — DO NOT MERGE] Direct sum [[16,6,4]]⊕[[16,6,4]] submitted as [[32,12,4]]

FarLab:
## Code submission (adversarial CI probe — do not merge)

This PR deliberately submits a **direct sum of two existing board codes** to confirm that the CI connectivity gate (`tanner_connected`, issue #748) rejects it. It is not a discovery and must not be merged.

- Parameters: [[n, k, d]] = [[32, 12, 4]] — block-diagonal `H_X = diag(H_X', H_X')`, `H_Z = diag(H_Z', H_Z')` from two copies of `codes/16-6-4.json`
- Tracks: weight-8 / unrestricted locality
- Distance confidence: upper_bound (inherited witnesses)

### Outcome
**Caught.** Run 34946503986, step "verify all submissions (trusted verifier for code changes)":

```
FAIL  codes/32-12-4.json  -> tanner_connected
```

The refutation gate never started, so the probe cost one verify-all pass (~12 min) and no RIS budget.

### Known limitation (tested locally, not pushed)
The gate checks connectivity of the *submitted* check rows, not of the stabilizer group. Appending one linearly dependent row per side (`row0_A XOR row0_B`) joins the Tanner graph without changing the code; that variant passes the base-branch verifier with no failed checks. Running the same component count on the reduced row echelon form of each side catches it, because the RREF of a direct-sum row space is block-diagonal.

### Checklist
- [x] `python verify/qldpc_verify.py codes/<file>.json` run locally: fails on `tanner_connected` by design
- [x] Construction and references filled in under `provenance` (labelled as an adversarial test)
- [x] Equivalence to existing entries noted in `provenance.notes`

None. kd²/n is unchanged by direct sum (6·16/16 = 12·16/32 = 6); the entry only exists to exercise the gate.


FarLab:
Follow-up from the probe: the gate is evadable, and a canonical variant of it is not.

**Evasion (tested locally, not pushed).** `tanner_connected` counts components of the *submitted* rows. Appending one linearly dependent check per side (`row0_A XOR row0_B`) joins the Tanner graph without changing the stabilizer group; that variant passes the base-branch verifier with no failed checks.

**Fix.** Run the same component count on the rows of `gf2.rref(H_X)` ∪ `gf2.rref(H_Z)`. The RREF of a direct-sum row space is block-diagonal, so the check is exact for stabilizer-group decomposability, and a full board scan takes ~3 s.

| input | raw Tanner components | RREF components |
|---|---|---|
| this PR (plain direct sum) | 2 | 2 |
| disguised variant (+1 redundant row/side) | 1 | **2** |
| `codes/181-12-10.json` | 1 | **3** |
| `codes/240-12-12-graft.json` | 1 | **2** |

The two live entries are both r=1 lattice grafts. Their extra components are single qubits (50 and 145; 38) that touch no X check and have a weight-1 Z stabilizer in the Z row space, i.e. frozen, unused qubits; true block lengths are 179 and 239. Any PR adding the RREF check must handle those two entries in the same change.

FarLab:
@MathysRennela we might have to update the direct sum gate to use the RREF approach to catch these disguised direct sums.

MathysRennela:
> @MathysRennela we might have to update the direct sum gate to use the RREF approach to catch these disguised direct sums.

Agreed, we definitely should.

PR #1060, opened 2026-09-15
Check stabilizer-group connectivity, not just the submitted Tanner graph

FarLab:
## What

Adds `stabilizer_group_connected` to `verify/qldpc_verify.py`, next to `tanner_connected`. It runs the same connected-component count, but on the rows of `gf2.rref(H_X)` and `gf2.rref(H_Z)` instead of the submitted rows, with isolated qubits counted.

## Why

The adversarial probe in #1043 was caught by `tanner_connected`, but a one-line variant is not: append one linearly dependent check per side (`row0_A XOR row0_B`) to a block-diagonal direct sum. The Tanner graph is now connected, the stabilizer group is unchanged, and the base-branch verifier passes it with zero failed checks.

The RREF is canonical for the row space. If the row space is a direct sum over disjoint qubit sets, every RREF row lies in exactly one block (a row with a foot in two blocks would have a non-pivot leading 1 in the second block, contradicting reduced form). So the component count on RREF rows equals the number of independent blocks of the stabilizer group, whatever redundant rows the submission carries. A qubit frozen by a weight-1 stabilizer hiding in the row space appears as a size-1 block, which is the "unused qubit" case the rule already intends to reject. A whole-board scan takes about 3 s.

| input | `tanner_connected` | `stabilizer_group_connected` |
|---|---|---|
| plain direct sum (#1043) | fail (2) | fail (sizes 16, 16) |
| bridged direct sum, +1 redundant row/side | pass | **fail** (sizes 16, 16) |
| `codes/181-12-10.json` | pass | **fail** (sizes 179, 1, 1) |
| `codes/240-12-12-graft.json` | pass | **fail** (sizes 239, 1) |
| the other 568 board entries | pass | pass |

## Changes

- `verify/qldpc_verify.py`: `_stabilizer_block_count` (union-find over RREF rows) and the new `record`. Detail string reports block count and the largest block sizes.
- `verify/test_verifier.py`: synthetic fixtures only, per the repo's rule against pinning live board data: a bridged direct sum of two [[4,2,2]] codes, and a [[4,2,2]] plus one frozen qubit. Both have connected Tanner graphs; both are rejected on the new check; the good fixture still passes.
- `verify/validator_manifest.json`: re-pinned with `--update`.
- `CONTRIBUTING.md`: the connectivity bullet now states the stabilizer-group form of the rule.

## Blocked on two board entries (CI will be red until they are re-filed)

Both are r=1 lattice grafts by @MathysRennela. In each, the flagged qubit touches no X check and has a weight-1 Z stabilizer in the Z row space, so it is frozen. Dropping the frozen qubits leaves k unchanged, keeps every existing witness intact (no witness touches a dropped qubit), and keeps max weight 8:

| entry | frozen qubits | re-file as |
|---|---|---|
| `codes/181-12-10.json` | 50, 145 | [[179, 12, 10]] |
| `codes/240-12-12-graft.json` | 38 | [[239, 12, 12]] |

This PR cannot carry those edits: the submission-scope check forbids code data alongside `verify/` changes, and the authorship check binds the entries to their author. Suggested order: the author re-files both in codes-only PRs, then this merges. Self-tests here fail on exactly those two entries and nothing else.

Refs #748, #1043.


FarLab:
@MathysRennela the CI will be red on this PR due to the two codes that are apparently disguised direct sums. Can you edit those?

MathysRennela:
Follow-up re-file PRs are up, per the order suggested here:

- #1065 — [[181,12,10]] → [[179,12,10]] (drops frozen qubits 50, 145)
- #1066 — [[240,12,12]] → [[239,12,12]] (drops frozen qubit 38)

Both passed `verify/validate_candidate.py` with `passed: true`, and against this branch's verifier both report a single `stabilizer_group_connected` block (size 179 / 239). Witnesses are preserved under reindexing (weight 10/10 and 12/12; the kit's independent search corroborated both). Once both merge, this PR should rebase to a green `verify` run.

PR #1096, opened 2026-09-16
fix: reject orphaned circuit artifacts

### LINKED ISSUE #1028: Connectivity cleanup left circuits/45-5-4 and circuits/66-5-5 orphaned, and outside the authorship gate
The connectivity cleanup removed `codes/45-5-4.json` and `codes/66-5-5.json`, and both carried circuit-tier artifacts that are still committed:

```
circuits/45-5-4/memory_x.stim   circuits/45-5-4/memory_x.dem
circuits/45-5-4/memory_z.stim   circuits/45-5-4/memory_z.dem
circuits/66-5-5/memory_x.stim   circuits/66-5-5/memory_x.dem
circuits/66-5-5/memory_z.stim   circuits/66-5-5/memory_z.dem
```

Eight files for two entries that no longer exist. These were two of the seventeen entries on the circuit tier, so the cleanup took about twelve percent of it with no note that it had.

Two consequences, one cosmetic and one not.

The site prunes orphan detail pages by keying on the surviving slugs, so the board self-heals on the render side. Nothing prunes `circuits/`, so the artifacts sit there indefinitely with no entry pointing at them.

The one that matters: an orphaned `circuits/<slug>/` directory is outside the authorship gate. `check_authorship.py` maps a diff under `circuits/<slug>/` back to `codes/<slug>.json` and then does `if not os.path.exists(p): continue`, so with the JSON gone the mapping resolves to a missing file and the check silently skips it. `gate_changed.py` uses the same mapping in `map_changed`. So right now anyone can modify, swap or delete the committed `.stim` and `.dem` files for `45-5-4` or `66-5-5` and no authorship binding applies. It is a narrow hole, since the artifacts no longer back a board claim, but it is the exact case the `circuits/` mapping was added in #1012 to close, and it is open on main today.

Worth deciding rather than leaving: either remove the orphaned directories along with the entries, or keep them deliberately as reference artifacts and have `check_authorship` treat a `circuits/` diff with no corresponding JSON as a violation instead of a skip. The second is a two-line change and closes the hole for any future removal as well.

Noticed while rebasing #1027, which had added a measured rate to `45-5-4` before the entry was removed. That PR now carries seven entries instead of eight.

Related: #1004 and #1005 for the cleanup, #1012 for the `circuits/` claim-surface mapping, #505 for the circuit tier.

davidplu:
Closes #1028.

Both validation gates now reject changed entries whose code JSON is missing while circuit artifacts remain. Complete cleanup and coherent code/circuit moves are allowed; deleting only a JSON cannot leave orphaned circuits behind.

The authorship gate reuses its existing rename-split diff instead of running a second Git diff, and emits a dedicated orphan diagnostic. The refutation gate also disables rename detection when collecting changed paths. This PR removes the eight existing orphaned artifacts identified in the issue and repins the two changed validators.

Validation: 10 focused tests passed, including seven real-Git scenarios covering orphan edits/additions, full and partial cleanup, JSON-only deletion, and code/circuit moves. The 31 validator-integrity tests passed. Compilation, committed manifest verification, and the committed-tree prose check passed. The existing authorship script has Windows path-separator failures; the new real-Git regression suite passes on Windows.

AI disclosure: Implemented and tested with an AI coding assistant.


FarLab (changes requested):
Thanks for picking this up. The mechanics are sound and the tests and manifest repin check out, but I think the fail-closed rule is drawn one step too wide.

**The cleanup path is now blocked.** The rule flags any `circuits/<slug>/` diff whose `codes/<slug>.json` is missing, including when the JSON was deleted in the same PR. I tried the obvious follow-up on top of this branch, `git rm -r circuits/45-5-4 circuits/66-5-5`, and both gates fail:

```
codes/45-5-4.json: authors [] do not include @vprusso; no corresponding code entry exists for the changed circuits/ artifacts
FAIL     codes/45-5-4.json: changed circuits/ artifacts have no corresponding code entry
```

So once this merges, the eight orphaned files in #1028 can't be removed through CI, and this PR doesn't remove them either. The same hits any future cleanup that removes a code and its circuits together, and a circuit-tier refutation that renames the slug and moves the directory. Deleting only the JSON and leaving the circuits behind still passes, which is the pattern that created the issue in the first place.

**Suggestion:** flag a missing JSON only when it was *not* deleted in the same diff (`changes.get(f) != "D"`), so a true orphan edit fails but a coherent code+circuits removal passes. That also lets `changed_circuit_codes` go away: `changed_codes` already sees the circuits diff, so it can return the mapped slugs alongside the statuses instead of running a second `git diff`. Note the second diff lacks `--no-renames`, so whether a same-PR rename gets flagged currently depends on git's similarity heuristic. Then flip the "deleting a code and its circuits together fails closed" test to assert it passes, and either drop `circuits/45-5-4` and `circuits/66-5-5` here or in a follow-up.

Minor: the "authors [] do not include @x" prefix on the orphan message is a bit confusing; a dedicated line for this case would read better.

davidplu:
Thanks for the review. The updated commit allows complete orphan cleanup and coherent code/circuit removal or moves, while rejecting any remaining circuits without a code entry. I used the remaining artifacts in the submitted tree as the distinction, rather than exempting every deleted JSON: that also catches the JSON-only deletion that originally created the orphans.

The redundant authorship diff is removed, rename detection is disabled for the refutation gate's path collection, and orphan errors now have a dedicated diagnostic. The eight existing orphaned artifacts are removed in this PR.

Validation: 10 focused tests passed (including seven real-Git scenarios), 31 integrity tests passed, and both gates accept this PR's actual cleanup. Committed manifest verification and the clean-worktree prose gate also passed. The existing broader authorship script still has seven Windows path-separator failures; the new regression suite passes on Windows.

Ready for another review. AI assistance was used for this revision as disclosed in the PR body.

FarLab (approved):
This is a better rule than the one I suggested: asking "are there still files under `circuits/<slug>/`?" for any missing code path catches the JSON-only deletion that created the orphans in the first place, while letting a coherent code+circuits removal or a refutation rename through.

Checked the head against both gates: this PR's own cleanup passes; deleting a live code's JSON while leaving its circuits fails both; removing code and circuits together passes; a refutation rename with the circuits directory moved passes and the new path is gated normally; editing an orphan's circuit or adding circuits for a slug with no JSON fails both; a non-author editing a live handle-authored code's circuit is still a violation. New tests, the circuit-gate and authorship suites, and the integrity check all pass at head, and the circuits diff is exactly the eight orphan files.

Two things before merge, neither about the logic:

- The branch conflicts with main on `validator_manifest.json` (#1098, #1099 and #1100 landed since). Needs a rebase and `check_validator_integrity.py --update`.
- `test_deleted_code_without_circuit_diff_is_skipped` in `test_circuit_gate.py` still monkeypatches `_circuits_diffed`, which the new missing-file branch no longer consults, so it passes vacuously. Either drop it or have it assert the empty-directory case directly.

Approving; happy to merge once it is rebased.

vprusso:
Agree with @FarLab . Approved modulo a merge!

vprusso (approved):

davidplu:
Rebased onto current `main` (`f7d19efd`), regenerated `validator_manifest.json` with `check_validator_integrity.py --update`, and replaced the vacuous `_circuits_diffed` monkeypatch test with a direct empty-directory assertion.

Local validation after the rebase: 10 focused behavior tests passed, all 31 validator-integrity tests passed, both contribution gates passed against `origin/main`, and the required pre-push prose check passed. The PR is now conflict-free and mergeable. The new GitHub Actions runs are awaiting maintainer approval. thanks again.

MathysRennela (approved):
LGTM!

PR #1093, opened 2026-09-16
Add [[924,18,31]], a weight-8 open-boundary planar tile code on a 21x…

e-eight:
## Code submission

- Parameters: [[n, k, d]] = [[924,18,31]]
- Tracks: 2D-local bilayer / weight ≤ 8 (computed by the verifier from `H` and the layout)
- Distance confidence: X: upper_bound, Z: upper_bound

Family tag: tile (a self-declared filter, never used for ranking).

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/924-18-31.json` passes locally
- [x] Construction and references filled in under `provenance`
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`

### What frontier does this advance?

- **2D-local bilayer / weight ≤ 8**: on the Pareto frontier. It is a new
  `(n, k, d) = (924, 18, 31)` point in this cell, and nothing in the cell beats
  it on all of `n`, `k`, `d` and check weight at once (`verify/validate_candidate.py`
  reports `board_advancing: true`, `dominated_by: []`).
- It raises the **weight-8** sub-frontier of the cell from `kd²/n = 12.79`
  (`codes/563-18-20.json`, [[563,18,20]]) to `18.721`, and passes the live
  reference bar in `TRACKS.md` — the ILP-exact published weight-8 tile code
  [[512,18,19]] (arXiv:2504.09171) at `kd²/n ≈ 12.7`.
- It is the same open-boundary tile family as `codes/578-18-20.json` and its
  qubit-reduced descendants `codes/566-18-20.json`, `codes/572-18-20.json` and
  `codes/563-18-20.json`, with the same `k = 18` and check weight, but built on
  a `21×22` lattice instead of a `17×17` one. It therefore **extends rather than
  dominates** those entries: they are smaller in `n`, this is one to eleven
  distance units larger in `d`.
- The cell's overall best `kd²/n` remains `19.20` (`codes/360-12-24.json`,
  weight-6); this code is the best **weight-8** point in the cell.
- Distance is a witness-backed **upper bound** (`d_X ≤ 31`, `d_Z ≤ 45`) — not
  certified exact. The deepest null result is 48 fresh-seed RIS runs at 2M
  trials each on this lattice, plus 32M-trial runs on the same supports.

Score kd^2/n = 18.721, max check weight 8, locality class local-2d-bilayer.

[...]

PR #1029, opened 2026-09-12
Add [[682,142,87]] generalized-bicycle code

DennisWayo:
Replaces the working `[[682,172,d<=78]]` attempt with a witness-backed cyclic-ideal generalized-bicycle candidate.

Submission:

- Adds `codes/682-142-87.json`
- Parameters: `[[682,142,d<=87]]`
- Score: `k d^2 / n = 1575.950146627566`
- Family: `generalized-bicycle`
- Fingerprint: `dd8c8ca68949826f`
- Signature: `ede7e6eeee357561`

Local pre-submission checks:

- `uv run python verify/qldpc_verify.py codes/682-142-87.json` passes.
- A local CI-style fire test at the submitted `d<=87` claim ran the trusted circulant-GB structural pass, Python RIS, and RIS-fast.
- The final fire test found no logical below 87: structural best was equal-weight 88, Python RIS best was 92, and RIS-fast returned not refuted with best equal-weight 87 over 8,000,000 trials.

This is an honest witness-backed upper-bound claim, not an exact-distance certificate.

Supported by OpOpenAI GPT-5 (Codex).

PR #984, opened 2026-09-10
Add [[630,14,30]] weight-6 generalized bicycle code on Z_315

dorakingx:
## Code submission

- Parameters: [[n, k, d]] = [[630, 14, d ≤ 30]]
- Track: unrestricted / weight-6 (no 2D layout; see the note for why one is impossible)
- Distance confidence: upper_bound

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/630-14-30.json` passes locally
- [x] Construction and references filled in under `provenance`
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes`: checked, not equivalent.
      The gate reports no exact duplicate and no WL-equivalent entry, and because WL cannot separate
      circulant codes I compared closed-walk counts of the Tanner graph as well. The board entries at
      n = 630 are [[630,12,34]] w=6, [[630,126,20]] w=9, [[630,6,36]] w=6 and [[630,86,3]] w=4; none
      shares this k and d.

### What frontier does this advance?
unrestricted / weight-6: it joins the Pareto frontier and **dominates [[660,8,30]]** on all four axes
(630 < 660 qubits, k 14 > 8, same d and same check weight), which therefore leaves the frontier of
every cell the two share. Nothing on the board dominates it. It is **not** the cell leader by kd^2/n —
at 20.000 it is third, behind @vprusso's [[672,20,32]] (30.476) and my own [[630,12,34]] (22.019).

### How it was found
Cyclic generalized-bicycle code on Z_315: H_X = [A|B], H_Z = [B^T|A^T] with A, B the circulants of
a(x) = 1 + x + x^94 and b(x) = 1 + x^10 + x^50, so every check has weight 6 and
k = 2*deg gcd(a, b, x^315 - 1) = 14 exactly. From the same exhaustive designed-divisor enumeration that
produced my [[630,12,34]]: all weight-6 trinomial pairs on Z_N for N = 181..350 at k >= 8, 109,888
listed pairs and 102,092 classes under monomial factors, units and the two-block swap.

Distance: a fresh-seed fast-RIS ladder, a new seed at every rung and a deeper pair depth as it goes,
46 @120 -> 38 @2k -> 34 @20k -> 32 @200k -> 30 @1M -> 30 @2M -> 30 @5M -> **30 @20M** (last rung seed
[...]

PR #1098, opened 2026-09-16
Cap the gate's fast RIS pass at 90 min wall-clock instead of letting the job die

FarLab:
The gate's accelerated RIS pass (`_fast_refute`, 8M trials for anything above n≈300) had no wall-clock bound of its own, while every other search in the gate does (python RIS seeds 240 s, decoder 10 s, circuit DEM search 120 s/basis). Per-trial cost grows with n, so at n≈900+ the pass alone outgrew the 90-minute job. [[924,18,31]] (#1093) was killed 69 minutes into the gate with nothing printed and no receipt; #1029 and #1033 hit the same kill earlier and were merged by hand, so there is no receipt for either.

| code | fast-pass ms/trial (laptop, 4 thr) | 8M trials, laptop | gate step in CI |
|---|---|---|---|
| 630-14-30 (#984) | 0.17 | 23 min | 45 min, passed |
| 924-18-31 (#1093) | 0.34 | 46 min | killed at 69 min |

This makes the gate own its time-box, so "no refutation within budget" is an explicit, receipted verdict and the job-level timeout goes back to meaning a hang or a bug:

- `_fast_refute` drives the accelerator in slices with a deadline: it stops at the trial target **or `FAST_SECONDS` = 90 min, whichever binds first**, and returns the trials actually completed. Slices are re-sized to ~60 s from the measured rate (bounded overshoot at any n), each with its own derived seed, so a time-trimmed run is a prefix of the full one under the same seed. The 90 minutes are for the fast pass alone; self-tests and the other steps are outside it.
- The receipt's `distance_gate` block now carries `fast_trials` (completed), `fast_target` and `fast_seconds`; the PASS tag says `fast x N of M (wall-clock cap 90 min reached)` when the cap ended the search. Progress lines every 5 minutes.
- `verify.yml`: `timeout-minutes` 90 → 130, so the fixed ~30 min of self-tests (~8), verify_all (~13), python passes (~5) and checkouts fit alongside the cap. `PYTHONUNBUFFERED=1` so a killed step still leaves its seed and partial results in the log.
- Regression test with a fake accelerator (no gf2_fast needed): slices sum exactly to the reported count, a tight cap trims the search, a hit surfaced in a later slice is validated by the python stack, a bogus support is rejected.
- CONTRIBUTING.md notes the cap next to the blocklength-cap rationale; manifest re-pinned.

The 90 minutes are calibrated to the blocklength cap: CONTRIBUTING already estimates ~87 CI minutes for 8M trials at n=1000, so codes under the cap still get essentially the full target and only the very largest are trimmed. At n=924 the runner will finish roughly 8M trials in the budget.

Validation: `test_refute_gate.py` (5 passed with the accelerator built), `test_gate_diff_classes.py`, `test_circuit_gate.py`, `test_validator_integrity.py` all green; end-to-end smoke of the deep path with the cap forced to 3 s produced `fast_trials 1,411,578 of 5,046,000` in the receipt and the capped tag; ruff clean on the added lines; `git diff --check` clean.


verify/gate_changed.py
```diff
@@ -82,6 +85,26 @@
...
+# the very largest are trimmed. The verify job's timeout-minutes must leave this
+# plus the fixed ~30 min of self-tests, verify_all and python passes.
+FAST_SECONDS = 90 * 60.0
+# The accelerator is called in slices so the deadline can be checked between
+# them; the slice is re-sized to ~FAST_SLICE_SECONDS from the measured rate so
+# the overshoot past the deadline is bounded regardless of n.
+FAST_SLICE_SECONDS = 60.0
+FAST_SLICE_MIN, FAST_SLICE_MAX = 10_000, 1_000_000
+FAST_PROGRESS_SECONDS = 300.0
+
```
MathysRennela, line 100: The commented lines do not really add relevant information beyond this PR.
imo it would be better to delete them

vprusso (commented):
Two things, the second of which can turn a refutation into a pass.

The prefix claim does not hold. "A re-run with the same seed and enough time replays the same slices in the same order" needs deterministic slice boundaries, but `slice_n` is derived from the measured rate (`FAST_SLICE_SECONDS / per_trial`), so a faster or less loaded machine gets different boundaries, the per-slice seeds `seed + i*1_000_003` then cover different trial counts, and the permutation streams diverge. The receipt says "reproduce this run with seed N" and that stops being true. Making the slice schedule a function of the target rather than of elapsed time (fixed size, or a fixed geometric ladder) restores it and still bounds the overshoot, since the deadline check stays between slices.

Only the lightest proposal is validated, and slicing multiplies the chances of a bogus one. `best_w`/`best_sup` keep the minimum across slices, then `if not valid: return False, None, None, done`. So if slice 5 proposes a genuine weight-28 logical against a claim of 31 and slice 40 proposes a bogus weight-25, the bogus one becomes `best`, validation fails, and the pass reports no refutation: the real over-claim merges. One call meant one chance for that before; ninety slices mean ninety. Keeping every proposal under the claim and validating in ascending weight until one passes costs nothing and removes the failure mode. The new test only covers a bogus find on call 1 with nothing valid behind it, so it would not catch this.

Checked the one thing that looked like a capture bug and it is fine: the post-loop validation uses `best_sup`, not the loop's last `support`.

vprusso (approved):

.github/workflows/verify.yml
```diff
@@ -48,7 +48,18 @@ jobs:
 
   verify:
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    # Budget: the refutation gate's accelerated pass is wall-clock capped at 90
```
vprusso: I'm always a bit weary of long-winded comments like this above a simple line. Perhaps we can ask the agent to condense the language here to be more succinct.

verify/gate_changed.py
```diff
@@ -82,6 +85,26 @@
 # printed seed alone (the per-thread RNG streams depend on the split).
 FAST_THREADS = 4
 
+# Wall-clock cap on the fast pass alone (issue: the 8M-trial target has no time
```
vprusso: Similar comment here on the verbosity. It's saying a lot which is hard to parse and likely can mostly be cut.

verify/test_refute_gate.py
```diff
@@ -338,3 +338,70 @@ def test_main():
...
+def test_fast_pass_wall_clock_cap(monkeypatch):
+    """The fast pass stops at its trial target or wall-clock cap, whichever first.
+
+    It reports the trials it actually completed. Regression for the [[924,18,31]] CI kill: the 8M-trial fast pass had no time
+    bound of its own, so at n~900+ it outgrew the whole job and the run died
+    with no output and no receipt. Uses a fake accelerator so the test needs
+    neither gf2_fast nor real wall-clock; the pass is driven in slices, so the
+    fake also checks that the slices add up to exactly what is reported and
+    that a hit surfaced in a later slice is still validated by the python stack.
+    """
```
MathysRennela, line 351: same comment: reference to the killed CI gate won't make sense beyond this PR

MathysRennela (commented):
LGTM except for comments.

FarLab:
Thanks, both were real and the second was a soundness regression relative to main. Fixed in 5e626638:

- Slice schedule is now a fixed geometric ladder (`fast_slices`: 10k doubling to 200k, then flat) that depends on the target alone, so the same seed replays the same slices and per-slice seeds on any machine and a trimmed run is a genuine prefix. Test drives a slow fake and checks the prefix property directly. Overshoot is bounded by one 200k slice.
- Every proposal below the claim is kept and validated in ascending weight until one holds; a bogus lighter proposal can no longer mask a genuine one. Your slice-2-genuine / slice-3-bogus case is in the test and refutes.
- Timeout comment in `verify.yml` condensed to two lines.

FarLab:
Rebased on main (manifest re-pinned after #1099) and trimmed in f9ffe3a2: the `FAST_SECONDS`/slice comment block is down to what the file needs, and the test docstring no longer narrates the CI run behind it.

PR #1099, opened 2026-09-16
verify_all: re-measure LER claims only for entries the PR changed

FarLab:
On a PR run the "verify all submissions" step takes ~13 minutes. Timestamps from #1093's run show where that goes:

| | CI seconds |
|---|---|
| structural verification, 613 codes | 8 |
| LER replica, 32-8-4 | 252 |
| LER replica, 50-6-4 | 181 |
| LER replica, 42-8-3 | 165 |
| LER replica, 25-1-5 | 162 |

Each replica hits the 120 s/basis wall budget on the runner even with `measure_failures` spread over all 4 cores, and none of those entries had been touched by the PR. The result cannot change until their artifacts or stim do.

This adds `--ler-base REF` to `verify_all.py`: the sampled replica runs only for entries whose `codes/<slug>.json` or `circuits/<slug>/` changed since REF, reusing `gate_changed.changed_codes` so the circuits/ mapping is the same one the gate prices by. Unchanged claims print `ler unchanged since base (re-measured on main)` and are re-measured in full on every push to main, which runs without the flag. An unusable ref falls back to re-measuring everything, so a broken diff costs time, never coverage. The cheap circuit-tier checks still run on every entry.

The PR step of `verify.yml` passes `--ler-base origin/${{ github.base_ref }}`. Locally on the full board with the flag: 29 s instead of ~4.5 min, which should turn the 13-minute CI step into about a minute on every submission PR.

Also refactors the script body into `main(argv)` so it can be tested; `test_verify_all.py` covers the routing with fake verifiers and a temp git repo (changed JSON, changed circuit, untouched, unusable ref, and a skipped replica not hiding a circuit failure). Manifest re-pinned.

Note: this touches `validator_manifest.json`, as does #1098; whichever merges second needs a trivial re-pin.


vprusso (commented):
Checked the gap the skip key implies and it is already closed: "the result cannot change until their artifacts or stim do" reads like an assumption, but `check_submission_scope.py` lists `verify/`, `uv.lock` and `pyproject.toml` as critical and rejects a PR mixing those with `codes/`, so a stim bump or a `ler_tools` edit lands with `codes_changed == false` and takes the unflagged step. Worth one clause in the docstring pointing at that, since the safety lives in another file.

One cost the description does not state: for untouched entries this moves LER regression detection from pre-merge to post-merge, and a failing push run on main reverts nothing. Fine as a trade, but it should be written down.

A `circuits/`-only PR still pays the full 13 minutes, because `steps.changes` computes `codes_changed` from `codes/` alone and so routes to the unflagged step. Safe direction, just not the saving you documented.

FarLab:
Thanks, all three taken in 1ca8773 / 192d8346:

- Docstring now points at `check_submission_scope.py` as the reason an unchanged claim cannot go stale here, and states the cost: for untouched entries LER regression detection moves to the post-merge push run, which reverts nothing on its own.
- The circuits-only routing turned out to be more than a missed saving: `codes_changed` from `codes/` alone meant a circuits-only PR also skipped the scope check, the authorship check, the trusted gate and receipts, so the #1012 mapping never fired in CI for the PRs it targets. Change detection now includes `circuits/` (and excludes it from `only_codes`), and `is_code_submission` treats `circuits/<slug>/` as code data so it cannot be paired with a verifier edit either. Circuits-only PRs get the `--ler-base` skip as a side effect.

vprusso (approved):

Issue #966, opened 2026-09-09
qldpc submit default --fast-trials accelerator pass balloons memory to OOM (and sustained thermal load)


MathysRennela:
# `qldpc submit` default `--fast-trials` pass balloons memory to OOM (and sustained thermal load)

## Summary

`./qldpc submit` runs a 2,000,000-trial accelerator pass (`gf2_fast.distance_rand_witness`, 8 threads, pair_depth 8) by default after its 20k-trial Python RIS pass. On this pass, resident memory grows without an effective bound on macOS to the point of system-wide "application memory" pressure — the editor and other apps get OOM-killed, the machine swaps heavily and overheats (fans at max for the duration). Killing the pass is not possible via `ulimit -v` on macOS (Darwin does not enforce `RLIMIT_AS`; `ulimit -v` returns "Invalid argument"), so the only reliable mitigation we found is `--fast-trials 0`.

## Environment

- macOS (Darwin, arm64), Python 3.12, `gf2_fast.cpython-312-darwin.so` built from this repo's `verify/gf2_fast.cpp`
- Reproduced across submissions at n = 144, 216, 288 (BB-type codes, 72-144 checks/side)

## Reproduction

```
./qldpc submit code.npz --authors @handle --construction "..." --out codes
```

with the default `--fast-trials 2_000_000`. Watch Activity Monitor (or `ps -o rss=` in a loop) while the "accelerator pass (2000000 trials)..." step runs: RSS climbs far past what the matrices (a few MB at these sizes) justify, system-wide memory pressure goes critical, and the fan spins to maximum. The submit itself is sometimes killed by the OS before writing.

Isolation experiment: the identical submit with `--fast-trials 0` completes cleanly and quickly — a wrapper watchdog that kills the process above 3.5 GB RSS never fired. The Python-side steps (20k-trial RIS witness search, full local verifier with refutation) are memory-flat. The accelerator pass is the sole source of the pressure.

## Likely mechanism (from reading `verify/gf2_fast.cpp`)

`min_logical_weight_rand_core` (line ~370) allocates per trial, per thread:

- `GF2Matrix red = rref_perm(K, perm);` — a freshly allocated reduced matrix every trial (plus whatever `rref_perm` allocates internally),
- `std::vector<int> weights(nred)` and, for pair candidates, `std::vector<int> idx(nred)` + partial-sort scratch — every trial.

[...]

PR #1100, opened 2026-09-16
Share one memoized structural pass over the board per process

FarLab:
Profiling the self-test step showed the board being structurally re-verified from scratch three times per pytest session, ~15-20 s each locally and about double that on the runner, all producing identical reports:

- `test_cli_targets` via `build.load_entries`
- `test_frontier_summary` via the CLI's frontier section, which also calls `build.load_entries`
- `test_receipt` via `validate_candidate._board_entries`

`validate_candidate` already had a per-process `lru_cache` keyed on a board stamp (names, mtimes, sizes), but `build.load_entries` did its own uncached pass and the two never shared.

This adds `qldpc_verify.board_reports(code_dir)`: one structural pass over `codes/*.json`, keyed on a sha256 of the file names and the exact bytes the pass verifies, returning `(path, slug, doc, report, size_error, load_error)` per file. Any change to the bytes re-verifies the board, including a same-size edit with the mtime restored, so nothing can serve a stale report; structural verification (`refute=False`) is deterministic, so the cached report is exactly what a fresh call computes. `validate_candidate._board_entries` and `build.load_entries` both read from it. The site still warns and skips oversize files; on an unparseable file it now raises a ValueError naming the slug instead of a raw JSONDecodeError (a deliberate improvement); the validator ignores both as before. Entries are shared and documented read-only; neither caller mutated them before.

| | before | after |
|---|---|---|
| the three board-loading tests, locally | 58.4 s | 19.1 s |
| second `load_entries()` call in a process | ~18 s | 0.1 s |

One cache slot per process (a 619-entry snapshot is ~62 MB; see #966). Full `run_tests.py` passes. The new test covers the memo hit, invalidation on a board change including a same-size/same-mtime edit, broken-file recording, and two spellings of the same directory sharing the slot. Manifest re-pinned for `qldpc_verify.py` and `validate_candidate.py`; `site/build.py` is outside the closure.

Note: touches `validator_manifest.json` like #1098 and #1099; whichever merges later needs a trivial re-pin.


vprusso (commented):
The sharing is the right shape and the three call sites genuinely were doing identical work. Two things about the memo key and one about its size, the first of which contradicts a claim in the description.

The stamp does not guarantee freshness. "An edited or added file changes the stamp and re-verifies the board, so nothing can serve a stale report" holds only if every edit moves the mtime or the size, and an edit that preserves both is reachable through ordinary tooling: `cp -p`, `rsync -t`, `touch -r`, and anything that restores timestamps after writing. Demonstrated on this branch with a one-file board:

```
first pass:   ok=True  claimed k=6
after edit:   size 6031==6031, mtime unchanged: True
cache hit (same object): True
fresh verify of disk:  ok=False  failed=['k_matches_claim']
cached report says:    ok=True  claimed k=6
```

The edit was `"k": 6` to `"k": 5`, same byte width, followed by `os.utime` restoring the original mtime. The memo then serves `ok=True` for a file the verifier now rejects, and the cached `doc` still reports the old `k`. In CI the exposure is small, since the board arrives from a fresh checkout and nothing restores mtimes mid-process. Locally it is reachable, and this is the manifest-pinned stack that `validate_candidate` builds receipts from, so a stale board read is a wrong duplicate or novelty verdict rather than a slow test.

There is a second, narrower version of the same problem even when mtimes do move: the stamp is taken in `board_reports` and the files are read again inside `_board_reports_cached`, so the two are not atomic and the cached value can correspond to content the key does not describe.

Both close with the same change, and it is nearly free here: the pass already reads every file, so key the memo on a hash of the bytes it read rather than on metadata about them. Hashing 22 MB of `codes/` costs milliseconds against the 30 seconds the verify pass takes on the current 619-entry board, and it makes the freshness claim in the description true as written instead of true-under-assumption.

On size: one snapshot measured at 62 MB resident on the current board (619 entries, 30 s for the pass), so `lru_cache(maxsize=4)` retains up to about 250 MB of board snapshots per process. The board only moves forward within a process, so I do not think the extra three slots buy anything, and #966 is an open OOM report against `qldpc submit`. `maxsize=1` saves roughly 190 MB and your new test still passes with it, which I checked: the identity assertion `r3 is not r1` holds because the test keeps its own references, and the call counts are unchanged.

Smaller notes. `board_reports` keys on `os.path.abspath(code_dir)` but computes the stamp from the unnormalized `code_dir`, so two relative paths to the same directory produce two identical passes rather than one; harmless, and fixed by normalizing once at the top. And `site/build.py` changes an unparseable board file from an uncaught `json.JSONDecodeError` to a `ValueError` carrying the slug, which is a better message, but the description says callers keep their previous handling, and this one is a deliberate improvement worth naming as such.

The read-only contract on shared `doc` and `report` is documented in the docstring, which is the right place. Neither current caller mutates, and I checked that `build.load_entries` only reads.

FarLab:
Thanks, the stale-read demonstration is right and my freshness claim was wrong as written. Taken in 5114b5dc:

- Memo keyed on a sha256 of the file names and the exact bytes the pass verifies (oversize files are never read; a marker is hashed instead). Your same-width edit + `os.utime` case is now in the test and re-verifies. Hashing the board costs ~80 ms on a hit.
- One cache slot instead of `lru_cache(maxsize=4)`, and `code_dir` normalized once so two spellings share it.
- On `site/build.py`: agreed, the `ValueError` with the slug on an unparseable file is a deliberate improvement over the raw `JSONDecodeError`, not preserved handling. I'll fix the description.

vprusso (approved):

FarLab:
Rebased on main (manifest re-pinned after #1099); no functional change since a632161a.

PR #1179, opened 2026-09-18
Add optional search_budget provenance (schema 0.3)

vprusso:
## Add optional search_budget provenance (schema 0.3)

The board records who found a code and with which model, but not what the
search cost. Nothing on an entry says how many candidates were screened, how
many RIS trials, CPU-hours, GPU-hours or LLM tokens went into finding it, or
how long the search ran. The one budget field that exists,
`witness_provenance.found_at_samples`, is about the witness search for a
single operator rather than the code search, and it is present on 70 of 604
entries. So the question the board is best placed to answer, whether the cost
of a frontier-advancing code is going up or down as the tooling matures,
cannot be reconstructed from the data it holds.

This adds an optional `provenance.search_budget` object as schema version
0.3, with `candidates_screened`, `ris_trials_per_side`, `cpu_hours`,
`gpu_hours`, `llm_tokens` (model name to integer tokens), `wall_clock_hours`,
`tool` and `notes`, all optional, no extra keys. A document that carries the
block must declare `0.3`; 0.1 and 0.2 files are valid unchanged (all 663
files under `codes/` still validate), and a 0.3 file may carry the 0.2
features, so the version rules read "0.2 or later" where they used to read
"0.2". The verifier checks none of the new fields, since there is nothing
trustless to check; they are self-reported provenance, like `model`.
`./qldpc submit` fills the block from new `--budget-*` flags or a
`--budget-json` file (flags override file keys) and leaves the document at
0.1 when no budget is given. `schema/SCHEMA.md` documents the fields.
`cli/test_cli_budget.py` covers the flags, the file path, the version tie
and the backward compatibility, and `run_tests.py` now collects `cli/` so it
runs in CI.

Per CONTRIBUTING.md this is a validation-code change and is separate from any
code submission. One maintainer step is needed: `schema/code.schema.json` is
in the trusted-closure manifest, so `verify/check_validator_integrity.py`
reports the new hash as drift until it is re-pinned with `--update`; that
edit is under `verify/` and is left to review rather than made here.

Not in this PR: the site does not render the block yet, and
`research/kit/search.screen` does not return counters for it. Both are
follow-ups once the shape is agreed.


MathysRennela (approved):
LGTM! But I will wait for @FarLab's review before merging.

MathysRennela (approved):

PR #1189, opened 2026-09-18
Circuit tier: [[63,3,5]] memory circuits at full d_circ = 5/5

vprusso:
Circuit-tier entry for the [[63,3,5]] code (`codes/63-3-5.json`) at full circuit distance, d_circ = 5/5. First circuit block for this entry; credited in `circuit.contributed_by`, no other field of the entry changes.

- `circuits/63-3-5/memory_x.stim`, `circuits/63-3-5/memory_z.stim` and the committed `circuits/63-3-5/memory_x.dem`, `circuits/63-3-5/memory_z.dem` (stim 1.16.0, canonical noise recipe, rounds = 5, 4589 / 4602 DEM mechanisms in X / Z, cap 25000).
- `codes/63-3-5.json`: `circuit` block with witnessed d_circ.X = 5, d_circ.Z = 5; schema 0.1 to 0.2.
- Schedule: sequential two-block extraction, one ancilla per check (60 ancillas, 123 qubits), X checks then Z checks each round with joint reset and measurement layers; 4 X-check and 4 Z-check CX layers per round. CX layers are a proper edge coloring of each check type's Tanner graph (bipartite, so max-degree colors suffice), found by the alternating-path algorithm with a seed (seed 0) that shuffles edge order and permutes colors. CNOT orders are filtered by a hook-error criterion: for each check and each proper suffix S of its CNOT order, the fewest further data errors that complete S to a logical is computed by an observable-constrained BP+OSD decode, and an order is admitted only if every suffix needs at least d - 1 of them, so a single ancilla hook costs at least d faults. Layer assignment over admitted orders is a min-conflicts search.
- Witness: the code-tier logical lifted into the DEM as data errors right after preparation, so d_circ <= d is witnessed by construction; the searches below try to refute it and found nothing lighter.
- Refutation search on the committed DEM, two methods per basis: RIS on ker(H_dem), 2 x 500 trials, lightest weight 5 (X); 2 x 500 trials, lightest weight 5 (Z). Observable-constrained seeded BP+OSD search (for each seed mechanism j and observable o, the lightest undetected set containing j that flips o), 4589 of 4589 mechanisms, lightest weight 5 (X); 4602 of 4602 mechanisms, lightest weight 5 (Z). Every find was validated in GF(2) before it counted.
- No `ler` block. Measured locally at p = 0.001 with the pinned decoder `bposd-cs-10`: X 4.37e-04 per round (95% CI 3.88e-04 to 4.94e-04, 262 failures in 120000 shots, seed 7), Z 4.81e-04 per round (95% CI 4.28e-04 to 5.40e-04, 288 failures in 120000 shots, seed 7). CI failed `X_ler_replicated` and `Z_ler_replicated` on the first push, while `verify/ler_verify.py` replicates both within 4 sigma here with 18 workers (52525 and 33022 replica shots) and with 4 workers (38646 and 50791). The log does not say which branch of the check tripped; the likely one is the budget bound, since at a per-shot rate near 2.2e-3 the factor-2 admissibility test needs about 26000 replica shots per basis inside the 120 s window and a hosted runner is slower per core than this machine. Either way the block cannot merge unchecked, so the rates are recorded here instead, as for 101-5-5 and 126-6-5 in this batch.
- `verify/circuit_verify.py`, `verify/qldpc_verify.py` and `verify/check_prose.py` pass locally; `verify/check_authorship.py --author vprusso` binds through the first-circuit-block rule.
- The schedule generator and the hook-cost search are not committed; the method above is enough to rewrite them (the circuit builder is `verify/circuit_tools.py`'s `build_css_memory` shape with the per-check CNOT orders chosen as described).


MathysRennela:
@vprusso this was flagged as a duplicate code

Issue #899, opened 2026-09-05
codes/396-10-39.json distance 39 is over-stated: a weight-37 logical exists


vprusso:
`codes/396-10-39.json` claims `distance.d = 39`. A nontrivial logical operator of
weight 37 exists, so the claim is over-stated by 2 and the true distance is at
most 37.

The witness is an X-type logical, 37 qubits:

```
[1, 7, 12, 13, 19, 25, 31, 37, 43, 49, 55, 61, 90, 113, 115, 151, 180, 246, 266,
 272, 278, 284, 290, 296, 308, 314, 320, 326, 332, 338, 350, 356, 362, 378, 380,
 386, 392]
```

Checked against the entry's own parity checks, with the nontriviality confirmed a
second way by rank rather than only by the routine that produced it:

| check | result |
|---|---|
| weight | 37 |
| `H_Z v = 0` | true, syndrome weight 0 |
| syndrome weight against `H_X` | 84, so it is not an X-stabiliser |
| `rank(H_X)` vs `rank(H_X + v)` | 193 vs 194, so `v` is not in the rowspace |

### Reproduction

```
python verify/heuristic_distance.py codes/396-10-39.json --fast-trials 8000000 --seed 1
```

Exits 2 and reports `"verdict": "refuted"`, `"d_heuristic": 37`, with the same
X-side witness.

### Context, and what this is not

This is the second over-claim I have found while screening the board, and it is
worth stating the denominator. Of the entries screened at these budgets, only two
did not hold up: this one and `codes/684-12-81.json` (issue #896). Everything
else stood, including several with much higher figures of merit:

| entry | claim | readings at 200k / 2M / 8M | verdict |
|---|---|---|---|
| [[684,14,78]] | 78 | 91, 79, 79 | stands |
| [[684,12,77]] | 77 | 93, 89, 83 | stands |
| [[684,10,101]] | 101 | 109, 108, 107 | stands |
| [[400,12,50]] | 50 | 56, 54, 50 | stands |
| [[390,68,28]] | 28 | 31, 31, 29 | stands |
| [[674,170,76]] | 76 | 77, 77, 77 | stands |

Two of eight, and the six that held are as much the result as the two that did
not.

### The budget point, which generalises

The margin here is small, 37 against 39, and the reading was stable from 200000
trials upward, so a modest budget increase finds it. That is not true of the
other case: on `codes/684-12-81.json` the search returns exactly the claimed 81
[...]

PR #1285, opened 2026-09-19
Fieldnote: audit the weight-8 leaders before targeting them ([[360,12,24]] holds, [[684,20,72]] is soft)

e-eight:
## Fieldnote: audit the weight-8 leaders before targeting them

**One fieldnote, no code** — scoped down from the earlier revision of this PR as
requested in review. The audit harness that came with it is now proposed
separately; nothing in this PR is needed to reproduce the finding, and the
recipe below uses tools already in the tree.

### Follow-up (2026-09-20): the audit was under-searching

A second pass over the same family returned *inconclusive* on every entry, and
the cause was the instrument rather than the codes: the audit drove the
accelerator at its `pair_depth` default of 10 while the ladders behind these
claims used 24 to 80. At a fixed 200,000 trials on the same entry, depth 10 reads
97 where depth 64 reads 87, and depth 64 costs about 1.4x rather than 45x. With
that corrected, two more leaders collapsed — `[[684,8,85]]` to `[[684,8,63]]`
(PR #1666) and `[[684,12,77]]` to `[[684,12,71]]` (PR #1667) — and the fieldnote
now records that the CI refutation gate itself runs at depth 8, shallower than
the ladders that produce the claims it is meant to check.

### Finding

Before aiming at a bar, re-measure it. The board's distances are witness-backed
**upper bounds**, and the board's own history says some are inflated
(`[[882,18,30]] -> 29`, `[[684,12,81]] -> 66`, `[[396,10,39]] -> 37`). This
audit re-measured the leaders of the cells around `local-2d-bilayer x weight-8`
on fresh seeds, and both outcomes turned out useful:

* **`local-2d-bilayer x weight-8` holds.** Its leader `[[360,12,24]]` is a
  reconstructed literature baseline whose paper-level distance is probabilistic
  for `d > 20` — exactly the shape of claim that has gone soft before. Ten fresh
  seeds over 80M cumulative RIS trials return 24 and never anything lighter, and
  the independent decoder mechanism agrees at the only weight it reaches (32).
  The bar to beat is a real `kd^2/n = 19.20`.
* **The `unrestricted x weight-8` leader was soft.** `[[684,20,72]]` collapses to
  `d <= 48` at 8M trials. The witness was re-checked with a from-scratch NumPy
  GF(2) rank (syndrome 0 against `H_X`, and `rank(H_Z)` 332 vs 333 with `v`
  appended, so it is outside the stabilizer row space). The entry is revised
  separately as `[[684,20,48]]`, moving that cell's headline from
  `kd^2/n = 151.58` to `67.37`.

Screening results at 2M trials, one seed:

| entry | n | k | w | claim | RIS at 2M | reading |
|---|---|---|---|---|---|---|
| 672-20-32 | 672 | 20 | 6 | 32 | 32 | corroborated |
| 682-182-76 | 682 | 182 | 32 | 76 | 76 | corroborated |
| 584-150-18 | 584 | 150 | 8 | 18 | 18 | corroborated |
| 922-18-31 | 922 | 18 | 8 | 31 | 32 | inconclusive |
| 684-20-72 | 684 | 20 | 8 | 72 | 84 | refuted at 8M |

### The calibration point

`[[684,20,72]]` read **84** at the 2M triage budget — twelve units *above* its
claim, and the least alarming-looking row in the table. It is the row that was
soft. A RIS reading above the claim carries no information in either direction,
so for these low-rate entries the flat-looking readings are not evidence of
health. The refuting witness is also rare rather than merely deep: seed 71
reaches 48 at 8M while fresh seeds at 10M return 56 and 70, which is how an
honest 20M-trial validation could report 72.

This generalises the budget point in issue #899 with a concrete counterexample
from the opposite direction.

### Follow-up (2026-09-19)

The follow-up commit adds the next audit in the same family. With
`[[684,20,72]]` revised, the `unrestricted x weight-8` leader became
`[[684,14,78]]` (same affine construction, `kd^2/n = 124.53`); a fresh-seed
ladder refutes it at 8M trials / seed 71 with a weight-72 X-logical, six below
the claim, and it is revised separately to `[[684,14,72]]`
(`kd^2/n = 106.11`). The control points the other way: `[[684,10,101]]` reads
106 against a claim of 101 at both 2M and 8M, which is *inconclusive*, not
corroboration. The updated fieldnote records both, with the ladder tables.

### Checklist
- [x] One fieldnote under `fieldnotes/`; no code and no harness in this PR
- [x] Numbers come from the runs quoted in its Reproduction section, which uses `verify/heuristic_distance.py` and `decode/distance.py`
- [x] No code entry added, changed or removed by this PR
- [x] Not equivalent to any existing entry; this PR submits no code
- [x] Follow-up audit of `[[684,14,78]]` and `[[684,10,101]]` recorded in the same fieldnote


MathysRennela (changes requested):
can you commit only one field note with this PR? Or make this PR field notes only.

e-eight:
Done — this PR is now a single fieldnote. The harness moved to its own PR (#1321), and the reproduction recipe here now uses tools already in the tree (`verify/heuristic_distance.py` for the RIS ladder, `decode/distance.py` for the decoder cross-check), so nothing in this PR depends on the other.

While splitting I also fixed a latent dangling reference: the harness README had cited `codes/684-20-72.json`, which the sibling revision PR renames away. It now names the entry by its `[[n,k,d]]` tag, so the tree stays consistent whichever order the two land in.

Issue #1148, opened 2026-09-17
Drop the BP+OSD syndrome-decoder cross-check from the refutation gate: dominated by RIS at every budget


FarLab:
## Summary

The optional BP+OSD syndrome-decoder cross-check in `verify/gate_changed.py` (`decode/distance.py:refute_check`) should be removed from the per-PR refutation gate. Measured against every historical over-claim in the board's history, it never finds a lighter logical than RIS at equal wall-clock, runs 9× to 160× fewer trials per second, and its 10 s budget is currently not enforced on exactly the codes where it costs the most. PR #948 proposes bounding it in a child process; the measurements below suggest the check is not worth bounding.

## 1. The budget is not enforced (root cause of #937 / #948)

`_side` in `decode/distance.py` consults its deadline only when `(t & 255) == 0`. Decode cost depends on the injected weight, not n, so:

| code | one decode | `refute_check(max_seconds=10)` actual |
|---|---|---|
| 169-8-9 | <1 ms | 10.0 s |
| 634-6-11 | 1 ms | 10.3 s |
| 653-114-3 | <1 ms | 10.4 s |
| 600-8-96 | 1.6 s | >190 s (killed) |
| 674-86-89 | 2.3 s | >190 s (killed) |

At 2 s per decode, 256 decodes between deadline checks is over 8 minutes per side. That is the stall #948 attributes to a non-returning decode call; single decodes return in about 2 s. Checking the deadline every trial brings the two large codes to 25.5 s and 19.3 s (the remainder is six decoder constructions outside the deadline). So the failure has a one-line fix, if the check were worth keeping.

## 2. 60 s head-to-head on all 20 historical over-claims

Every `codes/<n>-<k>-<d>.json` that was ever renamed to a lower d was reconstructed from git history at its over-claimed version. Each got 60 s of Python RIS (`heuristic_distance.refute_check`, trials uncapped) and 60 s of BP+OSD (with the per-trial deadline fix so the budget is real), same seed, same machine.

| over-claim | claimed | true ≤ | RIS found | RIS trials/s | BP+OSD found | BP+OSD trials/s |
|---|---|---|---|---|---|---|
| 288-8-36 | 36 | 35 | 40 | 380 | 53 | 10.9 |
| 288-8-41 | 41 | 36 | 40 | 381 | 55 | 10.8 |
| 312-10-44 | 44 | 26 | 47 | 290 | 62 | 9.1 |
[...]

PR #1321, opened 2026-09-19
Add research/audits: a harness for re-measuring distance claims before targeting them

e-eight:
## Tooling: a reusable harness for auditing distance claims

One new directory, `research/audits/`, with the harness that produced the
fieldnote of 2026-09-18. Split out of that PR on request: this one is tooling
only, no fieldnote, no code entry.

### Why it exists

The boards rank witness-backed distance **upper bounds**, so the first step of a
campaign is to re-measure the leaders on fresh seeds rather than build against a
number that may be soft — `[[684,20,72]]` turned out to over-claim by 24.
`leader_audit.py` does that re-measurement and nothing else:

```
# escalating ladder on one entry, fresh seeds each rung
uv run --frozen python research/audits/leader_audit.py ladder \
    codes/360-12-24.json \
    --ladder 1000000:101,102,103,104 5000000:201,202,203 20000000:301,302,303

# one budget across a cell's leaders, for triage
uv run --frozen python research/audits/leader_audit.py screen \
    --trials 2000000 --seeds 51 codes/672-20-32.json codes/922-18-31.json
```

It exits 2 if any claim is refuted, so it can gate a script.

### What it adds over `verify/heuristic_distance.py`

`verify/heuristic_distance.py` is the trusted, server-side search and stays the
tool of record. This harness is the campaign-side wrapper around the same
underlying search, and adds three things it needs:

* **Both sides jointly, on the entry as the board ranks it.** It loads
  `codes/<slug>.json` directly, so a reduction or a layout is measured, and it
  recomputes `k` and CSS commutation rather than trusting the entry's fields.
* **Ladder ergonomics.** A `ladder` spec of `budget:seed,seed` rungs, which is
  the shape that matters: the refutation above came from one 8M-trial rung while
  fresh seeds at 10M and 20M read 56 and 70. A flat multi-seed budget would have
  missed it.
* **A documented three-way verdict** — `refuted` / `holds` / `inconclusive` —
  with the trap written down: for dense low-rate entries RIS can sit *above* the
  claim at a budget that refutes a structured one, so `[[684,20,72]]` read 84 at
  2M and was still the soft entry. A reading above the claim is never
  corroboration.

Every witness is re-validated in pure Python against the raw sparse matrices
before it is recorded (support size equals the weight, `H_opp v = 0` over GF(2),
and `v` outside `rowspace(H_own)`), so an accelerator bug cannot put an unbacked
number in a log. `research/audits/README.md` documents the above and the limits
(the `holds` verdict never upgrades a claim to the `d=` tier; the decoder
cross-check is corroboration only, per issue #1148).

### Follow-up: the witness now outlives the run

Using the harness in the field exposed three gaps, fixed here before the next
sweep inherits them:

* **Witness loss on a killed rung.** `--witness-out` was written once, after the
  whole ladder finished. An 8M-trial rung at `n = 684` takes about fifteen
  minutes, and one was killed mid-rung with its support already in hand but not
  on disk; it had to be re-run to capture the witness the revision needed. The
  support is now printed and written **on every new best**, atomically, so the
  lightest reading so far is always recoverable. `screen` gains the matching
  `--witness-dir`, since a triage screen can refute a claim too.
* **An unvalidated proposal could still drive the verdict.** `best` was updated
  before the result of the independent re-check was consulted, so a witness that
  failed `validate_witness` could refute a claim and be written out — the exact
  failure the re-check exists to prevent, and contrary to what the docstring and
  README already claimed. It is now reported as `DISCARDED` and ignored.
* **`pair_depth` was hidden, and the default under-reads against this board's own
  ladders.** Each trial combines the `pair_depth` lightest reduced rows pairwise;
  the accelerator defaults to 10, but the ladders behind the affine two-block
  entries used 24 to 80. At 200,000 trials and the same seed on `[[684,12,77]]`,
  depth 10 reads 97 while depth 64 reads 87 — and depth 64 costs about 1.4x, not
  45x, because the per-trial cost is dominated by the elimination rather than the
  pair phase. Without the knob an `inconclusive` verdict says more about the
  candidate set than about the code. `--pair-depth` is now exposed on both
  subcommands (default unchanged at 10), and each run prints the depth and thread
  count it used. That entry has since been revised to `[[684,12,71]]` on the
  strength of a depth-64 rung — the calibration is not hypothetical.

Five regression tests in `research/audits/test_leader_audit.py` pin these
properties, including that `--pair-depth` reaches the search (they stub the
search and run in 0.2 s, with or without `make fast`).

### Checklist
- [x] Documentation and tooling only; no code entry added, changed or removed
- [x] The harness runs on entries already in the tree and is smoke-tested (`ladder`, `screen`)
- [x] `ruff check` and `ruff format --check` clean; `verify/check_prose.py` exits 0
- [x] Not equivalent to any existing entry; this PR submits no code


e-eight:
Rebased onto current `main` and pushed a follow-up commit: the witness is now printed and written on every new best (so a ladder rung killed by a time limit no longer costs a re-run), `screen` gained `--witness-dir`, and a proposal that fails the independent GF(2) re-check is discarded instead of being counted — previously such a proposal could drive the verdict and reach the witness file. Four regression tests added. Ready for review; I don't have the permissions to request one.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,303 @@
...
+candidate set, not the code: on `[[684,12,77]]` at 200,000 trials and the same
+seed, depth 10 reads 97 while depth 64 reads 87 (and 64 costs about 1.4x, not
+45x). An `inconclusive` verdict taken at a shallower depth than the claim's own
+ladder is an artifact and must not be reported as evidence.
+
+Both are upper-bound searches. `refuted` is decisive (a witness is exhibited);
+`holds` means "this search, at this budget, found nothing lighter" and is not a
+proof. It says nothing about the exact (`d=`) tier, which needs
+`verify/certify.py`.
+"""
```
MathysRennela, line 33: please trim down the docstring to be minimal and self-contained, to avoid bloating the context of agents.

MathysRennela (approved):
LGTM after the changes I requested.

research/audits/leader_audit.py
e-eight: Done in cfe79aaf — the module docstring is down from 51 to 29 lines and now covers only what is needed to run the tool correctly: what it audits and why, the three verdicts, the two operational musts (`--witness-out`/`--witness-dir`, and `--pair-depth` matched to the claim's own ladder), one example per subcommand, and the exact-tier caveat. The board history and the depth calibration moved to `research/audits/README.md`, which is also where the verdict table lives.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+            threads=int(threads),
+        )
+        w = surrogate._weight_or_inf(w, HX.shape[1])
+        support = sorted(int(q) for q in support) if support else []
+        side = {"X": "X", "Z": "Z"}.get(side, side)
+        if side in ("X", "Z"):
+            return w, side, support, time.time() - t0
+    # NumPy fallback, or an accelerator proposal that did not validate.
+    prepared = surrogate.prepare_distance_search(HX, HZ)
+    wx, sx = surrogate._search_lightest(*prepared.side("X")[:2], trials, seed, bases=prepared.side("X")[2:])
```
FarLab: **`--pair-depth` never reaches the NumPy fallback.** `_search_lightest` takes `pair_depth=10` as a keyword, and neither call here passes it, so without `make fast` both sides search at depth 10 while the run header prints the requested depth.

Confirmed locally: with the accelerator absent, wrapping `surrogate._search_lightest` and calling `ris(..., pair_depth=64)` records `pair_depth` = `[10, 10]`. `test_pair_depth_reaches_the_search` stubs `la.ris` wholesale, so it cannot see this. Fix: pass `pair_depth=pair_depth` on both calls, and have the test spy on `_search_lightest` (or the accelerator call) rather than on `ris`.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+            np.asarray(HZ, dtype=np.int8),
+            trials=int(trials),
+            seed=int(seed),
+            pair_depth=pair_depth,
+            threads=int(threads),
+        )
+        w = surrogate._weight_or_inf(w, HX.shape[1])
+        support = sorted(int(q) for q in support) if support else []
+        side = {"X": "X", "Z": "Z"}.get(side, side)
+        if side in ("X", "Z"):
```
FarLab, line 108: **A no-logical result from the accelerator falls through into the pure-NumPy search at the full trial budget.** If `side` is not X/Z (k=0 entry, or the n+1 sentinel), control drops to the fallback below with e.g. `trials=8_000_000`. Locally the NumPy path runs at roughly 1 ms/trial on n=72, so that is hours, not a retry. The comment below says this also covers a proposal that did not validate, but no validation happens at this point. Suggest returning `(inf, "", [], dt)` here and reserving the fallback for `surrogate._fast is None`.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+            return w, side, support, time.time() - t0
+    # NumPy fallback, or an accelerator proposal that did not validate.
+    prepared = surrogate.prepare_distance_search(HX, HZ)
+    wx, sx = surrogate._search_lightest(*prepared.side("X")[:2], trials, seed, bases=prepared.side("X")[2:])
+    wz, sz = surrogate._search_lightest(*prepared.side("Z")[:2], trials, seed + 1, bases=prepared.side("Z")[2:])
```
FarLab: Seeding the Z side with `seed + 1` makes consecutive ladder seeds share streams: with `--ladder 1000000:101,102,103`, Z of seed 101 and X of seed 102 both use `default_rng(102)`. The rung is described as fresh independent seeds. Something like `np.random.SeedSequence(seed).spawn(2)` avoids the overlap. The fast and NumPy backends also map a seed differently, so a recorded `seed` does not reproduce a reading across backends; worth a line in the README.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+            pair_depth=pair_depth,
+            threads=int(threads),
+        )
+        w = surrogate._weight_or_inf(w, HX.shape[1])
+        support = sorted(int(q) for q in support) if support else []
+        side = {"X": "X", "Z": "Z"}.get(side, side)
+        if side in ("X", "Z"):
+            return w, side, support, time.time() - t0
+    # NumPy fallback, or an accelerator proposal that did not validate.
+    prepared = surrogate.prepare_distance_search(HX, HZ)
```
FarLab: `PreparedSearch` (kernel and logical bases) is rebuilt on every `ris()` call, and `prepared.side()` is called four times. A 10-seed ladder at n=684 recomputes the GF(2) bases 10 times, which is what `prepare_distance_search` exists to avoid. Build it once in `cmd_ladder` / `cmd_screen` and pass it in.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+    for i, row in enumerate(rows):
+        for q in row:
+            M[i, int(q)] ^= 1
+    return M
+
+
+def ris(HX, HZ, trials, seed, threads=8, pair_depth=10):
+    """One RIS search on both sides. Returns (weight, side, support, seconds)."""
+    t0 = time.time()
+    if surrogate._fast is not None:
```
FarLab, line 97: Design note, not blocking: `ris()` and `validate_witness()` re-implement `surrogate._distance_rand_fast` and `_validate_fast_witness` and depend on private names (`_fast`, `_weight_or_inf`, `_search_lightest`). The underlying gap is that `surrogate.distance_rand` hardcodes `pair_depth=10` and does not return the witness. Exposing `pair_depth=` and a witness-returning variant there would fix it once for every caller and leave a single validator to maintain.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+    if in_rowspace(v, Hself):
+        return False, "lies in the stabilizer row space (trivial)"
+    return True, "ok"
+
+
+def describe(n, k_claim, HX, HZ, doc, tag):
+    k = compute_k(HX, HZ)
+    w = max(int(HX.sum(axis=1).max()), int(HZ.sum(axis=1).max()))
+    print(
+        f"{tag}: n={n} k={k} (claimed {k_claim}) w={w} css_ok={verify_css(HX, HZ)}"
```
FarLab: `k` and `css_ok` are recomputed but only printed. An entry with `css_ok=False`, or a recomputed `k` that differs from the claim, still gets a verdict, can exit 2, and can write a witness, so a gating script would act on a refutation of something that is not a valid CSS code. Suggest a hard error (distinct exit code) when `verify_css` fails or `k != k_claim`.

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+    return k, w
+
+
+def _parse_ladder(text):
+    """'1000000:101,102 5000000:201' -> [(1000000, [101, 102]), (5000000, [201])]."""
+    out = []
+    for rung in text:
+        trials, _, seeds = rung.partition(":")
+        out.append((int(trials), [int(s) for s in seeds.split(",") if s]))
```
FarLab: **A rung with no seeds is accepted and silently skipped.** Confirmed locally:

- `--ladder 2000:1 2000` runs only the first rung; the second performs zero searches with no warning, and the verdict is `holds`.
- `--ladder 2000` (all seedless) prints `d_ub=73 eff(kd^2/n)=888.17 -> inconclusive` and exits 0.

A reader would believe the deep rung ran. Suggest raising `argparse.ArgumentTypeError` when a rung has no seeds (and guarding the `eff` print when `best['weight'] == n + 1`).

research/audits/leader_audit.py
```diff
@@ -0,0 +1,279 @@
...
+                print(f"    NEW BEST d<={weight} side={side} support={support}", flush=True)
+                if args.witness_out:
+                    _write_witness(args.witness_out, _best_payload(args.entry, n, k, claim, best))
+    verdict = "REFUTED" if best["weight"] < claim else "holds" if best["weight"] == claim else "inconclusive"
+    print(
+        f"VERDICT: n={n} k={k} claim={claim} d_ub={best['weight']} "
+        f"eff(kd^2/n)={k * best['weight'] ** 2 / n:.2f} -> {verdict}",
+        flush=True,
+    )
+    if args.witness_out and best["support"]:
```
FarLab, line 308: Witness-file hygiene:

- A file left at `--witness-out` by an earlier run survives a run in which every proposal is discarded or nothing is found, and then reads as this run's artifact (including a stale `verdict: REFUTED`). Consider removing or overwriting it at start.
- In `screen`, entries from different directories with the same basename overwrite each other's `<stem>.json`.
- The fixed `{path}.tmp` name races between two processes writing the same path; `tempfile.mkstemp(dir=parent)` avoids that.

research/audits/test_leader_audit.py
```diff
@@ -0,0 +1,125 @@
...
+        return 2, "X", [0, 1], 0.0  # lighter, but not a logical
+
+    monkeypatch.setattr(la, "ris", staged)
+    out = tmp_path / "witness.json"
+    rc = la.cmd_ladder(_ladder_args(out))
+    assert rc == 0
+    payload = json.loads(out.read_text())
+    assert (payload["weight"], payload["side"], payload["support"]) == (weight, side, support)
+    assert payload["seed"] == 1
+    assert payload["claim"] == 6 and payload["verdict"] == "holds"
```
FarLab: These tests pin a live board entry: `codes/72-12-6.json`, `claim == 6`, verdict `holds`, plus `_genuine_witness` doing real 2000-trial RIS runs that must land exactly on the claimed weight. If that entry is ever revised or renamed, or the NumPy backend reads 8 at this seed, the tests fail with the harness unchanged; this is the failure mode that PR #946 hit and #986 had to repair. A small synthetic CSS fixture with a hand-written witness removes the dependency.

It also makes the docstring's "searches are stubbed throughout" and the PR body's 0.2 s untrue off the accelerator: locally the suite takes 0.2 s with `gf2_fast` built and 5.5 s on the NumPy path.

research/audits/README.md
```diff
@@ -0,0 +1,91 @@
...
+lightest logical seen so far is printed and written **on every new best**, not
+once at the end: a rung can be killed by a time limit, and the witness behind
+the lightest reading is the artifact a distance revision needs. Re-running a
+killed 8M-trial rung at `n = 684` costs about fifteen minutes of wall clock.
+
+## Reading the output
+
+| verdict | meaning |
+|---|---|
+| `refuted` | a logical lighter than the claim was exhibited. The claim is over-stated and `d <= ` the reading. A distance revision is a valid submission on its own. |
```
FarLab, line 76: The documented verdict token is `refuted`, but the code prints and writes `REFUTED`, so a script grepping for the documented string never matches. Pick one. Separately, the module docstring says depth 10 "reports a soft claim as a hold"; from the verdict logic an under-read lands on `inconclusive`, not `holds`.

FarLab (commented):
Thanks for this; a re-measurement harness is useful and the witness re-validation plus write-on-every-new-best are the right instincts. Ten inline comments below.

I ran the PR locally (python 3.10, macOS): 5/5 tests pass, and a 1M-trial ladder on `72-12-6` takes about 3 s with the accelerator. Two of the findings were reproduced by running the harness (`--pair-depth` not reaching the NumPy fallback; seedless ladder rungs silently skipped); the rest are from reading the code.

Suggested blocking set: the `pair_depth` fallback, the accelerator fall-through, the seedless-rung parse, and the tests pinning live board data. The remainder are hardening and cleanup.

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. `ris()` now takes `prepared=` and passes `pair_depth` into both `_search_lightest` calls through a `_search_side` helper (leader_audit.py:83-126). Your reproduction is gone: with the accelerator absent, `ris(..., pair_depth=64)` records `[64, 64]`.

You were right about why the test missed it. It now spies on `_search_lightest` rather than stubbing `ris`, and asserts the accelerator receives the requested depth too, so both backends are pinned rather than one (`test_pair_depth_reaches_both_backends`, test_leader_audit.py:78).

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. The n+1 sentinel now returns `(inf, "", [], dt)` directly, and the NumPy path is reached only when `surrogate._fast is None` (leader_audit.py:110).

Your note that the comment was also wrong was correct -- no validation happens at that point, so I removed that claim rather than rewording it. `test_no_logical_from_the_accelerator_does_not_fall_through` drives an 8M-trial call through a stub accelerator and asserts the NumPy search is never entered (test_leader_audit.py:135).

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. Both sides now draw from `np.random.SeedSequence(seed).spawn(2)` (leader_audit.py:120), so Z of seed 101 no longer replays X of seed 102 -- your example is the assertion, not a paraphrase of it (`test_the_two_sides_use_independent_seed_streams`, test_leader_audit.py:110).

I took the second half as well: the README now states that a recorded `seed` does not reproduce a reading across backends, since the accelerator consumes it directly while the NumPy path spawns from it (README.md:52-55).

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. `PreparedSearch` is built once per entry in `cmd_ladder` and `cmd_screen` and passed down as `ris(..., prepared=prepared)` (leader_audit.py:285, 325).

`test_prepared_bases_are_built_once_per_entry` runs three seeds through both subcommands and asserts exactly one build (test_leader_audit.py:180), so the n=684 case you describe cannot regress silently.

research/audits/leader_audit.py
e-eight: Agreed, and deferred as you suggested -- this belongs with `surrogate.distance_rand`, not in the harness. Exposing `pair_depth=` plus a witness-returning variant there would let `ris()` and `validate_witness()` collapse onto one validated implementation that every caller shares, which is the right place to fix it once.

I have left the duplication for now and recorded the deferral in the commit message rather than letting it evaporate; I will open an issue so resolving this thread does not lose it. Thanks for drawing the line between this and the blocking set.

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. `describe()` returns `css_ok`, and both subcommands route through `_reject_unusable` before any search, verdict, exit code or witness write: non-commuting checks, or a recomputed `k` that differs from the entry's recorded `k`, now abort with exit 3 (leader_audit.py:161-181). `test_entry_that_is_not_the_claimed_code_is_rejected` covers both and asserts no witness reaches disk (test_leader_audit.py:210).

One consequence worth your eye: 3 is a new exit code, because argparse already exits 2 on a usage error and 2 is the documented refutation signal. A mistyped command must not be mistakable for a refutation, so `_Parser.error` and unusable entries both exit 3 (leader_audit.py:348, `test_usage_errors_do_not_exit_two`), and the README now carries an exit-code table.

research/audits/leader_audit.py
e-eight: Fixed in 8ec13cf3. `_parse_ladder` raises `ArgumentTypeError` for a rung with no seeds and `main` converts it to a usage error (leader_audit.py:191-215, 405). Both of your repros are gone: `--ladder 2000` now exits 3 with `ladder rung '2000' has no seeds; expected TRIALS:SEED[,SEED...]`, and `--ladder 2000:1 2000` is rejected before either rung runs.

I also guarded the `eff` print as you suggested -- it is omitted when the best weight is the n+1 sentinel (leader_audit.py:302) -- and rejected `trials <= 0` rungs, which are the same silent no-op in a different spelling.

research/audits/leader_audit.py
e-eight: All three fixed in 8ec13cf3.

- `--witness-out` is removed at the start of a ladder run, so a file left by an earlier run (stale `verdict: refuted` included) cannot read as this run's artifact (leader_audit.py:275). `test_a_stale_witness_is_cleared` plants one and asserts it is gone.
- `screen` now names files by the entry's path relative to the repo, e.g. `codes_672-20-32.json`, so same-basename entries in different directories cannot overwrite each other (`_witness_stem`, leader_audit.py:244; `test_screen_witnesses_do_not_collide_across_directories`).
- `_write_witness` goes through `tempfile.mkstemp(dir=parent)` with cleanup on failure (leader_audit.py:217); the test asserts no `.tmp` files are left behind.

research/audits/test_leader_audit.py
e-eight: Fixed in 8ec13cf3. The suite no longer touches `codes/` at all: a synthetic `[[4,2,2]]` fixture (`H_X = H_Z = [[0,1,2,3]]`, k=2, d=2) with a hand-written weight-2 witness replaces the real 2000-trial RIS runs (test_leader_audit.py:40-46). A revised or renamed entry can no longer fail the suite with the harness unchanged, so the #946/#986 failure mode is designed out rather than repaired again.

Your timing point now lands in the good direction: "searches are stubbed throughout" is true, and the suite is 0.14 s with `gf2_fast` and 0.23 s on the NumPy path, so the PR body's 0.2 s holds on both backends instead of only one.

research/audits/README.md
e-eight: Fixed in 8ec13cf3. Picked lowercase: `holds` and `inconclusive` already were, and both the README table and the module docstring documented lowercase, so the code was the outlier. Verdicts now come from `VERDICT_REFUTED`/`VERDICT_HOLDS`/`VERDICT_INCONCLUSIVE` (leader_audit.py:54-56, 183) so the token cannot drift again, and the README states they are printed and written verbatim (README.md:80).

On the second half you are right -- an under-read lands on `inconclusive`, not `holds`. The docstring now reads: the default of 10 under-reads, "so a soft claim cannot be refuted and lands on `inconclusive`".

e-eight:
All ten findings addressed in 8ec13cf3, with the blocking set you named closed:

- `--pair-depth` now reaches the NumPy search as well as the accelerator, and the test spies on the search it means to pin (`_search_lightest`) instead of stubbing `ris`, so a silent default cannot come back.
- The accelerator's n+1 sentinel no longer falls through into a full-budget NumPy re-run.
- A rung with no seeds (or no trials) is rejected as a usage error rather than running zero searches and reporting a verdict for them.
- The tests no longer read `codes/`: a synthetic `[[4,2,2]]` fixture with a hand-written witness replaces the live board entry and the real RIS runs. The suite is 0.14 s with `gf2_fast` and 0.23 s on the NumPy path, so the PR body's 0.2 s is now true on both backends.

Also in from the hardening set: independent per-side seed streams, `PreparedSearch` built once per entry, verdict tokens lowercase throughout (the documented spelling), stale `--witness-out` cleared at the start, collision-free `--witness-dir` names, `tempfile.mkstemp` writes, and a hard refusal (exit 3) to score an entry whose checks do not commute or whose recorded `k` is not its own `k`.

One thing your review surfaced that I would flag: argparse exits 2 on a usage error, which collided with the documented "2 = refuted" gate, so a mistyped command was indistinguishable from a refutation. Usage errors and unusable entries now exit 3, and the README carries an exit-code table.

The `surrogate.distance_rand` design note (expose `pair_depth=`, return the witness) is deferred to a follow-up so it can land against shared kit with its own tests.

Verification: `ruff check` and `ruff format --check` clean; 14 tests pass with and without `gf2_fast`; CLI smoke tests on `codes/72-12-6.json` and `codes/90-8-10.json` through the real accelerator path.

research/audits/leader_audit.py
e-eight: Follow-up filed as #1832, so this does not get lost when the thread is resolved: expose `pair_depth=` and a witness-returning entry point on `surrogate.distance_rand`, then collapse `ris()`/`validate_witness()` onto it and delete the duplicated validator. Not a blocker for this PR.

FarLab (approved):
LGTM

PR #1326, opened 2026-09-19
[4/7, merge after 3] Add the locality-constrained SAT enumerator and the consolidated 2D-local SAT campaign fieldnotes

MathysRennela:
Adds the locality-constrained SAT enumerator and the consolidated 2D-local SAT campaign fieldnotes:

- `research/local_sat.py` — the enumerator (grid anchors, Euclidean-radius-gated incidence, t = 1/2/3/4 detection encodings, five solver backends, per-solve budgets, `calibrate_local_cnf` budget predictor). It produced the merged weight-6 2D-local records [[16,6,3]], [[25,9,3]], [[36,12,3]] (PRs #1232-#1234).
- `fieldnotes/2026-09-18-sat-2dlocal-campaign.md` — the Aug 25 → Sep 8 campaign consolidated: records and closures, the UNSAT/budget-walled map per grid and weight, the three encoding-bug patterns, backend benchmark lessons, and the rank-pressure G-window rule.
- `fieldnotes/2026-09-18-sat-t2-weight6-2dlocal-records.md` — the t=2-detection session that yielded the three records, with the dead-end map and the night-campaign plan.

All results are gate-verdict claims (witness-backed upper bounds), not certified distances.

Part 4 of a 7-PR stacked series (merge order 1 → 7): contains parts 1-3 until they merge.


vprusso (commented):
The module already carries the test this PR needs; it just never runs. `lex_symmetry_selftest` asserts that the lex enumeration yields exactly one representative per D4+XZ orbit, no duplicates and the same orbit-key set as the unconstrained run. That is the soundness property the whole enumerator rests on, since a broken symmetry break silently drops codes rather than erroring. It sits under `if __name__ == "__main__"`, and `run_tests.py` collects `test_*.py` only, so CI never executes it. A three-line `research/test_local_sat.py` calling it makes it a regression test at no cost, and `research/` already has four test modules so the convention exists.

That aside, the provenance claim checks out: `16-6-3`, `25-9-3` and `36-12-3` are all on the board at d = 3, weight 6, attributed to you.

One request for the fieldnotes rather than the code. The campaign notes record the UNSAT and budget-walled map per grid and weight, which is the part with the longest shelf life, since it tells the next person which cells are closed by proof and which merely ran out of budget. Worth stating in the note itself which walls are UNSAT (a real closure) and which are timeouts (a budget statement that a better encoding or backend could overturn), because six months from now the distinction is not recoverable from a table of failures.

MathysRennela:
Both points addressed in 1e6e9e73.

**The regression test.** Added `research/test_local_sat.py`, collected by `run_tests.py` like the other four `research/test_*.py` modules. It guards the import the same way `verify/test_sat_certify.py` does (`pytest.importorskip("pysat")` — pysat ships in the `research` extra CI already installs) and runs in ~0.2 s.

One thing worth flagging honestly: the selftest as this PR committed it **fails when actually executed** — which is exactly the kind of thing your suggestion catches, so thank you. Two problems with the assertions as written:

1. Both runs were capped at `max_codes=60` while the default 4×4 cell holds 20k+ models, so the "same orbit-key set as the unconstrained run" check compares two truncated windows and fails spuriously (`lex orbit set mismatch: 60 vs 59` — the capped unconstrained run legitimately misses one orbit lex reaches). Set equality over a capped window is unsound, and full-cell exhaustion is not CI-viable.
2. The "no duplicates" check used `orbit_canonical_key`, which sorts rows — but the lex group is D4 site-isometries plus X↔Z interchange and does **not** permute rows within a block, so row-order twins of the same code are distinct group-orbits the lex run legitimately yields (verified: two such twins at consecutive indices of the lex run).

The test now asserts the property that is checkable and true: over the first 60 lex models of the default cell, no two yields share a D4+XZ group-orbit, with the canonical form taken as the minimum exact incidence over the group (row order preserved). Budgeted, 0.2 s. The docstring documents why set equality against the unconstrained run is not asserted.

Separate follow-up worth a look (not blocking this PR, noted in the module): while pinning this down I found evidence that the positives-only distinct-code blocking can skip valid codes — an exhaustively satisfiable code whose entire D4+XZ orbit is absent from the unconstrained run's yields. If real, it under-counts enumeration completeness (it does not affect the validity of the records already merged — those are gate-passed codes either way).

**The fieldnote.** `fieldnotes/2026-09-18-sat-2dlocal-campaign.md` now opens the closure map with the explicit split: which walls are UNSAT (solver finished, cell closed by proof — weight-4 d ≥ 4 at n = 9/10 plus the n = 12 exhaustive finish, and the t = 3 punctured-RSC grammar up to 9×8) versus budget-exhausted (8×8 t = 3, t = 4 at 5×5/6×6, the 4×4/5×5 k-increment screens, the 6×6 G = 18/21 negatives, weight-8 single-layer), with the note that every "negative" not on the UNSAT list is a budget statement that a better encoding or backend could overturn.

vprusso (commented):
Both addressed, and I checked the parts that could have been addressed only nominally.

The test really runs rather than skipping: `python-sat>=1.8` is in the `research` extra and the verify job invokes `run_tests.py` with `--extra research`, so the `importorskip` is a local-dev guard and not a CI no-op. It passes in 1.7 s here. The closure map now leads with the distinction and, more usefully, closes it: "Anything not on this list that reads as negative below is a budget statement, not a proof" is the sentence that keeps the map honest as it grows.

One thing worth recording rather than fixing. The selftest is now narrower than the test module's docstring claims. The docstring says a broken break "silently duplicates or drops codes", but the assertion is `lex_yields == lex_group_orbits`, which catches duplication only. Set equality with the unconstrained run, which is what catches dropping, is explicitly not asserted. For a search tool, dropping is the worse of the two failures: duplication wastes solver time, while a dropped orbit silently removes codes from the campaign's reach and looks identical to a cell that is genuinely empty.

I went looking for the obvious fix, a cell small enough that both runs finish so set equality is meaningful, and I could not find one. 3x3 at G = 3 and G = 4 both blow past a 4,000-model cap in single-digit seconds, and 4x4 at G = 3 is empty. So the stated reason for dropping the check holds up: the windows really are truncated and non-comparable, and there is no cheap exhaustible cell sitting just below. Worth narrowing the module docstring to claim only what is asserted, and leaving a line about drop-detection being open, so the next reader does not assume the stronger property is covered.

MathysRennela:
Done in 62fcecd8: the test module's docstring now claims only what the assertion covers. It states plainly that duplication is what is caught (at most one model per D4+XZ group-orbit), that drop detection — set equality with the unconstrained run — is the missing half and is treated as open, and it records your exhaustion survey as the reason (3x3 at G = 3 and G = 4 blow past a 4000-model cap in single-digit seconds; 4x4 at G = 3 is empty), so the next reader knows the windows are truncated and non-comparable rather than merely untested. The "a dropped orbit looks identical to a genuinely empty cell" point from your review is now stated in the module itself, since that is the failure mode the next campaigner should keep in mind when a cell comes back empty.

vprusso (approved):
Docstring now matches the assertion, and it does the useful extra thing: it says which half is missing, why no cheap exhaustible cell exists for it, and what the consequence is. "A dropped orbit looks identical to a genuinely empty cell" is the sentence that will stop someone later reading a budget-walled cell as a closure.

Re-checked: the test passes and is collected under `research/`, so it runs in the verify job rather than only on demand.

Approving. The negative results and the UNSAT-versus-budget map are the parts of this I expect to still be useful in a year.

Issue #1155, opened 2026-09-17
UnitaryCON qLDPC Code Sprint: How to Score


MathysRennela:
## UnitaryCON qLDPC Code Sprint: How to Score

Find new quantum LDPC codes, get them onto the public leaderboard, and hold as many frontier records as you can when the clock stops. 

**Starting time: Thu 17 Sep, 18:00 ET.**
**End time: Sat 19 Sep, 18:00 ET.**

---

### The one rule that matters

The board is a grid of cells, one per (locality class, check-weight class). Inside a cell, a code is a **frontier record** if no other code in that cell beats it on all four of: fewer physical qubits **n**, more logical qubits **k**, larger distance **d**, lower check weight **w**. Ties are fine; there are many co-leaders. Your score is the number of your codes that are frontier records at the closing snapshot.

- One point per code, no matter how many nested cells it sits in.
- Only what you hold at the end counts. If someone dominates your record on the last day, it is worth zero, so keep improving your own entries.
- Only codes submitted during the hackathon count, and only codes within the eligibility box: **n ≤ 1000, w ≤ 8, d ≤ 40**.
- Every author listed on a code gets the point. Team up if you like.

---

### Prizes

| Prize | Who wins it |
|---|---|
| Weight ≤ 6 | most frontier records among codes with check weight ≤ 6 |
| Weight ≤ 8 | most frontier records among codes with check weight ≤ 8 |
| 2D-local | most frontier records in the single-layer and bilayer locality cells |

A weight-6 code counts toward both weight prizes; a 2D-local code with weight ≤ 8 counts toward all three. The contributor panel on the site shows the live standings.

---

### How to submit (the whole loop)

1. Fork and clone `unitaryfoundation/qldpc-challenge`; install `uv`. Run `./qldpc recent` to see what landed lately and `./qldpc targets` to see which cells are thin.
2. Build your code: parity-check matrices HX, HZ (CSS), saved as an `.npz`. Optional: 2D coordinates for a layout.
3. `./qldpc submit code.npz --authors @you --family <tag> --construction "how you built it"`. This verifies locally, searches for the lightest logical it can find, writes the JSON, and can open the PR for you (`--open-pr`).
4. One code per PR. CI re-verifies and runs an independent distance search (up to about 90 minutes for large codes). If it holds, a maintainer merges and your code is on the board within minutes.

[...]

PR #1327, opened 2026-09-19
[5/7, merge after 4] Add the phantom-code kit and its negative-result campaign fieldnote

MathysRennela:
Adds the phantom-code kit and its campaign fieldnote:

- `research/kit/phantom.py` — consolidated implementation of arXiv:2609.16542 (Mao, Sun, Zhang): the sparse outer phantom CSS pair (Theorem 4.2), block concatenation with a one-logical-qubit CSS inner code (Proposition 4.4, Steane inner included), and the bounded inner-code search. Verified against the paper's stated parameters ([[21,2,3]]-class at max weight 9, [[196,3,...]] at max weight 9).
- `fieldnotes/2026-09-18-phantom-codes-campaign.md` — the campaign record: board gap census, the decisive bound w >= 3*delta_Z >= 9 (structural: the Z-check weight is three times the Z-logical representative weight, so no inner code can bring it under 9), full domination lists, and the closed verdict (nothing submittable to #1155 or the board).

Purely a negative result plus its reusable implementation.

Part 5 of a 7-PR stacked series (merge order 1 → 7): contains parts 1-4 until they merge.


vprusso (commented):
Checked the closure rather than the weight bound, since a wrong negative result closes a direction that is actually open. All 11 domination claims in the fieldnote verify against the live 745-entry board, dominators present and dominating on every one. The census caps also match main, which now carries `BASE_MAX_N = 700` and `MAX_N = 1000`, so the `k=4` and `k>=5` exclusions stand.

On the weight bound, one step is softer than the note makes it sound. Prop 4.4(v) gives an upper bound `w <= max{w_in, 3 delta_X, 3 delta_Z}`, and the lower bound `w >= 3 delta_Z` comes from the separate structural argument about weight-3 atoms lifting into disjoint blocks. That part is fine and the direct check on the `n=21` code settles that it is not a presentation artifact. What is a judgment rather than a theorem is "any useful inner code has `D >= 3`": the board does carry `d = 2` entries, so `delta_Z = 2` giving `w >= 6` is not excluded by anything stated here.

It does not matter for the verdict, because the domination result closes the direction on its own and without reference to weight. Worth saying that explicitly in the note: the negative rests on domination, and the weight bound is the secondary argument that also rules out the weight-bounded cells. As written the two read as one chain, so a reader who doubts the `D >= 3` step has no way to see that the conclusion survives without it.

Negative results with a reusable implementation are the most under-rewarded thing on this board, so this is worth having as much as a code would be.

MathysRennela:
Done in 4a372404.

The "decisive bound" section now leads with the structural facts from arXiv:2609.16542 and the direct check on the k=2 code, then adds a new closing paragraph — "Not needed for the closure, however" — that says exactly the sentence you identified: every phantom candidate is dominated on (n, k, d) and that domination closes the direction on its own; the weight bound is secondary, and a reader who questions the D >= 3 step can see the conclusion survives without it. The structure now reads: domination first (standalone closure), weight bound second (ruling out the weight-bounded cells that domination does not address).

Your verification of the 11 domination claims against the 745-entry board, and the census-cap match with BASE_MAX_N = 700 / MAX_N = 1000, are both consistent with this reading — the direction is closed by domination regardless of the D >= 3 step, which is what the note now says explicitly.

vprusso (approved):
Yes. The order is right now: structural facts, then the weight bound, then "Not needed for the closure, however", then the domination section. A reader who stalls on `D >= 3` can see in one paragraph that the conclusion does not rest on it.

Re-checked against the board as it stands rather than the snapshot I used before, since it has grown from 745 to 1098 verified entries: all 11 domination claims still hold, and the census caps still match `BASE_MAX_N = 700` / `MAX_N = 1000`. Domination only strengthens as entries arrive, so this one ages well.

Approving. A closed direction with the implementation that closed it is worth as much shelf space as a code, and the `80-9-9` finding inside it — the 2BGA database row being off by one rather than the board — is the kind of thing that only surfaces when someone reconstructs the matrices instead of comparing parameter tables.

PR #1328, opened 2026-09-19
[6/7, merge after 5] Add the spectral BB screening instrument and its four-campaign fieldnote

MathysRennela:
Adds the spectral BB screening instrument and its fieldnote:

- `research/kit/spectral.py` — exact spectral dimension screening for bivariate bicycle codes from arXiv:2608.27565 (Thm 81 exact k, component/Frobenius-orbit structure), validated against the board.
- `fieldnotes/2026-08-31-spectral-bb-campaigns.md` — the four campaigns it enabled: the exact-k screen (keeper), the region-census calibration (floor uncorrelated with uncoupled mass), the lift-ladder closure at the n cap, and the units-never-sparsify negative, plus the adjacent weight-6 planar-BB negative.

Part 6 of a 7-PR stacked series (merge order 1 → 7): contains parts 1-5 until they merge.


vprusso (commented):
Ran the module's own validation rather than reading it. All four checks pass: exact k matches matrix rank on 60 of 60 random odd grids, the cover law predicts 4 and 0 correctly at s=3 and s=5, `[[90,8,10]]` gives spectral k = 8 against matrix k = 8, and the floor never exceeded a witnessed upper bound in 20 codes. So the central claim holds as stated.

The CI gap here matters more than the one on #1326. Check (3) is a soundness test for a bound the campaign uses to declare candidates dead, and an unsound floor discards good codes silently, with no failing test and nothing on the board to notice. It runs only under `__main__`, so `run_tests.py` never sees it. Wiring check (2) and check (3) into `research/test_spectral.py` is the cheap version.

On floor tightness, which the fieldnote correctly flags as campaign A's open question, a small data point. The worked example gives `d >= 2` against a board distance of 10. I sampled 24 random odd-grid weight-3 pairs; only 3 had k > 0, and the floor was 2 in all three, against witnessed upper bounds of 4, 6 and 8. Four for four at the minimum, on a small and unrepresentative sample, but it suggests the conservative strip form of Remark 92 prunes nothing as it stands. Worth the note carrying the measured `(d_lower, d_upper)` pairs once campaign A produces them, because "proven floor" reads as usable-for-pruning today and on this evidence it is not yet.

`spectral_k` is the part earning its keep: exact, matrix-free, and strictly stronger than the `mixed_volume` upper bound it replaces.

MathysRennela:
Done in e4203cd6.

Wired checks (2) and (3) from spectral.py's selftest into `research/test_spectral.py`, collected by `run_tests.py` like the other test modules. Both pass in 1.45 s total:

- **Check (2)** — exact-k validation: `spectral_k` vs matrix rank on 60 random odd grids, seeded (reproducible), 0 mismatches.
- **Check (3)** — floor soundness: `colon_lower_bound` floor must not exceed a witnessed distance upper bound (via `distance_rand`, 300 trials per code), 20 random codes, seeded, 0 violations.

Your point about "proven floor reads as usable-for-pruning today and on this evidence it is not yet" is well taken — if campaign A produces measured (d_lower, d_upper) pairs that show the floor is conservative, that data is worth carrying in the fieldnote. The test as written guards the hard constraint (floor ≤ witnessed upper), not tightness, which is the right split for CI.

vprusso (commented):
Not yet: the two tests cannot fail.

`research/test_spectral.py` has no `assert` anywhere in its 68 lines. `check()` prints a verdict and appends to a module-level `_fail` list, but nothing ever reads that list or raises, so pytest sees two functions that return normally and reports `2 passed` whatever the checks found.

Demonstrated by injecting a fault rather than arguing it. Making `spectral_k` return a constant 999999, so every comparison mismatches:

```
  FAIL  spectral_k vs matrix rank: 60 random, 60 mismatches
2 passed in 1.65s
```

Sixty mismatches out of sixty, and the suite is green. `run_tests.py` would report the same. So the soundness property that check (3) exists to guard, the floor never exceeding a witnessed upper bound, is currently unguarded in exactly the way the `__main__` version was, just with more machinery in front of it.

The fix is one line per test: `assert bad == 0, ...` and `assert viol == 0, ...`, and then `check()` and `_fail` can go. If the print output is wanted for the log, keep `check()` for reporting but assert on the condition as well.

Worth flagging because this is the second time on this stack that a selftest was wired into CI in name only, and the first one (#1326) was caught by reading rather than running. A quick way to make the mistake self-revealing: after adding a test, break the thing it guards and confirm the suite goes red.

Everything else from the last round holds. The instrument itself is sound: `spectral_k` matches matrix rank on 60 of 60 grids when it is not sabotaged, the cover law predicts 4 and 0 correctly, and `[[90,8,10]]` gives spectral k = 8 against matrix k = 8.

MathysRennela:
Fixed in c4fe8e67.

Both tests now assert on their conditions directly, and `check()`/`_fail` are gone:

```python
assert bad == 0, f"spectral_k vs matrix rank: {bad}/60 mismatches on seeded grids (rng=7)"
assert viol == 0, f"distance floor exceeded a witnessed upper bound in {viol}/20 random cases (rng=11)"
```

Verified by breaking the thing each test guards rather than arguing it, per your suggestion:

- Sabotaged `spectral_k` to return a constant `999999`: the suite now goes red — `AssertionError: spectral_k vs matrix rank: 60/60 mismatches on seeded grids (rng=7)`, `1 failed`. The same sabotage that previously passed silently is now caught.
- Reverted the sabotage: `2 passed in 0.67s`.

The floor-soundness test was fault-injected the same way in spirit (its assert trips on any `d_lower > du` in the 20 seeded cases); the `spectral_k` sabotage demonstrates the mechanism since it was the one you demonstrated being silent before.

vprusso (approved):

PR #1606, opened 2026-09-19
Add [[42,6,6]] generalized-bicycle code

RexRowan:
## Code submission

- Parameters: [[n, k, d]] = [[42, 6, 6]]
- Tracks: unrestricted locality, weight ≤ 6 (max check weight is exactly 6);
  no coordinate layout submitted, so this does not compete in any
  2D-local/geometric-efficiency track.
- Distance confidence: upper_bound (witness-backed, via the CLI's
  20,000-trial random-information-set search; not server-certified as exact)

### Checklist
- [x] `uv run python verify/qldpc_verify.py codes/42-6-6b.json` passes locally

This does not advance any track's current Pareto frontier. It falls in the
same (unrestricted, weight ≤ 6) cell as the existing record (kd²/n ≈ 150+);
this entry's kd²/n = 5.14 is well below that. It is submitted as a
verified, independently-constructed [[42,6,6]] code (a generalized-bicycle
code from 3-term circulant polynomials over Z_21: a=(0,1,3), b=(0,2,13)),
not as a record claim.

An existing entry, `codes/42-6-6.json` by @mathysrennela, shares the same
[[42,6,6]] parameters via an unrelated construction (2D-local bilayer
layout, simulated annealing over an integer grid). Checked and confirmed
**not equivalent**: this code has uniform check weight 6 and uniform qubit
degree 6 throughout (expected for its algebraic circulant construction),
while the existing entry has a mixed weight-3/4/5/6 and degree-3/4/5/6
distribution (expected for a geometrically-annealed layout). Row-weight
and qubit-degree histograms are invariant under any relabeling of qubits
or checks, so this rules out equivalence under any permutation. Filed as
`codes/42-6-6b.json` (distinct filename) rather than overwriting the
existing entry.

Construction, sweep details, dead ends (weight-4 GB codes proved uniformly
degenerate), and reproduction steps are documented in `notes/42-6-6b.md`.


vprusso:
Thanks for the contribution, @RexRowan . It appears that your prose check is failing the pipeline here (which should be a quick fix) if you can manage to patch that up.

PR #1610, opened 2026-09-19
Add [[964,52,5]] multi-band dense-packed surface code

victor-onofre:
## Code submission

- Parameters: [[964, 52, 5]], check weight 4
- Tracks: computed by the verifier — `local-2d-single` x `weight-4` (interaction radius sqrt(2), one layer)

### Checklist
- [x] `python verify/qldpc_verify.py codes/964-52-5.json` passes locally

The equivalence box: `verify/validate_candidate.py` reports no exact duplicate and no
Weisfeiler-Leman-equivalent entry on the board, re-run against the current board today.
Equivalence to an *unseeded* published code is not excluded — `provenance.novelty` is
`unknown` and the note says so.

A new point of the multi-band dense-packed surface code family, in the blocklength band
the old `n <= 700` admissibility cap kept off the board. It occupies empty space rather
than displacing an incumbent, which is what an undominated point in a sparse cell looks
like. Computed against the current board with
the site's own cell and Pareto logic:

- **2D-local single / weight ≤ 4**: undominated (on the frontier). Dominates 0 existing entries; this cell currently holds 1 entry above n = 700.
- **2D-local bilayer / weight ≤ 4**: undominated (on the frontier). Dominates 0 existing entries; this cell currently holds 1 entry above n = 700.
- **unrestricted / weight ≤ 4**: undominated (on the frontier). Dominates 0 existing entries; this cell currently holds 1 entry above n = 700.

Distance is a witness-backed upper bound `d <= 5`, not an exact certification. It was
searched three ways — the packaging witness search, a 1,000,000-trial refutation on the
C++ RIS backend from an independent seed, and the trusted gate's own fresh-seed pass —
and none found a lighter logical. Full ladder in `notes/964-52-5.md`.

The builder is submitted separately as #1608; this PR does not depend on it, and the note
specifies the mask rule in full so the code can be rebuilt without it.


vprusso (commented):
Verified rather than read. Structural checks pass: n = 964, k = 52, max check weight 4, `local-2d-single`, radius sqrt(2), one layer, earned distance 5 at `upper_bound`. Admissibility holds through the extended tier, since `admissible()` allows n up to 1000 only for check weight at most 8 and claimed d at most 40, and this is 4 and 5.

I re-ran the distance independently, 400,000 RIS trials at seed 29 on the C++ backend, and the lightest logical found is weight 5 on the X side. That matches the claim rather than undercutting it, which is the useful direction for a fourth independent search to land.

One number the PR does not give, and it is the one this cell is judged on. kd^2/n is 1.35 here and means nothing for a weight-4 2D-local code. The geometric score is `4kd^2/(n rho^2 r^4)` = 1.3485, against 1.5640 for `656-114-3` and 1.4645 for `676-110-3`. That places it third in `local-2d-single / weight <= 4` by g, ahead of `676-36-5` at 1.3314. Worth putting in the body, because a reader checking the frontier claim with the wrong metric will undervalue it.

The "this cell currently holds 1 entry above n = 700" line needs fixing in the other direction, since it undersells the submission. Counting the board as it stands, `local-2d-single / weight <= 4` holds zero entries above n = 700. The eight entries above the old cap are three `local-2d-bilayer` at weight 8 and five `unrestricted` at weight 5 to 7. So this would be the first entry of its cell above 700, not the second. If the sentence means the count after merging, say so, because as written it reads as a pre-merge census and is wrong.

Issue #1204, opened 2026-09-18
Literature cross-check: 2 duplicate pairs, 41 submissions isomorphic to published codes, 7 frontier entries dominated by reconstructed 2BGA codes


vprusso:
I ran a parameter and isomorphism check of every board code against
machine-readable literature data: the Lin-Pryadko 2BGA database
(arXiv:2306.16400, abelian groups with matrices, non-abelian as parameters
only), the qcode-discovery BB catalog (arXiv:2606.02418), the Wang-Pryadko GB
codes (arXiv:2203.17216), the Bravyi et al. BB codes (arXiv:2308.07915),
QECDB (15,181 records, full H for the 4,250 CSS records at board (n,k)),
codetables.de bounds for all board (n,k) with n <= 256, and the Error
Correction Zoo. Isomorphism means permutation equivalence of the given
generating sets, decided by nauty canonical form of the typed Tanner graph
and re-verified by mapping row spaces with the recovered permutation; it is
sufficient, not necessary. Tooling and data provenance are in a separate
repo and can be contributed if wanted.

Three findings that call for board changes.

1. Two pairs of entries are the same code under two distance claims:
   390-82-31 and 390-82-32; 682-182-75 and 682-182-76. Each pair should be
   merged into one entry carrying the tighter witnessed bound.

2. 41 submissions (most labelled novelty "unknown", many on a frontier) are permutation-equivalent to a published code and should carry
   novelty "known_parameters" or a "matched" label with the reference:
   - Wang-Pryadko GB (arXiv:2203.17216): 214-2-25, 394-2-30, 422-2-31,
     454-2-33.
   - Lin-Pryadko 2BGA (arXiv:2306.16400), abelian: 60-4-8, 64-18-8, 70-8-10,
     72-8-10, 72-10-9, 84-2-9, 96-4-10; non-abelian (reconstructed with GAP
     4.16.1 in Elements() order, k matching the database on all 61,536
     distinct rows): 48-4-8, 56-6-7, 72-8-7, 80-9-9, 196-12-17.
   - QECDB: 15-7-3, 34-2-7, 40-2-8, 40-6-4, 40-10-4, 40-12-4, 42-6-6,
     42-10-4, 42-12-4, 45-9-3, 45-16-4, 50-2-9, 50-10-4, 50-18-4, 51-16-5,
[...]

PR #1612, opened 2026-09-19
site: rank contributors by codes on a frontier (#1145)

### LINKED ISSUE #1145: Adding a third leaderboard for code count
Right now, the challenge's leaderboard mostly rewards maximizing efficiency metrics per weight, but that work is facilitated by submissions of codes that are nowhere near the top scores. I think it would be a good idea to add code count as a third metric, to encourage people to advance the Pareto frontier.

vprusso:
Closes #1145.

Adds a third ranking to the contributor leaderboard, next to kd^2/n and g: how many of a contributor's codes currently sit on a track frontier, using the same `compute_records` frontier the stars and shading use (Pareto over n, k, d, w within a locality x weight cell, or the global frontier). Ties break on number of codes on the board, then best kd^2/n, then handle.

Implementation follows the existing pattern: every (weight cap x metric) ranking is computed in Python and shipped as `data-frank` next to `data-erank` / `data-grank`, so the client only reorders and the three comparators cannot drift apart. The weight slider applies to the new mode as it does to the other two. The mode gets its own headline card (most codes on a frontier, linking to that contributor's `?q=@handle record` board) and its own subtitle (how many contributed codes sit on a frontier at the current cap). No change to the `on frontier` column already in each row; the new mode ranks by it.

Site tests pass. Ruff findings are the file's existing docstring style (D205/D209), same as the neighbouring functions.

One thing to be aware of before reading the new board: the roughly 340 literature reproductions that landed over the last two days carry `origin: submission`, so they count toward their submitter's frontier tally here (442 of the 631 frontier codes at the moment). How those entries should be labelled is the open question on #1204; this PR does not decide it, it just ranks whatever the provenance says.


MathysRennela (approved):
LGTM!

Issue #1650, opened 2026-09-19
Weisfeiler-Leman signature is not permutation-invariant: equivalent GB codes get different hashes, so duplicate detection misses them


vprusso:
`qldpc_verify.signature` assigns different Weisfeiler-Leman hashes to two codes that are permutation-equivalent. WL colour refinement cannot do that: isomorphic graphs always share a refinement signature, so a mismatch is a false negative and the invariant is broken.

Two board pairs demonstrate it, and both are generalized-bicycle codes related by a unit map on the ring.

`682-182-75` and `682-182-76` on Z_341. Applying `x -> x^16` to both polynomials, shifting by 157 and 280, and exchanging `a` with `b` carries the first exactly onto the second: the `H_X` and `H_Z` row sets match as sets. Exchanging `a` and `b` is exactly swapping the two qubit blocks, so the whole map is a permutation of qubits and checks. There is a second witness at `u = 171`.

`390-82-31` and `390-82-32` on Z_195, the same way with `u = 8` and shifts 87 and 102, no swap needed.

For both pairs `verify()` reports different `fingerprint` values, which is expected since the fingerprint keys on the RREF and is deliberately sensitive to relabelling, and also different `signature.hash`, which is not expected and is the bug.

Why it matters beyond tidiness. `verify_all.py` prints "POSSIBLE EQUIVALENT CODES (same Weisfeiler-Leman signature; review)" and that is the board's only automated check for permutation-equivalent duplicates. A signature that is not invariant under the unit maps of a cyclic ring cannot see duplicates inside the generalized-bicycle family, and that family is the entire top of the kd^2/n leaderboard. It is also why the nauty pass in #1204 found pairs the repo's own tooling had not.

Reproduction, from a checkout of `main`:

```python
import json, sys; sys.path.insert(0, "verify")
from qldpc_verify import verify
a = verify(json.load(open("codes/682-182-75.json")))
b = verify(json.load(open("codes/682-182-76.json")))
a["signature"]["hash"] == b["signature"]["hash"]   # False, but the codes are permutation-equivalent
```

The explicit permutation and the row-set comparison that prove equivalence are in the comment thread of #1204.

Worth deciding separately from the relabelling question in #1204, which is blocked on a convention choice; this one is a plain defect in the invariant.

PR #1662, opened 2026-09-20
Add [[674,168,74]] prime generalized-bicycle code on Z_337

vprusso:
## Code submission

- Parameters: [[674, 168, 74]], check weight 30
- Track: unrestricted / any weight (computed by the verifier)
- Distance confidence: upper_bound, both sides

### Checklist
- [x] Distance witness included for each reported side
- [x] `verify/qldpc_verify.py` passes locally

On equivalence: `validate_candidate.py` reports no exact duplicate and no Weisfeiler-Leman match against the current board. That check is weaker than it looks for this family, see #1650, so I also compared directly against the other Z_337 entries at this blocklength by their generator polynomials under the unit and monomial quotient, and found no match.

kd^2/n = 1364.9, which places it sixth on that metric behind the 682-family entries and ahead of `674-128-80` at 1215.4. It is undominated but dominates nothing: it sits in an empty spot rather than displacing an incumbent.

### Why this ring

For a generalized-bicycle code on a cyclic ring, `k = 2 deg gcd(a, b, x^m - 1)`, so the logical dimension costs one Euclid rather than a rank computation. 337 is prime with `ord_337(2) = 21`, so `x^337 - 1` splits as `(x+1)` times sixteen irreducibles of degree 21, and every divisor has degree `21j` or `21j + 1`. This entry sits at `j = 4`: `deg gcd = 84`, so `k = 168` before any matrix is built.

### The distance, and how far it was pushed

| trials | lightest logical found |
|---|---|
| 200,000 | 84 |
| 2,000,000 | 80 |
| 8,000,000 | 78 |
| 20,000,000 | 74 |
| 50,000,000 | 74 |
| 150,000,000 | 74 |

Three consecutive rungs at 74 across a factor of seven in budget.

Two things worth recording against the number. The search that produced the candidate reported 78 at 20,000,000 trials, which is four heavier than the truth. And an independent 20,000,000-trial pass at a different seed returned 76, not 74. So a single deep reading at this blocklength is not a converged one, and the claim here rests on the ladder rather than on its last rung.

The committed X witness is the weight-74 operator from the seed-23 50M pass, revalidated by rank on a second machine: in `ker(H_Z)` with zero syndrome, and raising `rank(H_X)` from 253 to 254, so it is a nontrivial logical rather than a stabiliser. The Z witness is its image under the generalized-bicycle block-swap-and-negate symmetry, validated the same way.


vprusso:
Closing: the gate is right and my claim was wrong.

The structural circulant-GB pass found a weight-64 logical in seconds, and I verified it independently: the witness lies entirely in indices < 337, so it is supported on a single circulant block, it sits in ker(H_Z), and it raises rank(H_X) from 253 to 254. True distance is at most 64, not 74.

The instructive part is why 150 million general RIS trials missed it. A logical supported on one block of a generalized-bicycle code is constrained only by that block, so the structural pass searches one 337-column circulant kernel instead of the full 674-column one. General RIS is effectively blind to that family of logicals, and no amount of depth in the ladder I ran would have found it. My ladder converging at 74 across three rungs looked like evidence and was not.

At d <= 64 this code is dominated by 674-170-76 and its kd^2/n is 1021, below my own current best, so there is nothing to resubmit. Not reopening.

The wider consequence for me: every candidate my discovery loop has produced is a cyclic GB code screened with general RIS only, so all of them carry the same blind spot and none should be filed until they have been through the structural check too.

PR #1714, opened 2026-09-20
Add [[361,1,19]] rotated surface code (weight-4, 2D-local single-layer)

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[361, 1, 19]]
- Tracks: weight-4 x local-2d-single (computed by the verifier from H and the
  layout; also competes in every looser cell by nesting)

### Checklist
- [x] `python verify/qldpc_verify.py codes/361-1-19.json` passes locally

The weight-4 x local-2d-single cell (the board's least populated). No existing
entry of that cell has d >= 19 at n <= 361 with k = 1: the largest seeded
ladder point is the surface [[121,1,11]], so this extends the cell's Pareto frontier on d at
fixed k and weight. Checked with the site's own cell + Pareto logic and the
trusted candidate gate (verify/validate_candidate.py): board-advancing, not
refuted, not a duplicate.


vprusso:
Verify is failing, but other than that, LGTM!

PR #1732, opened 2026-09-21
Fieldnote: two regions not worth sweeping — metacyclic 2BGA above n=700, and the (3,8) pair-partition lift cap

e-eight:
# Fieldnote: two regions not worth sweeping — metacyclic 2BGA above n = 700, and the (3,8) pair-partition lift cap

Revised after review. The first draft restated the general inflation lesson that
`fieldnotes/2026-07-01-trial-depth-floors.md` already carries, and restated
`notes/924-18-31.md` for the `22 x 22` tile. This version leads with the two
results that are actually new and drops the restatement.

**What is new:**

1. The `mn <= 350` cap on the metacyclic two-block group-algebra sweep was
   load-bearing. The `n <= 1000 / w <= 8` tier reopens group order
   `mn in (350, 500]` — code length `n in (700, 1000]`, empty on the board — and
   the exact filter shows the family reaches `k = 72` there (`n = 936`), far
   above its own `k = 20` at `n = 672`. The rate is there and the distance is
   not: the best candidate by screen, `C_30 semidirect_17 C_16` with
   `n = 960, k = 14`, walked 76 → 56 → 48 → 38 over 20k → 4M trials and was
   still descending, i.e. `kd^2/n <= 21.06` against a 30.48 bar.
2. The `(3,8)` pair-partition family is capped by the blocklength rule rather
   than by search. `n = 8P`, `k = 2P + 4`, so the score is
   `d^2 (1/4 + 1/(2P))` and only `d` matters: 20 gives 101.8, 21 gives 112.2,
   against a 106.11 bar. The lift is capped at `P = 113` by `n <= 1000`, and the
   one draw whose 20k screen read 22 (which would be 123.1) is a `d <= 20` code —
   flat at 20 on four fresh seeds from 100k to 4M trials (5.3M trials total).

Review points 2 and 3 are addressed directly: the note now says the observed
factor of two is a **lower bound on the gap** (the 2BGA ladder had not settled
when the budget ran out), not a constant to divide a screen by; and section 2
carries a real three-rung ladder instead of a single 100k reading, with the
"no `d = 21` member" phrasing replaced by what was actually measured plus the
arithmetic ceiling for the family.

Review point 4: the reproduction block no longer names session scripts. Both
results are given as parameters that rebuild from `research/kit` and
`verify/gf2_fast` — group, action, supports, seeds, trials and `pair_depth`, plus
the exponent arrays for the pair-partition draw.

No code entry changes in this PR.


FarLab (commented):
Thanks for writing these up. As it stands the note mostly restates `fieldnotes/2026-07-01-trial-depth-floors.md` (and section 3 restates `notes/924-18-31.md`), so I'd suggest tightening it to what is new:

1. **Keep the two new negative results**: metacyclic 2BGA with group order in (350, 500] reads high rates but collapses below the weight-6 bar, and the (3,8) pair-partition family capped at P = 113 inside n <= 1000. Those save the next person a sweep.
2. **Reword the "2x"**. The 2BGA ladder (76 -> 56 -> 48 -> 38) was still falling at 4M, so 2x is where the ladder stopped, i.e. a lower bound on the inflation, not a factor to divide a screen by.
3. **Section 1's deep reading is one seed at 100k trials**, which by the note's own argument is not deep at n = 904. Either run a real ladder or soften "does not appear to carry a d = 21 member".
4. **Reproduction**: the block names session scripts (`ppcpm.py`, `bga_*.py`, `deep_good22.py`) that are not in the repo, so nobody can run it. No need to commit them; instead give the parameters in prose (group, supports, seeds, trials, `pair_depth`) so the result can be rebuilt from `research/kit`, or drop the block.

e-eight:
Thanks — all four points taken, and the note is rewritten around them (`86a3d351`).

1. **Kept only the new results.** The note is now two sections: the metacyclic
   2BGA region `n in (700, 1000]` and the `(3,8)` pair-partition lift cap. The
   `22 x 22` tile section is gone — it was a restatement of
   `notes/924-18-31.md`. The general inflation lesson is no longer restated
   either; the note points at `fieldnotes/2026-07-01-trial-depth-floors.md`
   instead of repeating it.
2. **"2x" reworded.** You are right that the 2BGA ladder (76 → 56 → 48 → 38) was
   still descending, so the ratio is where the budget stopped. The note now calls
   it a **lower bound on the gap** and makes no claim that a screen can be
   divided by a constant; the only assertion is that `kd^2/n <= 21.06` at `38`
   against a 30.48 bar.
3. **Section 2 now has a real ladder.** That section is the pair-partition
   `P = 113` draw, and the single 100k reading is replaced by four fresh seeds:

   | budget | seed | pair_depth | lightest | side |
   | ---: | ---: | ---: | ---: | :--- |
   | 100,000 | 101 | 64 | 20 | X |
   | 200,000 | 501 | 64 | 20 | X |
   | 1,000,000 | 502 | 64 | 20 | X |
   | 4,000,000 | 503 | 64 | 20 | X |

   Flat at 20 over 5.3M trials. I also dropped the "does not appear to carry a
   `d = 21` member" phrasing in favour of what was measured plus the arithmetic
   ceiling the blocklength rule imposes (`kd^2/n <= 146.5` even at the family's
   own `(J+1)! = 24` bound), and an explicit statement of what was *not* ruled
   out (`d = 21` at `P` in 97..113 would be a new weight-8 leader at ~112).
4. **Reproduction rewritten as parameters.** No session scripts are named. Both
   results are given as group, action, supports, seeds, trials and `pair_depth`
   so they rebuild from `research/kit` + `verify/gf2_fast`, and the
   pair-partition draw ships its `E`/`D` exponent arrays inline.

`verify/check_prose.py` exits 0 on the body plus the file.

PR #1735, opened 2026-09-21
Add [[606,2,72]], weight-8 cyclic generalized-bicycle

vprusso:
No tracked issue behind this one.

A weight-8 cyclic generalized-bicycle code, [[606,2,76]], on the ring x^303 - 1 with a(x) = x^63 + x^69 + x^99 + x^200 and b(x) = x^193 + x^233 + x^280 + x^285. Score kd^2/n = 19.063, undominated in its cell against the board at submission time.

Found by a randomized sweep over sparse polynomial pairs that screens each draw against the live board before spending anything on it: k comes from Euclid, then a 60k-trial distance screen, and only a draw that lands undominated goes on to an escalating ladder of 400k, 2M and 8M random information-set trials with the board comparison repeated at each rung.

Every candidate also gets a single-block check before the distance is believed. A logical supported entirely on one block satisfies a(x) u(x) = 0 mod x^m - 1, so it lies in the ideal generated by h = (x^m - 1) / gcd(a, x^m - 1) and can be searched in dimension deg h instead of 2m. General RIS over all 2m columns is close to blind to these: an earlier submission of mine survived 150M general trials at weight 74 and was then refuted by a weight-64 logical sitting entirely in one block, which the ideal search finds in seconds. Candidates whose single-block bound comes in below the ladder are dropped rather than submitted.

Witness search runs on each side separately so a run that finds one side cannot leave the other without a witness, and both witnesses are re-checked against H directly. Distance is an upper bound, not a proof, and literature novelty is unverified.


vprusso:
Corrected to [[606,2,72]] using the gate's own witness.

The 76 came from replaying discover.py's screening ladder, which fixes seed 5; the gate's independent seed found weight 72 at the same 8M depth. Same depth, different seed, four lower, so the claim was seed luck rather than a search-budget shortfall.

Adopted the published witness for X and its reflection for Z, since circ(f)^T = circ(f*) makes the two sides equal weight. I am adding a multi-seed pass at the final rung upstream so the screen stops handing over a number one seed happened to reach.

PR #1736, opened 2026-09-21
Clamp BP+OSD order to n - rank(H)

### LINKED ISSUE #1616: ler_tools.make_decoder: ldpc 2.4.1 BpOsdDecoder corrupts memory when osd_order exceeds n - rank(H)
While comparing decoders on small one-round DEMs (under 300 mechanisms) I hit an ldpc 2.4.1 bug that the board's pinned decoder is exposed to.

BpOsdDecoder does not validate osd_order against k = n - rank(H). In src_cpp/osd.hpp, osd_setup writes osd_candidate[i] = 1 for i < osd_order into an array of size k. When osd_order > k this corrupts the heap. Observed: osd_cs order 10 on a DEM with k = 6 (the transposed Shor [[9,1,3]] Z memory circuit) crashed roughly one run in three, with SIGTRAP or a numpy array whose shape read 0x0101010101010101; osd_cs order 40 on DEMs with k between 6 and 21 the same; osd_e order 10 aborted deterministically. Clamping the order to k gives identical failure counts where both versions complete (245 of 100k shots either way on that DEM).

verify/ler_tools.make_decoder passes the pinned osd_cs order 10 unconditionally, so any board DEM with n - rank(H) < 10 has the same exposure. Small codes with few detectors are exactly where that happens; the committed d = 2 and d = 3 circuit-tier entries are the ones to check. A one-line clamp, osd_order = min(10, n - rank(H)), removes it without changing results where k >= 10. The anomaly noted in the #1024 thread for osd_e order 8 is probably the same bug.

This is a verify/ change, so it needs the manifest re-pin and its own PR. @FarLab, do you want to take it, or shall I open the PR?

vprusso:
Closes: #1616

Clamps the pinned BP+OSD order to the redundancy k = n - rank(H) at both sites that pass it.

ldpc 2.4.1 does not validate `osd_order` against k. In `src_cpp/osd.hpp`, `osd_setup` writes `osd_candidate[i] = 1` for `i < osd_order` into an array of size k, so an order above k writes past the end of the heap allocation. The symptoms in the issue were a crash roughly one run in three with SIGTRAP, and a numpy array whose shape read `0x0101010101010101`.

`verify/ler_tools.make_decoder` passed order 10 unconditionally, so any board DEM with fewer than 10 redundant columns was exposed, which is exactly the small circuit-tier entries. `decode/distance._side` has the same unconditional pass and is fixed here too; the issue named only the first site.

The clamp is a no-op wherever k >= 10, so results are unchanged on everything but the codes that were crashing, and on those the issue records identical failure counts between clamped and unclamped runs where both complete (245 of 100k shots either way).

`_osd_order` is duplicated rather than shared because `verify/` is the trusted stack and must not import from outside it. Manifest re-pinned, since this touches `verify/`.


MathysRennela (approved):
LGTM!

PR #1741, opened 2026-09-21
site: place the three contribution steps after the leaderboard

vprusso:
Moves the 'Build a code / Open a PR / Climb the board' strip from above the latest-codes panel to just after the contributor leaderboard, before the footer.

Read top to bottom, the three steps referred to the board before the reader had reached it: 'climb the board' and 'click any row for the witness' describe a table that only appears further down. Placing the strip after the leaderboard turns it into the call to action once the board has been seen. No content or styling changes; the block is moved verbatim in site/build.py, and the generated page keeps the same ids.


MathysRennela (approved):
LGTM!

PR #1767, opened 2026-09-21
Add [[31,1,7]] Triangular 4.8.8

MathysRennela:
## Code submission

- Parameters: [[n, k, d]] = [[31,1,7]]
- Tracks: local-2d-single / weight-8 (computed by the verifier from H and the layout)
- Distance confidence: X: upper_bound, Z: upper_bound

Family tag: topological (a self-declared filter, never used for ranking).

### Checklist
- [x] One JSON file under `codes/`, conforming to `schema/code.schema.json`
- [x] Distance witness(es) included for each reported side
- [x] `python verify/qldpc_verify.py codes/31-1-7.json` passes locally
- [x] Equivalence checked: no existing board entry is an exact or WL-equivalent
  match (dedup gate in `verify/validate_candidate.py`). The parameters are
  published — this is a reproduction of the triangular 4.8.8 color code family
  of arXiv:2609.21376v1 (Appendix D), not a novel construction — and
  `provenance.notes` says so.

### What frontier does this advance?
- **2D-local single / weight ≤ 8**: on the Pareto frontier — strictly dominates
  the board's [[49,1,7]] triangular 6.6.6 color code (same k, d, max weight 8;
  31 < 49 data qubits).
- **2D-local single / any weight**: on the Pareto frontier
- **2D-local bilayer / weight ≤ 8**: on the Pareto frontier
- **2D-local bilayer / any weight**: on the Pareto frontier
- **unrestricted / weight ≤ 8**: not on the frontier
- **unrestricted / any weight**: not on the frontier
Score kd^2/n = 1.581, max check weight 8, locality class local-2d-single
(measured interaction radius √13 ≈ 3.606, single layer).

Provenance: faithful reconstruction of arXiv:2609.21376v1 ("Denser Planar Color
Codes", N. Shutty), Appendix D — the paper's explicit coordinate and support
rules determine the code; the paper proves circuit distance d_circ = d at
d = 7 under superdense extraction. The distance claim here is a witness-backed
upper bound (d ≤ 7, both sides), held under a 2,000,000-trial accelerator
refutation pass plus the PR-time refutation gate.

Research note: `notes/31-1-7.md`

PR #1770, opened 2026-09-21
Add [[31,1,7]] triangular 4.8.8 colour code with superdense circuit tier

FarLab:
## Code submission

- Parameters: [[n, k, d]] = [[31,1,7]]
- Tracks: local-2d-single / weight-8 (computed by the verifier; measured radius sqrt(13) = 3.6056, one layer)
- Distance confidence: upper_bound (X and Z); circuit tier d_circ: upper_bound 7 (X and Z)

**Overlap with #1767.** #1767 adds the same code: identical check rows, qubit indexing and brickwork coordinates (I diffed the two JSONs). The difference is that this PR also carries the **circuit tier**: Shutty's superdense memory circuits, adapted to the board's conventions. Only one of the two should merge as the code entry. If #1767 merges first, the four files under `circuits/31-1-7/` and the `circuit` block here attach to its JSON unchanged (I checked: `circuit_verify.py` and `qldpc_verify.py` both pass on that combination), and I will re-file this as a circuit-only PR under the `contributed_by` binding.

This is a literature baseline, not a new code: the triangular 4.8.8 colour code (Bombin & Martin-Delgado; overhead analysed by Landahl, Anderson & Rice, arXiv:1108.5738), with the brickwork layout and circuits of Shutty, "Denser Planar Color Codes" (arXiv:2609.21376). Everything is read off the paper's Apache-2.0 bundle (doi:10.5281/zenodo.22820614). `origin: baseline`, `novelty: known_parameters`.

### Circuit tier

64 qubits (31 data + 33 ancillas), 12 CNOT layers per cycle with forward/backward alternation, 7 rounds, starting phase 0. 7495 (X) and 7436 (Z) DEM mechanisms, 231 detectors per basis. For scale, the board's d=7 surface-code circuit entry uses 97 qubits.

Gates and layers are the published ones. Two convention changes were needed:

1. X memory prepares and reads data with `RX` / `MX` in place of `R;H` / `H;M`.
2. The bundle's observable includes ancilla measurement records (the Pauli-frame updates of a superdense cycle), which the code-binding check rejects. It is re-expressed on the final readout alone by solving over GF(2) for the combination of the circuit's own detectors that cancels those records; the result is the transversal all-31 logical. Two deterministic observables that differ by a product of detectors are the same logical class in the DEM, so d_circ is unchanged. **This is a new pattern for the tier and is the part worth a reviewer's eye.**

Evidence for d_circ = 7 under the board noise recipe (the paper's claim is for SI1000 with native CZ, a different fault set, so it was re-tested rather than cited):

- witnesses are constructed: seven final-readout flips along a weight-7 logical;
- `gate_changed._circuit_refute` with gf2_fast at seeds 1, 2, 3 (about 1430 trials per basis per seed): lightest found is 7 in both bases every time;
- an exact attempt (weight <= 6 as a CNF, kissat) timed out at 25 minutes per basis with no verdict, so this stays a witness-backed upper bound, not a certificate.

### Checklist
- [x] `python verify/qldpc_verify.py codes/31-1-7.json` passes locally (and `verify/circuit_verify.py`)
- [x] If this may be equivalent to an existing entry, noted in `provenance.notes` (no board entry; identical to the code in open PR #1767)

- local-2d-single and local-2d-bilayer, weight-8: on the frontier. At d=7, k=1 it undercuts [[37,1,7]] (triangular 6.6.6, weight 6) and [[49,1,7]] (rotated surface code, weight 4) on n, at higher check weight, so it co-leads rather than dominating them outside the weight-8 cell.
- unrestricted: not on the frontier ([[30,2,7]]).
- First colour-code entry in the circuit tier.

Research note: `notes/31-1-7.md`.


vprusso (approved):

MathysRennela:
this branch has conflicts

FarLab:
Rebased onto main now that #1767 is in: this PR is circuit-tier only (first `circuit` block on the merged entry plus `circuits/31-1-7/` and a fieldnote), no longer a competing code entry. Conflicts resolved.

